PR #1, opened 2018-11-04
Add CMake support to the project

tobanteGaming:
Hi TheCherno,

I updated your project to be built with CMake. I thought it would be a good idea to start using CMake early on, so it will be easier to migrate to different platforms.

I only added three ```CMakeLists.txt``` files & updated ```.gitignore ```. I did no changes to the source code, so it's still windows only. The CMake project already has the necessary structure for future platforms & external libraries (e.g. logging). I tried to document the CMake files as much as possible, so hopefully, there are no unclear lines.

You can obviously still build the project using the Visual Studio solution file ```.sln```. I'm happy to maintain the CMake build files if you don't want to, but they only need to change for new c++ files, defines or new libraries anyway.

Here is an example on how to build the project with CMake from git-bash, since I'm not very familiar with cmd:

```sh
cd Hazel
mkdir build && cd build
cmake ..
cmake --build .
```
Then you can copy the hazel-engine.dll to the sandox folder and run the .exe

Hope this helps. 

Greetings,
**tobante**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,12 @@
-# Directories
+# Editors
 .vs/
+.vscode/
+.atom/
+.idea/
+
+# Build & Binary
+Build/
+build/ 
 bin/
 bin-int/
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,15 @@
+cmake_minimum_required(VERSION 3.8)
+
+# Project
+project(Hazel)
+
+# Options
+option(HAZEL_BUILD_SANDBOX "Build sandbox application" ON)
+
+# Hazel
+add_subdirectory(Hazel)
+
+# Sandbox
+if(HAZEL_BUILD_SANDBOX)
+    add_subdirectory(Sandbox)
+endif()
--- a/Hazel/CMakeLists.txt
+++ b/Hazel/CMakeLists.txt
@@ -1,0 +1,40 @@
+project(hazel-engine)
+
+# Add sources (only .cpp)
+set (SANDBOX_SOURCES
+    src/Hazel/Application.cpp
+)
+
+# Define shared library
+# Create nice looking alias for clients of Hazel to link to 
+add_library(${PROJECT_NAME} SHARED ${SANDBOX_SOURCES})
+add_library(Hazel::Engine ALIAS ${PROJECT_NAME}) 
+
+# Platform specific preprocessor defines
+# PUBLIC  = define is visible to the user of Hazel
+# PRIVATE = define is visible Hazel only
+if (WIN32)
+    target_compile_definitions(${PROJECT_NAME} PUBLIC HZ_PLATFORM_WINDOWS=1)
+    target_compile_definitions(${PROJECT_NAME} PRIVATE HZ_BUILD_DLL=1)
+endif()
+if(APPLE)
+    # for MacOS X or iOS, watchOS, tvOS (since 3.10.3)
+    target_compile_definitions(${PROJECT_NAME} PUBLIC HZ_PLATFORM_APPLE=1)
+endif()
+if(UNIX AND NOT APPLE)
+    # for Linux, BSD, Solaris, Minix
+    target_compile_definitions(${PROJECT_NAME} PUBLIC HZ_PLATFORM_UNIX=1)
+endif()
+
+# Set include directories
+target_include_directories(${PROJECT_NAME} PUBLIC ${PROJECT_SOURCE_DIR}/src)
+target_include_directories(${PROJECT_NAME} PRIVATE ${PROJECT_SOURCE_DIR}/src/Hazel)
+
+# Language standard
+target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_17)
+
+# Link libraries (e.g. logging)
+target_link_libraries(${PROJECT_NAME} 
+    # ${CMAKE_THREAD_LIBS_INIT}     # Needed for unix builds to link to pthread
+    # LOGGING
+)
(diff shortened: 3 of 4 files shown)


IronicallySerious:
This is great.

With Visual Studio, the latest CMake works with MSBuild to dynamically link external libraries.

However, on Linux and Mac, these linkages would be needed to be done/ordered manually as I had learned while working in a similar project which used libraries like GLFW, OpenGL and SFML.

The necessary changes needed to be done for using these libraries to work on other platforms can be dealt with later though.

EdSwordsmith:
Shouldn’t you write use install(...) to make the executable and library files appear in Hazel/bin/Debug-x64/(...)

tobanteGaming:
@EdSwordsmith Sure definitely an option, or we could use ``` add_costum_command ``` to copy the dll after the sandbox build. But the ```install``` command is probably the cleaner solution.

taxborn:
@IronicallySerious 

> However, on Linux and Mac, these linkages would be needed to be done/ordered manually as I had learned while working in a similar project which used libraries like GLFW, OpenGL and SFML.

Yeah that'll be a problem down the road, but as of currently isn't Hazel only supporting Windows in the early days?

Gatunox:
Great idea, Thanks tobanteGaming

khaledhn1:
Good Point.

IronicallySerious:
@taxborn 

> Yeah, that'll be a problem down the road, but as of currently isn't Hazel only supporting Windows in the early days?

Yes, it will only be a problem down the road, not right now. I am just acknowledging the need to add more Linux and Mac targetting CMake scripting in the future. Hazel is planned to be made cross-platform as TheCherno explained in the intro video.

Jarf10:
For what I find on the internet the syntax is a bit easier for Premake as it is used as the lua language. So thats good!

I only wonder why Cherno likes to have 2 repositories apperently one for his patreons and one for the public. Can't he just only put the videos for the patreons up earlier? In that way the community can get more involved in the project by creating these pull requests. When Cherno has grown a big community this engine may even get really serious like the other big engines out there. Who knows but its up to Cherno and his patreons I think. :) Keep on the good work!

klukule:
Second repository is cherno's development repo... Where he tests And implements the fetures as he goes... And patrons have access to it.. So it's more like interactive sneakpeak, since it will be implemented here in the future. And patrons are recommended to report issues and make pull requests to this repo :)

rpmcdougall:
This was extremely helpful to help me as I only own Macs and use CLion and I didn't feel like bootcamping to install Visual Studio. Thanks for doing this, glad to be able to follow along!

S-Marais:
I forked this repo and created a new branch called `linux-master`. It's aimed at running the project on linux before cross platform is implemented. As a solution manager it uses CMake which can help people use it. The project will use premake tho but the plugin for premake to generate CMake is 5 years old at best and can be messy.
I use CLion on all platforms and this IDE uses CMake only to manage projects so I'll keep CMake updated.

flamendless:
@S-Marais thanks! Im on linux so i definitely need that

_Sent from my Liquid Zest 4G using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

GauravRajSharma:
@S-Marais Hey, can you help me with CMake? I am trying to learn this build system. Can you give suggestions from where should I learn CMake?

S-Marais:
@GauravRajSharma I've learnt by starting with small project and then expanding my wish looking for answers mostly in the CMake documentation and stackoverflow. You can already check the CMakeLists.txt from my linux-master branch as I intentionally keep it as simple as possible using minimal CMake functions and for now I don't use custom macros or functions.

GeorgeWeb:
The CMake files seem good, but one thing that probably won't be accepted with the current state of this **MR** is the edited .gitignore. The reason is that `build/` and `Build/` directories explicit ignore is redundant, even inappropriate is the project aims building in `bin/` directory instead. Also, the comment replacement (`Directories` to `Editors`) can be considered noise but may pass if the CMake does.
Other than that, great for CMake die-hard fans or people that just have a CMake-bound IDE setup.

IndustriousNomad:
I use Codeblocks as my IDE with MinGW ( GCC 8.2.0 ) x64 and although I can follow along the tutorials and everything works fine, including C++17, PREMAKE is the one thing that doesn't play nice. It's not universal. CMake is. So I really hope he takes CMake to heart.

Gaztin:
@IndustriousNomad Have you tried https://github.com/chris-be/premake-codeblocks? Premake is just as universal as CMake. It just relies on community support rather than IDE integrations.

tobanteGaming:
@TheCherno 

My main reason for wanting CMake was the ease of use when it comes to **IDE support**. **VS, VS Code, CLion** & more can open folders with a `CMakeLists.txt` directly without running a generate command first.

Also once we have Linux support it would not be difficult to support other  Unix platforms, such as FreeBSD or OpenBSD, which as far as I know are not supported by premake currently (I might be wrong!). I know I'm probably the only one using those platforms on a daily bases, but good portability never hurts. I will close it, for now, once we have Linux support I might open a new issue to discuss other Unixes.

I can totally understand why you dislike CMake, old `CMakeLists.txt` pre 3.0 looked horrible, but with the "new" syntax it's a lot better. Also integrating other libraries is mostly easier with CMake since they already use it. 

Thanks for all the create videos.

tobanteGaming (toby)

IndustriousNomad:
@tobanteGaming  Why close this though ? It's still needed to be open. This is NOT just for linux that CMake is needed. I'm on Windows. Read my above post to understand why.

TheZitroX:
cmake on top. dont close this thread

PR #2, opened 2018-11-05
Use unique pointers instead of raw pointers

mvandi:
Hello Yan,

I have updated the `EntryPoint.h`, `Application.h` and `SandboxApp.cpp` files replacing the raw application pointer with an unique pointer so that its disposal can be managed by the executing program.

I have also added a Post-Build Event in the Sandbox project that copies the `Hazel.dll` in the Sandbox output directory.
Adding this Post-Build Event prevents the manual copy of the generated `.dll` file every time the solution (or the project) is compiled.

Greeting,
Mattia


IronicallySerious:
Can you explain how you managed to add a Post-Build Process through the UI in VS? I would like to know how it's done. All I am seeing are some changes in the `.vcxproj` file.

mvandi:
Of course, just right-click on the project, go to `Preferences > Build Events > Post-Build Event`, then you can the batch commands to execute once the project is built.

In this case: `xcopy $(SolutionDir)bin\$(Configuration)-$(Platform)\Hazel\Hazel.dll $(OutDir)`

PhilipKatis:
Shouldn't you use `xcopy /y` to avoid errors about the file already existing?

Gaztin:
Good initiative with smart pointers!

However, along with the migration to the Premake build system, these post-build commands are already implemented.

ldm0:
Sounds great! But it maybe a bit early to send a pull request for this project. 🤣

TheCherno:
Hi Mattia,

Thank you for your pull request! Here are some points:
- I'm deliberately not using a unique_ptr here because it's both unnecessary _and_ I want to be able to explicitly call delete to shutdown the application when required. For example the code later may read:
```cpp
int main(int argc, char** argv)
{
    Hazel::InitCore();
    auto app = Hazel::CreateApplication();
    app->Run();
    delete app;
    // Shutdown core after app shutdown
    Hazel::ShutdownCore();
}
```
Yes, you could argue that I could just enclose `app` in a scope, but this works just as well, and keeps the client API a little cleaner (`Hazel::Application*` vs `std::unique_ptr<Hazel::Application>`), and I happen to be one of those people who doesn't mind having raw pointers around because they're really not as bad as people have made them out to be...
- The `return 0;` in main is not needed (see 4) under Explanation here: https://en.cppreference.com/w/cpp/language/main_function)
- The post-build copy step is great, thanks for that. However we're going to switch to using a build system next episode and this will be obsolete since we won't define our own vsxproj files.

PR #16, opened 2019-01-09
Replace compiler flags with premake properties

Dragon9815:
I updated the premake file with the properties for the runtime library.

The type of the runtime library is set by the to properties `runtime` and `staticruntime`
- `runtime` specifies if we want the debug or he release library
- `staticruntime` specifies if we want static linking or dynamic linking

This replaces the build options `/MDd` and `/MD`.


TheCherno:
Obsolete; this has already been done. Thanks for the suggestion. 🙂

PR #17, opened 2019-01-16
Start Project

AleksandarSherbula:
Added start project to be Sandbox in premake.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -20,6 +20,8 @@ include "Hazel/vendor/GLFW"
 include "Hazel/vendor/Glad"
 include "Hazel/vendor/imgui"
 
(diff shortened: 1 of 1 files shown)


AleksandarSherbula:
Not exactly sure what do you mean? prj.name is used as a reference to the project name when you are in its scope in premake. There is nothing that can be defined here(as far as i am aware of).

vamidi:
Like startproject "%{prj.name}"

AleksandarSherbula:
You can't do that, since it's out of scope of Sandbox project in premake. Otherwise, the script will have no idea what are you refering to.

vamidi:
Ahh that's too bad alright thanks!

PR #22, opened 2019-01-27
Postbuild command for Hazel project modified

Husamdababneh:
this will ensure that XCOPY command will know that argument number 2  is folder not a file


Husamdababneh:
Done!!

Husamdababneh:
Just to make it clear ... Premake for some reason is removing the last slash from the destination argument (argument number 2) which windows will consider it a file not a folder ... so what i did is just making sure that Premake don't remove that slash by adding quotation marks bracing the destination

zixin96:
Would doing this make sure that VS will create a folder if it doesn't exist, and thus solving the issue of first build failure?

PR #26, opened 2019-02-03
Modify Premake configuration to suppress warnings

IoanThomas:
Hi Cherno, this is just a small change to suppress the warnings you mentioned in your latest video ([C4996](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996) and [C4251](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4251) specifically) when compiling for Windows.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -1,6 +1,9 @@
 workspace "Hazel"
 	architecture "x64"
 	startproject "Sandbox"
+	
+	filter "system:windows"
+		disablewarnings { "4996", "4251" }
 
 	configurations
 	{


Feacur:
I suggest fixing the source of these warnings, like mentioned in #32

IoanThomas:
@Feacur that would be ideal. You should create a pull request for the fixes that you provided in your issue. I created this pull request because Cherno said in his video that he wanted a way to suppress these warnings because ["they're fine".](https://youtu.be/h0OgLT6KJi8?t=16m18s)

Husamdababneh:
> C4251 can be ignored if you are deriving from a type in the C++ Standard Library, compiling a debug release (/MTd) and where the compiler error message refers to _Container_base.

https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4251?view=vs-2017

since we are using standard libraries I think disabling the warning if just fine.

Gaztin:
Is this still relevant? C4251 is not an issue any more since we moved to static linking.

IoanThomas:
You are correct. I had not checked this repository in a while so I didn't know that `_CRT_SECURE_NO_WARNINGS` has been defined in the Premake configuration file and that the switch to static linking had removed the C4251 warnings. I will mark this pull request as closed.

PR #31, opened 2019-02-06
Add MacOS support

ProfessorCampi:
Hi! I've added Mac support to Hazel. If you decide to merge this, you must also merge [TheCherno/glfw/#3](https://github.com/TheCherno/glfw/pull/3) so that the premake5.lua will work on Mac.

### DIFF
--- a/Hazel/src/Hazel/Core.h
+++ b/Hazel/src/Hazel/Core.h
@@ -6,8 +6,10 @@
 	#else
 		#define HAZEL_API __declspec(dllimport)
 	#endif
+#elif HZ_PLATFORM_MACOS
+	#define HAZEL_API
 #else
-	#error Hazel only supports Windows!
+	#error Hazel does not support your platform!
 #endif
 
 #ifdef HZ_DEBUG
@@ -15,8 +17,13 @@
 #endif
 
 #ifdef HZ_ENABLE_ASSERTS
-	#define HZ_ASSERT(x, ...) { if(!(x)) { HZ_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
-	#define HZ_CORE_ASSERT(x, ...) { if(!(x)) { HZ_CORE_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
+	#ifdef HZ_PLATFORM_WINDOWS
+		#define HZ_ASSERT(x, ...) { if(!(x)) { HZ_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
+		#define HZ_CORE_ASSERT(x, ...) { if(!(x)) { HZ_CORE_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
+	#else
+		#define HZ_ASSERT(x, ...) { if(!(x)) { HZ_ERROR("Assertion Failed: {0}", __VA_ARGS__); } }
+		#define HZ_CORE_ASSERT(x, ...) { if(!(x)) { HZ_CORE_ERROR("Assertion Failed: {0}", __VA_ARGS__); } }
+	#endif
 #else
 	#define HZ_ASSERT(x, ...)
 	#define HZ_CORE_ASSERT(x, ...)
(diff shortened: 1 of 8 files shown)


Hazel/vendor/Glad/premake5.lua
```diff
@@ -1,25 +1,38 @@
...
+		staticruntime "On"
         
-    filter { "system:windows", "configurations:Release" }
-        buildoptions "/MT"
\ No newline at end of file
+	filter { "system:windows", "configurations:Release" }
+		buildoptions "/MT"
+
+	filter { "system:macosx", "configurations:Release" }
+		buildoptions "/MT"
```
WhoseTheNerd: Why enable MT build options flag when `staticruntime` is on?

Hazel/src/Hazel/Core.h
```diff
@@ -6,22 +6,29 @@
...
 #endif
 
 #ifdef HZ_DEBUG
 	#define HZ_ENABLE_ASSERTS
 #endif
 
 #ifdef HZ_ENABLE_ASSERTS
-	#define HZ_ASSERT(x, ...) { if(!(x)) { HZ_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
-	#define HZ_CORE_ASSERT(x, ...) { if(!(x)) { HZ_CORE_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
+	#ifdef HZ_PLATFORM_WINDOWS
```
WhoseTheNerd: Why separate the code when they are the same? `spdlog `should be cross platform already.

ProfessorCampi: Because there is no `__debugbreak();` on macOS.

Hazel/vendor/Glad/premake5.lua
ProfessorCampi: My bad... I've removed this now.

premake5.lua
```diff
@@ -48,12 +59,31 @@ project "Hazel"
...
+
+	filter {}
+
+	links
+	{
 		"GLFW",
 		"Glad",
-		"ImGui",
-		"opengl32.lib"
+		"ImGui"
```
Gaztin: Where did `opengl32.lib` go for Windows?

premake5.lua
```diff
@@ -30,8 +30,19 @@ project "Hazel"
 	targetdir ("bin/" .. outputdir .. "/%{prj.name}")
 	objdir ("bin-int/" .. outputdir .. "/%{prj.name}")
 
-	pchheader "hzpch.h"
-	pchsource "Hazel/src/hzpch.cpp"
+	filter "action:vs2017"
```
Gaztin: I think this can be changed to `filter "action:vs*"` to cover all versions of Visual Studio

Hazel/src/Hazel/Events/Event.h
```diff
@@ -29,7 +29,7 @@ namespace Hazel {
 		EventCategoryMouseButton    = BIT(4)
 	};
 
-#define EVENT_CLASS_TYPE(type) static EventType GetStaticType() { return EventType::##type; }\
+#define EVENT_CLASS_TYPE(type) static EventType GetStaticType() { return EventType::type; }\
```
Gaztin: Was this intentional? I don't see how this would compile on any system. The point of `##type` is to concatenate whatever type you pass to `EVENT_CLASS_TYPE`

ReDucTor: You don't need token pasting for it, your not merging things its easily identifiable by the preprocessor

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -48,6 +48,13 @@ namespace Hazel {
 			s_GLFWInitialized = true;
 		}
 
+		#ifdef HZ_PLATFORM_MACOS
```
ReDucTor: Might want to checkout PR #137 its doing something similar, with a comment suggesting this approach instead.

RamiAwar:
I got this working too but the problem is so much of the OpenGL code is not supported pre 4.5 versions. For example texture loading, I had to rewrite nearly everything there and in two versions  using a dirty ifdef implementation. Also in Create vs Gen for example (Buffers, VertexArrays etc...). 
I ended up moving my macosx implementation to a separate branch since these things are so different and I don't want to support 2 OpenGL versions.

PR #33, opened 2019-02-07
LayerStack only need to search for the layer to be pop in its correspond location

3x3y3z3t:
Currently the LayerStack is divide into two part: any Overlay will be pushed on the top, and any (other) Layer will be inserted in the middle (on top of inserted Layer).
When popping a layer, we den't need to search the whole stack for that layer: we only search the top half of the stack for an Overlay (there is no Layer pushed in there) and search the bottom half of the stack for a Layer (there is no Overlay inserted in there).


WhoseTheNerd:
This will break the code. `Expression: vector iterators in range are from different containers`

MrBrN197:
How do we know where the layer is divided?

3x3y3z3t:
We have `m_LayerInsert` to store the position where the new Layer is inserted (not Overlay), so that is the point where the layer stack is divided.
@WhoseTheNerd your comment makes me doubt my understanding of `iterator`s..

LovelySanta:
I've updated this PR to the current changes present in the LayerStack.cpp on the master branch.
Currently waiting on @3x3y3z3t to merge 3x3y3z3t/Hazel#1 into this PR.

LovelySanta:
@Gaztin this is ready to be merged

3x3y3z3t:
Sorry for having a hard time to keep up. I've updated the project to keep it up to date with the master branch, ready to be merged. Thank @LovelySanta.

Gaztin:
Good work!

Issue #34, opened 2019-02-09
push_back/emplace_back in LayerStack invalidates m_LayerInsert iterator


pooria1366:
Currently when we push_back/emplace_back (in case of overlays), since memory is reallocated, m_LayerInsert iterator gets invalidated.
And afterwards when we're pushing a layer, since we're using the invalidated m_LayerInsert iterator, it'll throw an exception.
One solution is to use an index instead of m_LayerInsert iterator and generate an iterator by m_Layers.begin()+index whenever we needed one to use insert/emplace methods.

PapaFl:
I might be wrong, but right now it looks like the iterator isnt necessary at all as it always points to `begin()`, which doesnt look like intended behavior.
Also, `LayerStack::PopLayer`s call to `erase` would also invalidate the iterator if it worked as intended.

MrBrN197:
@PapaFl I'm new to c++ but I also noticed this and was so confused. because it looked like `m_LayerInsert` was always  supposed to be the same as `m_Layers.begin()`. @pooria1366 when inserting into the vector the insert returns another iterator that we assign to `m_LayerInsert`.

Can anyone explain to me why when we erase a layer in `PopLayer()` we decrement `m_LayerInsert`.

PapaFl:
@MrBrN197 `m_LayerInsert` is not supposed to point to `begin()`, but to the top of the layer stack (or the bottom of the overlay stack) so that new layers will always be added below overlays.

MrBrN197:
@PapaFl I think we are saying the same thing if what your saying is that the way the current system is created is flawed; I also think it is. We should be able to insert layers at the end of the layers and overlays at the end of the vector. 
But this also makes me question how this layers are supposed to work. Does the order of layers matter(I know the order of overlays matters), are we going to always process every single layer?

PapaFl:
@MrBrN197 Yup, thats exactly what im saying. i've implemented that behavior in #37 

The order of layers matter. In `Hazel::Application`, you can see that Events get dispatched from the end of the vector to the beginning, and `Layer::OnUpdate` gets called from the beginning to the end.
This is done so that whatever layer/overlay gets rendered/updated last is the first one to receive events.

Gaztin:
Closing because this has been fixed.

PR #35, opened 2019-02-09
Fix iterator invalidation in LayerStack.

rtri:
A call to `LayerStack::PushOverlay` can cause `m_Layers` to be resized and invalidate `m_LayerInsert`, which will lead to UB in subsequent `PushLayer`'s. This PR simply reconstructs the iterator to prevent nastiness.

Edit: seems I'm late to this particular party, https://github.com/TheCherno/Hazel/issues/34 already noticed the same issue.

### DIFF
--- a/Hazel/src/Hazel/LayerStack.cpp
+++ b/Hazel/src/Hazel/LayerStack.cpp
@@ -21,7 +21,11 @@ namespace Hazel {
 
 	void LayerStack::PushOverlay(Layer* overlay)
 	{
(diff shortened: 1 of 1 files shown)


MrBrN197:
I don't know if this is the best way to deal with. I think we are not far enough into using this layers to see how it should be properly implemented.

Also I don't quite understand what you've written, your getting the difference between `m_LayerInsert` and `begin()` then proceeding to set 'm_LayerInsert` to `begin()` + that same difference

Gaztin:
Closing because this has been fixed.

PR #49, opened 2019-03-11
Layer's OnAttach and OnDetach methods aren't called inside LayerStack implementation.

gold604:
Hey! 

I've just moved calls to OnAttach event from Application LayerStack wrappers to actual LayerStack implementation, because I think it makes more sense and could possibly in future cause errors.

And I've noticed as well that OnDetach event wasn't called at all so it is now called in LayerStack implementation.

Have a nice day!


LovelySanta:
Made pull request to update the code (gold604/Hazel#1) as OnAttach is called **after** adding it to the layer stack, same way OnDetach must be called **before** removing it from the layer stack.

I also updated the file so it is not conflicting for merge into Hazel.
Current status: waiting for @gold604 to merge into this pull request.

LovelySanta:
@Gaztin not sure what the exact merge conflicts are, but both master and pull request files should look the same (except from additional lines added). It's ready to merge.

Gaztin:
LGTM. Nice find!

gold604:
This PR introduced a missing semicolon. @Gaztin Could you have a look at my new PR #72 which fixes that, please? Thanks!

bovacu:
Hi, I wrote another memory leak issue I found on issue #96 and I've been told to share here the problem so everything is together. You can read the issue for full info and the full log of memory leaks, but i'm doing a little sum up.

Important: I've checked this issue on both my repo and last repo of Hazel, so is not just a problem of mine.

I was wandering if there was any memory leaks during the writing of the code (just in case I missed something or wrote something wrong) so I looked for a library to check memory leaks on Windows and the web page of Visual Studio gave this code (marked with <-):

```
#ifdef CGE_WINDOWS
extern CGE::Game* CGE::BuildGame();

#define _CRTDBG_MAP_ALLOC      <-
#include <stdlib.h>                        <-
#include <crtdbg.h>                      <-

int main(int argc, char** argv) {
	CGE::Logger::initLogger();

	auto _game = CGE::BuildGame();
	_game->run();
	delete _game;

	_CrtDumpMemoryLeaks();      <-
}
 #else
#error Only windows supported yet
#endif
```

As I wrote on top, the full log is in the issue #96 (I can post it here if needed) and there's a lot of memory leaks. I figured out that they appear once on the premake we start using 
`
buildoptions "/MDd" / buildoptions "/MD"`

what we change to 

`runtime "Debug" / runtime "Release" `
and 
`staticruntime "Off"`

The only pointers we have to delete are the Layers from the LayerStack and we are deleting them, so it might be some GLFW problem where we aren't calling a specific function that needs to be called (?)

I don't really know, so any suggestion is welcomed. As I said, the output of the code provided by Visual Studio is here #96 and there are a lot of memory leaks.

PR #50, opened 2019-03-21
Added support for mode-switched fullscreen and borderless fullscreen

jonestristand:
Added support for mode-switched full screen and borderless fullscreen to the window classes. Saves previous location and size of windowed mode window after full screen or borderless mode. All arguments optional so this is a non-breaking change to clients using Hazel's exposed API.

### DIFF
--- a/Hazel/src/Platform/Windows/WindowsWindow.h
+++ b/Hazel/src/Platform/Windows/WindowsWindow.h
@@ -21,6 +21,7 @@ namespace Hazel {
 		inline void SetEventCallback(const EventCallbackFn& callback) override { m_Data.EventCallback = callback; }
 		void SetVSync(bool enabled) override;
 		bool IsVSync() const override;
+		void SetWindowMode(const WindowMode& mode, unsigned int width, unsigned int height) override;
 
 		inline virtual void* GetNativeWindow() const { return m_Window; }
 	private:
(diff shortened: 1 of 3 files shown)


Hazel/src/Platform/Windows/WindowsWindow.h
```diff
@@ -21,24 +21,35 @@ namespace Hazel {
...
+		void SetWindowMode(const WindowMode& mode, unsigned int width, unsigned int height) override;
 
 		inline virtual void* GetNativeWindow() const { return m_Window; }
 	private:
 		virtual void Init(const WindowProps& props);
 		virtual void Shutdown();
 	private:
 		GLFWwindow* m_Window;
+		GLFWmonitor* m_PrimaryMonitor; // Stores a reference to the primary monitor
+		GLFWvidmode m_BaseVideoMode;   // Stores the underlying video mode being used by the OS
```
Gaztin: If we store the primary monitor from the first time we create the window, it's always going to choose that monitor regardless if the user changes primary monitor while the game is running. I think it's safe enough to fetch the primary monitor for each call to `SetWindowMode`.

WhoseTheNerd: I think it is better to refetch those rather cache it.

CleverSource:
This PR is about 5 months old now and none of the requested reviews were changed. Should someone consider remaking this PR or closing this one as it's clearly being neglected?

jonestristand:
Sorry, big life changes recently. I'll get right on it!

LovelySanta:
Since regular full screen works, see https://github.com/TheCherno/Hazel/pull/108#issuecomment-531545843 and since we had put in a lot of effort to make this work, I think we can close this issue, as it is more than functional enough for now.

PR #52, opened 2019-04-01
Added .editorconfig

jonestristand:
.editorconfig file added to solution so that all contributors use similar style (e.g. tabs not spaces, crlf line endings)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = crlf
+insert_final_newline = true
+indent_style = tab
--- a/premake5.lua
+++ b/premake5.lua
@@ -1,3 +1,25 @@
+-- Implement the workspace files command for solution-scope files
+require('vstudio')
+premake.api.register {
+	name = "workspace_files",
+	scope = "workspace",
+	kind = "list:string",
+}
+
+premake.override(premake.vstudio.sln2005, "projects", function(base, wks)
+	if wks.workspace_files and #wks.workspace_files > 0 then
+		premake.push('Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{' .. os.uuid("Solution Items:"..wks.name) .. '}"')
+		premake.push("ProjectSection(SolutionItems) = preProject")
+		for _, path in ipairs(wks.workspace_files) do
+			premake.w(path.." = "..path)
+		end
+		premake.pop("EndProjectSection")
+		premake.pop("EndProject")
+	end
+	base(wks)
+end)
+
+-- Premake configuration begins here
 workspace "Hazel"
 	architecture "x64"
 	startproject "Sandbox"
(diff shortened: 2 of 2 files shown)


Gaztin:
LGTM although I would probably like to see a `premake_customizations.lua` file in the future where we can shove all the overrides and customizations.
That said, I think @TheCherno will need to approve of this before it can be merged.

CleverSource:
I'd love to see this happen, I keep seeing Lovely or Gaztin have to tell people to "keep code inline with the coding style" when this would solve the issue. Hopefully Cherno can check this out soon.

LovelySanta:
I would like to see a `premake_customizations.lua` or `premake5preferences.lua` file in root directory that is included in the main file with the overrides instead of having it clutering above the workspace...

jonestristand:
I've implemented a separate file for premake customizations. I think this can now be merged.

T

premake5.lua
```diff
@@ -144,4 +151,6 @@ project "Sandbox"
 	filter "configurations:Dist"
 		defines "HZ_DIST"
 		runtime "Release"
-		optimize "on"
+		optimize "On"
```
LovelySanta: why did you change this line? Doesn't it work in all small caps? In that case `"configurations:Release"` has the same issue? Either one of those requires changes then...

jonestristand: No idea why I did that. Reverted

.editorconfig
```diff
@@ -0,0 +1,8 @@
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = crlf
```
WhoseTheNerd, line 6: Unix style newlines? This is dos style newlines. Should be `end_of_line = lf`.

premake_customizations.lua
```diff
@@ -0,0 +1,20 @@
...
+require('vstudio')
+premake.api.register {
+	name = "workspace_files",
+	scope = "workspace",
+	kind = "list:string",
+}
+
+premake.override(premake.vstudio.sln2005, "projects", function(base, wks)
+	if wks.workspace_files and #wks.workspace_files > 0 then
+		premake.push('Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{' .. os.uuid("Solution Items:"..wks.name) .. '}"')
```
Gaztin, line 11: Where does this UUID come from, exactly?

premake5.lua
```diff
@@ -9,6 +11,11 @@ workspace "Hazel"
 		"Dist"
 	}
 
+	workspace_files 
+	{
+		".editorconfig"
+	}
+
```
Gaztin, line 18: Overscoped a little by calling it `workspace_files`. I think this can just be an `editorconfig` field since we have no plans to use it for anything else.

premake_customizations.lua
LovelySanta: It's a GUID, and it's the one for the "solution scope". I do feel like defining it as a constant and then using that constant instead would make more sense. I found a list of the GUID in VS [here](https://www.codeproject.com/Reference/720512/List-of-Visual-Studio-Project-Type-GUIDs).

premake5.lua
LovelySanta: I agree that workspace files sounds a bit weird. Since it's VS related, I would tend to call it `solution files` ? 0x actualy called them Solution Items, which sounds appropriate as well, since it's not limited to files only...

PR #64, opened 2019-06-03
premake5.exe missing fix. 

TomerZeitune:
Added premake5.exe and changed GenrateProjects.bat to generate for the latest version of Visual Studio (vs2019).


DaveAxiom:
Git isn't supposed to host binary files and besides, developers can download their own updated copy of premake5.

TomerZeitune:
Premake5.exe is a small executable 1.3 MB in size. It shouldn't be problematic to include it inside the Hazel repo for one click easy generation of project files. Currently GenerateProjects.bat just doesn't work so either fix it or remove it.

DaveAxiom:
That executable isn't any help to me since I'm on a Linux system. That file is readily available otherwise. Tell me why you must include an executable file that is a simple download.

Gaztin:
I agree with @DaveAxiom on this one. It's not about how big or small the executable is, but rather about platform independence.

TomerZeitune:
Well this is more of a design discussion. But according to this: https://youtu.be/sULV3aB2qeU done by Cherno at the Premake episode I think that GenrateProjects.bat being totally broken isn't intentional. I am pretty sure Cherno also mentioned this problem in a video. So this is a fix.

Gaztin:
The intention was probably for each person to download the premake executable on their own and put it in /vendor/bin/premake/ themselves.

Gaztin:
It looks like Cherno added the missing premake executable. Closing since the issue was resolved.

PR #65, opened 2019-06-06
Using default destructors over empty ones

### LINKED ISSUE #38: Using default destructors over empty ones
You are using empty destructors for your interfaces. This is potentially performance bug since class which is not `std::is_trivially_destructible<A>::value`. This is usually not a problem, but IMHO it is better to explicitly use default instead of empty destructors. Example of it is Window class.

AMS21:
Closes #38. See that issue for explanation why.


Gaztin:
LGTM

PR #66, opened 2019-06-10
Use direct state access in OpenGL Buffer classes

AHUMANTHATISHUMAN:
The reason to use `glCreateBuffers` is that it actually creates the buffers rather than just giving unused names and creating them when they are first bound. This way, the direct state access functions (GL_ARB_direct_state_access) can preserve the currently bound buffer without needing to restore it later or check if the buffer has been previously bound. If the buffer is bound immediately after the call to `glCreateBuffers`, it will require OpenGL 4.5 for no good purpose.


JiggelTime:
I see hidden binding going on in the Hazel master branch, eg. in AddIndexBuffer and AddVertexBuffer in OpenGLVertexArray. If the switch to DSA is made, the hidden binding can be avoided. DSA simply makes more sense for OOP-style classes.

I strongly dislike the idea of having hidden driver overhead of unnecessary binding/unbinding/etc.. inside certain OpenGL calls in a game engine.

Most modern devices support OpenGL 4.5, and since Hazel is currently taking the "get stuff working, add other platform implementations later" road, creating a fallback older (non-DSA) OpenGL implementation can be done later in the future if support for those devices is desired.

LovelySanta:
> Most modern devices support OpenGL 4.3, and since Hazel is currently taking the "get stuff working, add other platform implementations later" road, creating a fallback older (non-DSA) OpenGL implementation can be done later in the future if support for those devices is desired.

In the _"get stuff working, add other platform implementations later"_ approach I would say revert to older implementation as it will work on more devices. But TheCherno also stated that he only wanted to support OpenGL 4.5 (not lower).

> I strongly dislike the idea of having hidden driver overhead of unnecessary binding/unbinding/etc.. inside certain OpenGL calls in a game engine.

TheCherno said he wanted to focus on performance to get as close to "real time" as possible. In that case I am also supporting to not have the extra overhead in the calls. But the 'line' between renderer and renderAPI must be taken somewhere. This is more up to TheCherno himself to decide.

AHUMANTHATISHUMAN:
I'm working on updating it. I'll finish when I have time.

AHUMANTHATISHUMAN:
Done. It works fine with direct state access now. Textures were already using it.

Hazel/src/Platform/OpenGL/OpenGLBuffer.h
```diff
@@ -15,6 +15,8 @@ namespace Hazel {
 
 		virtual const BufferLayout& GetLayout() const override { return m_Layout; }
 		virtual void SetLayout(const BufferLayout& layout) override { m_Layout = layout; }
+
+		friend class OpenGLVertexArray;
```
WhoseTheNerd: Why is `friend class OpenGLVertexArray;` declared twice?

Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
```diff
@@ -50,19 +51,30 @@ namespace Hazel {
 	{
 		HZ_CORE_ASSERT(vertexBuffer->GetLayout().GetElements().size(), "Vertex Buffer has no layout!");
 
-		glBindVertexArray(m_RendererID);
-		vertexBuffer->Bind();
+		//glBindVertexArray(m_RendererID);
```
WhoseTheNerd: Remove commented lines.

Hazel/src/Platform/OpenGL/OpenGLBuffer.h
Puddlestomper: I think it is once for each of the two classes in this file.

Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
```diff
@@ -1,10 +1,11 @@
 #include "hzpch.h"
 #include "OpenGLVertexArray.h"
+#include "OpenGLBuffer.h"
 
 #include <glad/glad.h>
 
-namespace Hazel {
```
Puddlestomper: There is an extra open line here.

Hazel/src/Hazel/Application.cpp
```diff
@@ -29,6 +29,7 @@ namespace Hazel {
 		PushOverlay(m_ImGuiLayer);
 	}
 
+
```
Puddlestomper: This is just an extra open line.

Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
```diff
@@ -50,19 +51,30 @@ namespace Hazel {
...
 		{
-			glEnableVertexAttribArray(m_VertexBufferIndex);
-			glVertexAttribPointer(m_VertexBufferIndex,
+			//glEnableVertexAttribArray(m_VertexBufferIndex);
+			//glVertexAttribPointer(m_VertexBufferIndex,
+			//	element.GetComponentCount(),
+			//	ShaderDataTypeToOpenGLBaseType(element.Type),
+			//	element.Normalized ? GL_TRUE : GL_FALSE,
+			//	layout.GetStride(),
+			//	(const void*)(intptr_t)element.Offset);
```
Puddlestomper: More comments to remove as @WhoseTheNerd said.

LovelySanta:
The conflicting files just required to pull the master branch into this PR

Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/66/commits/d91579dd7ff16b35eb4a684faa738f5bf1638b3d

Hazel/src/Hazel/Application.cpp
LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/66/commits/d91579dd7ff16b35eb4a684faa738f5bf1638b3d

Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/66/commits/d91579dd7ff16b35eb4a684faa738f5bf1638b3d

Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/66/commits/d91579dd7ff16b35eb4a684faa738f5bf1638b3d

Hazel/src/Platform/OpenGL/OpenGLBuffer.h
LovelySanta: This is indeed once for the `OpenGLVertexBuffer` and once for the `OpenGLIndexBuffer`

Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
```diff
@@ -71,8 +73,8 @@ namespace Hazel {
 
 	void OpenGLVertexArray::SetIndexBuffer(const std::shared_ptr<IndexBuffer>& indexBuffer)
 	{
-		glBindVertexArray(m_RendererID);
-		indexBuffer->Bind();
+		glVertexArrayElementBuffer(m_RendererID,
+			std::dynamic_pointer_cast<OpenGLIndexBuffer>(indexBuffer)->m_RendererID);
```
LovelySanta: Same reasoning, convert to a static_pointer_cast.

```suggestion
			std::static_pointer_cast<OpenGLIndexBuffer>(indexBuffer)->m_RendererID);
```

Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
```diff
@@ -50,19 +51,20 @@ namespace Hazel {
...
-		vertexBuffer->Bind();
-
 		const auto& layout = vertexBuffer->GetLayout();
 		for (const auto& element : layout)
 		{
-			glEnableVertexAttribArray(m_VertexBufferIndex);
-			glVertexAttribPointer(m_VertexBufferIndex,
+			glEnableVertexArrayAttrib(m_RendererID, m_VertexBufferIndex);
+			glVertexArrayVertexBuffer(m_RendererID, m_VertexBufferIndex, 
+				std::dynamic_pointer_cast<OpenGLVertexBuffer>(vertexBuffer)->m_RendererID,
```
LovelySanta: Why the requirement to do a `dynamic_pointer_cast`? We're inside the OpenGL renderer, so we can assume we only use OpenGLVertexBuffer. With this reason, can't we do a `static_pointer_cast` instead?

```suggestion
				std::static_pointer_cast<OpenGLVertexBuffer>(vertexBuffer)->m_RendererID,
```

PR #68, opened 2019-06-16
Take shared_ptr by value in VertexArray.

Pocakking:
Take shared_ptr by value in VertexArray. This will potentially reduce the number of times the shared_ptr's ref count needs to be (atomically) incremented/decremented, which is a (relatively) slow operation. Therefore potentially increasing performance.

### DIFF
--- a/Hazel/src/Hazel/Renderer/VertexArray.h
+++ b/Hazel/src/Hazel/Renderer/VertexArray.h
@@ -13,8 +13,8 @@ namespace Hazel {
 		virtual void Bind() const = 0;
 		virtual void Unbind() const = 0;
 
-		virtual void AddVertexBuffer(const std::shared_ptr<VertexBuffer>& vertexBuffer) = 0;
-		virtual void SetIndexBuffer(const std::shared_ptr<IndexBuffer>& indexBuffer) = 0;
+		virtual void AddVertexBuffer(std::shared_ptr<VertexBuffer> vertexBuffer) = 0;
+		virtual void SetIndexBuffer(std::shared_ptr<IndexBuffer> indexBuffer) = 0;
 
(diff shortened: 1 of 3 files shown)


WhoseTheNerd:
Micro-optimizing is not a good thing to do.

Gaztin:
I fail to see how this is reducing the amount of atomic reference counting. All you've done is move the place where it copies the shared_ptr from here: https://github.com/TheCherno/Hazel/blob/0d292a31d7b2827799388ffcfa677c90ed0ccb30/Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp#L70 to here: https://github.com/TheCherno/Hazel/blob/4e4b797bfb1d310dc8306219280e0be3cd048e91/Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp#L49

Pocakking:
> I fail to see how this is reducing the amount of atomic reference counting. All you've done is move the place where it copies the shared_ptr from here:
> 
> https://github.com/TheCherno/Hazel/blob/0d292a31d7b2827799388ffcfa677c90ed0ccb30/Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp#L70
> 
> to here:
> https://github.com/TheCherno/Hazel/blob/4e4b797bfb1d310dc8306219280e0be3cd048e91/Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp#L49

This lets you move it in.

Gaztin:
But you're just "moving in" the copy of the shared_ptr that is passed to the function as an argument.
Later when you're calling it, it will look something like:
```cpp
const std::shared_ptr<VertexBuffer>& ref = ... // Fetch from somewhere
m_VertexArray->AddVertexBuffer( ref ) // <-- Reference count happens here because we're passing the std::shared_ptr by value.
```

Gaztin:
Closing for now. Please reopen if you have more to add.

PR #72, opened 2019-07-02
Added a missing semicolon

gold604:
Hello, there's a semicolon missing in LayerStack.cpp. I've added it back and it should be O.K.
Have a nice day!


JarrettSJohnson:
Automatic building upon PR would catch bugs like this. I wonder if they'll consider integrating CI.

PR #76, opened 2019-07-07
Optimizing BufferLayout::CalculateOffsetsAndStride

LovelySanta:
Both m_Stride and offset do the += element.Size operation. We can remove the requirement for the offset variable and only use m_Stride to do the same calculation. This operation will not require new memory allocation of a `uint32_t offset`.


WhoseTheNerd:
Stack allocation is like the fastest thing in c++. So clearly this is micro-optimization.

IronicallySerious:
I think these operations are optimized pretty neatly by the modern compilers. Better would be to keep the code as understandable as possible without worrying about micro optimisations

PR #77, opened 2019-07-08
Generate project in separate directory, keep submodule dirs clean

halcyonx:
Changed the output directory for premake generated projects and obj files.
I think it is better to keep generated solution and intermediate binaries in the separated directory than in many other places in the repo.

The result is following directory tree:
─project                           
 ├───build                         
 │   ├───bin                       
 │   │   └───Debug-windows-x86_64  
 │   │       ├───Glad              
 │   │       ├───GLFW              
 │   │       ├───Hazel             
 │   │       ├───ImGui             
 │   │       └───Sandbox           
 │   └───bin-int                   
 │       └───Debug-windows-x86_64  
 │           ├───Glad              
 │           ├───GLFW              
 │           ├───Hazel             
 │           ├───ImGui             
 │           └───Sandbox           
 ├───Hazel                         
 └───Sandbox                       

There are also requests for glfw and imgui submodules:
https://github.com/TheCherno/imgui/pull/6
https://github.com/TheCherno/glfw/pull/7


DaveAxiom:
All intermediate build files are stored in "bin-int/". This is explicit to keep all output directories "bin" and "bin-int" in alphabetical order to make it easy to delete.

halcyonx:
I think there is no any reason to keep 'bin', 'bin-int' or any other generated files in submodules or in other subdirectories of repository.
now git shows us:
`
    modified:   Hazel/vendor/GLFW (untracked content)
    modified:   Hazel/vendor/imgui (untracked content)
` + a generated files like *.vcxproj, *.filters and Hazel.sln  that located in different directories.
I think keeping all this stuff together in separate dir(build or project) is more consistent way.
git status is clean and `rm -rf project` is easier for cleaning all stuff.

LovelySanta:
> now git shows us:
> `modified: Hazel/vendor/GLFW (untracked content) modified: Hazel/vendor/imgui (untracked content)` + a generated files like *.vcxproj, *.filters and Hazel.sln that located in different directories.
> I think keeping all this stuff together in separate dir(build or project) is more consistent way.
> git status is clean and `rm -rf project` is easier for cleaning all stuff.

The argument of moving the bin and bin-int out of the submodules to keep `git status` clean is invalid as the visual studio files (`*.vcxproj`, `*.vcxproj.filters`) files will always be in that directory anyway.

A beter solution for this is to add the bin; bin-int and vs-files to the .gitignore of the submodules. This way it should ignore those and won't show up as untracked content.

DaveAxiom:
The compiled objects in the submodules aren't relevant because they honestly don't need to be rebuilt. The hacking is done on Hazel itself. I'll give it that the file management is a little problematic since the Hazel repository puts premake files into the submodule directories!

halcyonx:
> The argument of moving the bin and bin-int out of the submodules to keep git status clean is invalid as the visual studio files (*.vcxproj, *.vcxproj.filters) files will always be in that directory anyway.
I don't think they should be tracked by git because it should be generated by premake. I thought it is not a bad idea to keep all generated and compiled stuff in some build dir, like in cmake: 
`mkdir build && cd build
cmake ..
`

CleverSource:
Should this have the label of "Awaiting Cherno's Approval"? Reason being is because I feel like removal of files or changing the directory requires an "executive" decision from him.

LovelySanta:
This requires the submodules to be merged first anyway... and only cherno can merge in submodules anyway...

LovelySanta:
With https://github.com/TheCherno/Hazel/pull/228 merged, the `git status` command no longer tells us that ImGui is modified. It was only spitting it out because the VS files were not ignored. As mentioned in #215, we can still think about generating all project files (maybe all except the solution file) in a separate build directory, which could also include the bin and bin-int directories. Make a separate PR for that.

PR #79, opened 2019-07-09
 Rename m_SceneData to s_SceneData

LovelySanta:
This to be inline with cherno's naming convention;
 - `s_` prefix for static fields;
 - `m_` prefix for member fields.


Gaztin:
LGTM

PR #81, opened 2019-07-10
Invert Camera's rotate direction.

3x3y3z3t:
We have invert the View Matrix but that inversion only affect the position (or move direction) of the object.
When rotate Camera in one direction, the rendered object should be rotate in opposite direction, or the camera rotation should be inverted as well.


LovelySanta:
Isn't it easier to just rotate the rotation axis to {0.0f, 0.0f, -1.0f} ?

3x3y3z3t:
Sorry, I forgot that positive angle increment is counter-clockwise. Thanks for pointing that out.
This PR can be closed.

PR #83, opened 2019-07-20
 Smoother project generation

LostMikely:
This new batch file will generate the project and if everything is alright the command line will close itself when it's finished. If there is any error, it will show the error message and then pause.

### DIFF
--- a/GenerateProjects.bat
+++ b/GenerateProjects.bat
@@ -1,5 +1,5 @@
 @echo off
-call vendor\bin\premake\premake5.exe vs2019
+call vendor\bin\premake\premake5.exe vs2017
 IF %ERRORLEVEL% NEQ 0 (
   PAUSE
 )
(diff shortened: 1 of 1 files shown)


WhoseTheNerd:
Batch files aren't cross-platform so I don't think that Cherno would accept it. Hazel was meant to be cross-platform.

LovelySanta:
This PR is just enhancing the batch file we already have.. I don't see why this is an issue, but I like to see all the projects it generated before manualy closing it. I think this is just a preference...

Gaztin:
I like this. It gives people who run the script directly from explorer a chance to catch errors.

@LovelySanta Since I usually have to re-generate the project files now and then, I always keep a terminal open so I can run my project generation and see the output after.

CleverSource:
Was it decided this would be merged? It is such a small change but I think it's a great thing, you don't need to worry about the output unless there is an error.

LGTM

PR #88, opened 2019-07-30
Fix error missing file (src/Hazel/Core/Timestep.h)

LovelySanta:
In the latest commits by Cherno (https://github.com/TheCherno/Hazel/commit/3352191dbcd5e18754be70b18a5e6744692a77ca), files where added about transforms. Part of this is the timestep class that is hidden in the Core folder. This file wasn't commited and so Hazel couldn't compile.

Took some commits to get my branch up to date, I suggest squash & merge into a single commit for merging instead.


LovelySanta:
Fixed in https://github.com/TheCherno/Hazel/commit/888b3e8f7b07d16c6a52be0e49bd2526271cab36

PR #91, opened 2019-08-04
ImGuiConsole

LovelySanta:
New feature that will come in handy in the future when we use docking in the Hazel development:

Since we're using docking (see vid where he shows the Hazel-dev) in the future. It is annoying that we have a debug console that has nothing to do with our application, so we're always stuck with two windows.

I've tapped into spdlog, the way spdlog works, is it creates sinks, where each sink is responsible to output to a single destination. Currently we're only using the `stdout_color_sink_mt` to output to the actual console. Now I've created a second sink, extending the base_sink class that allows me to create my custom sink. I've tapped this into ImGui so we can render the messages onto an ImGui window that we actualy can dock.

To keep performance up, I've limited the amount of messages stored in the sink to 200. This is a static variable that we can change if we want a longer history of logs. I've also implemented small basic features:

 - Auto scrolling to the bottom, so when a new log message appears, it scrolls to the bottom, just like the actual console does.
 - Setting to disable the auto scrolling, so we can scroll back if we whish to do so
 - Display filter; The sink stores 200 messages that are logged. The log level is defined in `log.cpp` and is set to trace, meaning all messages with level trace or higher will be logged in the sink. This only counts for future messages. However, in the `ImGuiLogSink` I've added an option to filter allmessages, meaning it will only display messages above the set level, in future and in past. It will still store the other messages (max 200) in the sink, but will display (less) messages on the screen.

In the current state, it looks something like:
image
As the picture shows, I've put the original console and then the ImGuiConsole both in the picture, they look exactly the same (except that the original console is at 200% UI, hench the reason it looks bigger). In the picture below, i've changed the display level to warnings, meaning it will hide the (green) info messages from the ImGuiConsole.
image

Any feedback and suggestions are welcome.
I think this is a neat addition to Hazel for future development

Kind regards
lovely_santa


LovelySanta:
Ideas:
 - Thinking of redoing the ownership of the `ImGuiLogSink`
    - Make it a singleton, so only one instance can be created
    - Create static function as get or create the instance
- Beter way of obtaining the ImGuiConsole instance out fo the Log/Sink class
    - Result of beter ownership of the `ImGuiLogSink`; cast the sink to ImGuiConsole
    - Since `ImGuiLogSink` is a singleton, ImGuiConsole is also a singleton
- Move log levels over to the log class for more generic use

DaveAxiom:
This is a new feature added to ImGui. It'd probably be best to contribute it to ImGui and use it downstream in Hazel as needed.

CleverSource:
LovelySanta isn't changing any of ImGui's internals, making it a feature that isn't added to ImGui. He has no reason to contribute it to ImGui as what he is doing is possible in ImGui since he uses ImGui to do it.

LovelySanta:
I am just using the ImGui API to render and using the spdlog API to 'sink' log messages as explained in the initial post. I'm simply linking 2 things together.

Sandbox/imgui.ini
```diff
@@ -9,10 +9,19 @@ Size=92,48
 Collapsed=0
```
WhoseTheNerd: ImGui preference files should not be tracked, in my opinion.

LovelySanta: Once we set up a docking viewports, we'll be commiting multiple ini files. This is just the default one for now. Rather have 1 'working' versus none at all

CleverSource:
I really hope this PR gets accepted. Plus it works just like the ImGui Demo where it can be toggled off. I hope Cherno can check this out soon.

LovelySanta:
It's not correct yet, flush is clearing it, the sink needs to hold a (small) internal buffer and flush when buffer full or when called explicitly to show it in the imgui console

drewerth:
Have you considered adding a text input towards the bottom of the console? A command system could be created sometime down the line to allow developers to quickly edit and configure their programs without needing bulky menus or a source recompile. I can understand the decision to refrain from adding one now given the lack of a need for commands though.

Something like the Valve Source engine developer console comes to mind as an example.

LovelySanta:
I didn't concider it yet, but it would be nice to have an option for it, even have an event that gets distributed maybe?

However, I might give the ImGuiConsole another write-over and add some optimalisations down the line..

drewerth:
Events are plausible. It'll be something to look at once this is implemented and see what works best for performance and usability. Don't want to tack too much into one Pull Request.

Good luck with the optimizations.

Hazel/src/Hazel/ImGui/ImGuiConsole.h
```diff
@@ -0,0 +1,65 @@
...
+	public:
+		class Message
+		{
+		public:
+			enum class Level : int8_t
+			{
+				Invalid = -1,
+				Trace = 0,
+				Debug = 1,
+				Info = 2,
```
lhurt51: Hey, Santa you should flip the debug and info enums so that Info is first and debug is second. This will make sure that your buttons and error levels function properly.

Hazel/src/Hazel/ImGui/ImGuiConsole.cpp
```diff
@@ -0,0 +1,249 @@
...
+	ImGuiConsole::Message::Color ImGuiConsole::Message::GetRenderColor(Level level)
+	{
+		switch (level)
+		{
+			case ImGuiConsole::Message::Level::Trace   : return { 0.75f, 0.75f, 0.75f, 1.00f }; // White-ish gray
+			case ImGuiConsole::Message::Level::Info    : return { 0.00f, 0.50f, 0.00f, 1.00f }; // Green
+			case ImGuiConsole::Message::Level::Debug   : return { 0.00f, 0.50f, 0.50f, 1.00f }; // Cyan
+			case ImGuiConsole::Message::Level::Warn    : return { 1.00f, 1.00f, 0.00f, 1.00f }; // Yellow
+			case ImGuiConsole::Message::Level::Error   : return { 1.00f, 0.00f, 0.00f, 1.00f }; // Red
+			case ImGuiConsole::Message::Level::Critical: return { 1.00f, 1.00f, 1.00f, 1.00f }; // White-white
```
lhurt51: Also, a suggestion for the color of critical messages, maybe try magenta which is = { 1.00f, 0.00f, 1.00f, 1.00f }, just so the user can see the message more clearly.

Hazel/src/hzpch.h
```diff
@@ -2,12 +2,14 @@
 
 #include <iostream>
 #include <memory>
+#include <mutex>
```
ReDucTor: Is mutex going to be used everywhere? (I hope not), this just seems like its bloating the pch that is already extremely bloated

Hazel/src/Hazel/Log.h
```diff
@@ -4,6 +4,8 @@
 #include "spdlog/spdlog.h"
 #include "spdlog/fmt/ostr.h"
 
+#include "ImGui/ImGuiConsole.h"
```
ReDucTor: Why does this include the ImGuiConsole here? Is that necessary for this header file?

Hazel/src/Hazel/Log.cpp
```diff
@@ -10,11 +11,20 @@ namespace Hazel {
 
 	void Log::Init()
 	{
+		// create the sinks
+		std::vector<spdlog::sink_ptr> sinks;
+		sinks.emplace_back(std::make_shared<spdlog::sinks::stdout_color_sink_mt>()); // VS debug console
```
ReDucTor: Isn't this stdout not visual studio debug console?

Hazel/src/Hazel/ImGui/ImGuiConsoleSink.h
```diff
@@ -0,0 +1,62 @@
...
+	private:
+		uint8_t m_MessagesBuffered = 0;
+		uint8_t m_MessageBufferCapacity;
+		std::vector<Ref<ImGuiConsole::Message>> m_MessageBuffer;
+	};
+
+}
+
+#include "spdlog/details/null_mutex.h"
+#include <mutex>
```
ReDucTor: mid-file includes are bad because they can easily end up inside a namespace then you'll get weird issues, also is there any need for the single threaded one?

Hazel/src/Hazel/Log.cpp
```diff
@@ -10,11 +11,20 @@ namespace Hazel {
...
+		sinks.emplace_back(std::make_shared<ImGuiConsoleSink_mt>(true)); // ImGuiConsole
```
ReDucTor: Can this be added before ImGui is initialized?

Hazel/src/Hazel/ImGui/ImGuiLayer.h
```diff
@@ -6,12 +6,14 @@
...
 
+#include "ImGuiConsole.h"
+
 namespace Hazel {
 
 	class HAZEL_API ImGuiLayer : public Layer
 	{
 	public:
-		ImGuiLayer();
+		ImGuiLayer(bool renderDemo = true, bool renderConsole = true);
```
ReDucTor: One thing to be careful with adding a bunch of boolean variables like this is that if someone adds another one or reorders them then they will be the incorrect boolean combination with no warnings or errors

Hazel/src/Hazel/ImGui/ImGuiConsoleSink.h
```diff
@@ -0,0 +1,62 @@
+#include "spdlog/sinks/base_sink.h"
+#include "ImGuiConsole.h"
+
+namespace Hazel {
+
+	template<class Mutex>
```
ReDucTor: Mutex is unused here

Hazel/src/Hazel/ImGui/ImGuiConsoleSink.h
```diff
@@ -0,0 +1,62 @@
...
+	class ImGuiConsoleSink : public spdlog::sinks::base_sink<std::mutex>
+	{
+	public:
+		explicit ImGuiConsoleSink(bool forceFlush = false, uint8_t bufferCapacity = 10)
+			: m_MessageBufferCapacity(forceFlush ? 1 : bufferCapacity), m_MessageBuffer(std::vector<Ref<ImGuiConsole::Message>>(forceFlush ? 1 : bufferCapacity))
```
ReDucTor: When would you want this to buffer? This flushing mechanism seems unusual there is no disk io or anything, its just a queue of messages which will be shown to the user and displayed within the draw tick

Hazel/src/Hazel/ImGui/ImGuiConsoleSink.h
```diff
@@ -0,0 +1,62 @@
...
+		ImGuiConsoleSink(const ImGuiConsoleSink&) = delete;
+		ImGuiConsoleSink& operator=(const ImGuiConsoleSink&) = delete;
+		virtual ~ImGuiConsoleSink() = default;
+	protected:
+		void sink_it_(const spdlog::details::log_msg& msg) override
+		{
+			fmt::memory_buffer formatted;
+			sink::formatter_->format(msg, formatted);
+
+			*(m_MessageBuffer.begin() + m_MessagesBuffered) = std::make_shared<ImGuiConsole::Message>(fmt::to_string(formatted), GetMessageLevel(msg.level));
```
ReDucTor: this could have just been `m_MessageBuffer[m_MessagesBuffered] = ...`

Hazel/src/Hazel/ImGui/ImGuiConsoleSink.h
```diff
@@ -0,0 +1,62 @@
...
+			if (++m_MessagesBuffered == m_MessageBufferCapacity)
+				flush_();
+		}
+
+		void flush_() override
```
ReDucTor: This seems like its going to keep the old messages around even after they are unused

Hazel/src/Hazel/ImGui/ImGuiConsole.cpp
```diff
@@ -0,0 +1,249 @@
...
+		ImGuiConsole::Message::Level::Trace,
+		//ImGuiConsole::Message::Level::Debug,
+		ImGuiConsole::Message::Level::Info,
+		ImGuiConsole::Message::Level::Warn,
+		ImGuiConsole::Message::Level::Error,
+		ImGuiConsole::Message::Level::Critical,
+		ImGuiConsole::Message::Level::Off
+	};
+
+	ImGuiConsole::Message::Message(const std::string message, Level level)
```
ReDucTor: Was this meant to be 'const std::string &' ? This potentially should a standard std::string then you can std::move it into m_Message

Hazel/src/Hazel/ImGui/ImGuiConsole.h
```diff
@@ -0,0 +1,65 @@
...
+				Warn = 3,
+				Error = 4,
+				Critical = 5,
+				Off = 6, // Display nothing
+			};
+		private:
+			struct Color { float r, g, b, a; };
+		public:
+			Message(const std::string message = "", Level level = Level::Invalid);
```
ReDucTor: Why does this need a default value for message and level? Are they ever unused?

Hazel/src/Hazel/ImGui/ImGuiConsole.cpp
```diff
@@ -0,0 +1,249 @@
...
+			if (s_RequestScrollToBottom && ImGui::GetScrollMaxY() > 0)
+			{
+				ImGui::SetScrollY(ImGui::GetScrollMaxY());
+				s_RequestScrollToBottom = false;
+			}
+		}
+		ImGui::EndChild();
+	}
+
+	std::vector<ImGuiConsole::Message::Level> ImGuiConsole::Message::s_Levels{
```
ReDucTor: Is this necessary isn't this just a list of values from the enum?

Hazel/src/Hazel/ImGui/ImGuiConsole.cpp
```diff
@@ -0,0 +1,249 @@
+#include "hzpch.h"
+#include "ImGuiConsole.h"
+
+#include "imgui.h"
+#include "examples/imgui_impl_glfw.h"
+#include "examples/imgui_impl_opengl3.h"
+
+namespace Hazel {
+
+	float ImGuiConsole::s_DisplayScale = 1.0f;
```
ReDucTor: Alot of these static members seem like they should be actual members of 'ImGuiConsole' and 'ImGuiConsole' should be the spdlog sink

Hazel/src/Hazel/ImGui/ImGuiConsole.cpp
```diff
@@ -0,0 +1,249 @@
...
+	uint16_t ImGuiConsole::s_MessageBufferCapacity = 200;
+	uint16_t ImGuiConsole::s_MessageBufferSize = 0;
+	uint16_t ImGuiConsole::s_MessageBufferBegin = 0;
+	ImGuiConsole::Message::Level ImGuiConsole::s_MessageBufferRenderFilter = ImGuiConsole::Message::Level::Trace;
+	std::vector<Ref<ImGuiConsole::Message>> ImGuiConsole::s_MessageBuffer(s_MessageBufferCapacity);
+	bool ImGuiConsole::s_AllowScrollingToBottom = true;
+	bool ImGuiConsole::s_RequestScrollToBottom = false;
+
+	void ImGuiConsole::AddMessage(Ref<Message> message)
```
ReDucTor: Why make this a ref which has shared ownership? It should be able to just be a message, if you construct the message or move it into here, then s_Message buffer can be a vector of Messages and you avoid additional indirection and unclear ownership of messages

Hazel/src/Hazel/ImGui/ImGuiConsole.cpp
```diff
@@ -0,0 +1,249 @@
...
+	{
+		if (message->m_Level == Message::Level::Invalid)
+			return;
+
+		*(s_MessageBuffer.begin() + s_MessageBufferBegin) = message;
```
ReDucTor: This overriding behaviour seems very suspect, you should be able to just have a vector and push/emplace back, if you want to pre-reserve space you can use vector.reserve(), doing this is just going to keep additional memory in-use until its overwritten which just increase memory fragmentation.

ReDucTor: oops, I see its an attempt at a circular buffer, in which case you can still just do 
```
s_MessageBuffer[s_MessageBufferBegin++] = message;
if (s_MessageBufferBegin == s_MessageBufferCapacity)
  s_MessageBufferBegin = 0;
```

Hazel/src/Hazel/ImGui/ImGuiConsole.cpp
```diff
@@ -0,0 +1,249 @@
...
+		if (++s_MessageBufferBegin == s_MessageBufferCapacity)
+			s_MessageBufferBegin = 0;
+		if (s_MessageBufferSize < s_MessageBufferCapacity)
```
ReDucTor: This seems unused and unnecessary

Hazel/src/Hazel/ImGui/ImGuiConsole.cpp
```diff
@@ -0,0 +1,249 @@
...
+			(*message) = std::make_shared<Message>();
+		s_MessageBufferBegin = 0;
+	}
+
+	void ImGuiConsole::OnImGuiRender(bool* show)
+	{
+		ImGui::SetNextWindowSize(ImVec2(640, 480), ImGuiCond_FirstUseEver);
+		ImGui::Begin("Console", show);
+		{
+			ImGuiRendering::ImGuiRenderHeader();
```
ReDucTor: One thing to be careful when naming things similar to the API for ImGui is that a reader could misunderstand and think this is an ImGui API when instead its part of the console logging API

Hazel/src/Hazel/ImGui/ImGuiConsole.cpp
```diff
@@ -0,0 +1,249 @@
...
+		ImGui::SameLine(0.0f, spacing);
+
+		// Dropdown with levels
+		ImGui::PushItemWidth(ImGui::CalcTextSize("Critical").x * 1.36f);
+		if (ImGui::BeginCombo(
+			"##MessageRenderFilter",
+			Message::GetLevelName(s_MessageBufferRenderFilter),
+			ImGuiComboFlags_NoArrowButton))
+		{
+			for (auto level = Message::s_Levels.begin(); level != Message::s_Levels.end(); level++)
```
ReDucTor: You can count from trace to off here (i'd probably add duplicate values to the enum which are 'first' and 'last' indicators instead)

Hazel/src/Hazel/ImGui/ImGuiConsole.cpp
```diff
@@ -0,0 +1,249 @@
...
+		// Text change level
+		ImGui::AlignFirstTextHeightToWidgets();
+		ImGui::Text("Display");
+
+		ImGui::SameLine(0.0f, 2.0f * spacing);
+
+		// Buttons to quickly change level left
+		if (ImGui::ArrowButton("##MessageRenderFilter_L", ImGuiDir_Left))
+		{
+			s_MessageBufferRenderFilter = Message::GetLowerLevel(s_MessageBufferRenderFilter);
```
ReDucTor: Can't you just increment decrement this value as an integer and set a min on it?

ReDucTor: Same with the upper level one

Hazel/src/Hazel/ImGui/ImGuiConsole.h
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Hazel {
+
+	class ImGuiConsole
+	{
+	public:
+		class Message
+		{
+		public:
+			enum class Level : int8_t
```
ReDucTor: Why not leave this as default? Smaller integer types can come at a cost because the compiler then needs to ensure they are within that range

Hazel/src/Hazel/ImGui/ImGuiConsole.cpp
```diff
@@ -0,0 +1,249 @@
+#include "hzpch.h"
+#include "ImGuiConsole.h"
+
+#include "imgui.h"
+#include "examples/imgui_impl_glfw.h"
```
ReDucTor: Does this really need the impl files?

Hazel/src/Hazel/ImGui/ImGuiLayer.cpp
```diff
@@ -6,15 +6,16 @@
 #include "examples/imgui_impl_opengl3.h"
 
 #include "Hazel/Application.h"
+#include "Hazel\Log.h"
```
GeorgeWeb: I suppose that's Visual Studio's auto-completion fault but consider avoiding Windows-specific use of backslashes for paths. `Hazel\Log.h` -> `Hazel/Log.h`

LovelySanta:
Closing this as we will not need this any time soon. I'll leave the branch on my fork and rework this when we actually have our Hazelnut editor in place.

PR #92, opened 2019-08-04
Fix warning casting from 'const uint32_t' to 'const void*'

LovelySanta:
Fixed that lonely warning as it was working on my nerves...
image

Fixed by first casting the integer to a pointer before casting to the void pointer.

Kind regards
lovely_santa


KeithDobbelaere:
I was just going to suggest first casting this to a size_t to get rid of that warning, but then I see you beat me to it.  Either works though.

Gaztin:
LGTM

PR #93, opened 2019-08-04
Fixed a bug in LayerStack which would occur if passing an overlay or …

### LINKED ISSUE #87: LayerStack::PopLayer is incorrect
In LayerStack::PopLayer, a comparison is made as

if (it != m_Layers.end())

this should, however, be:

if (it != m_Layers.begin() + m_LayerInsertIndex)

otherwise, if the function is passed an overlay layer, it will happily remove and shift the last non-overlay layer to be an overlay when it decrements m_LayerInsertIndex.

jdunlap:
This PR fixes an issue that would occur when passing either an overlay or a non-existent layer in to PopLayer in LayerStack.  For the Find, it stops at the correct spot, but then compares the iterator to end(), which is incorrect when searching for a layer.

This PR resolves #87

### DIFF
--- a/Hazel/src/Hazel/LayerStack.cpp
+++ b/Hazel/src/Hazel/LayerStack.cpp
@@ -29,7 +29,7 @@ namespace Hazel {
 	void LayerStack::PopLayer(Layer* layer)
 	{
 		auto it = std::find(m_Layers.begin(), m_Layers.begin() + m_LayerInsertIndex, layer);
-		if (it != m_Layers.end())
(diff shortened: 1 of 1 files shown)


LovelySanta:
Perfect!

Small feedback: Please edit the initial post and use [any closing keywords](https://help.github.com/en/articles/closing-issues-using-keywords) such as `resolves #87` so it will close the issue when this PR gets committed.

Kind regards
lovely_santa

PR #102, opened 2019-08-24
Fix for bug in Vertex Array

### LINKED ISSUE #100: OpenGLVertexArray not adding Vertex Buffers correctly
When adding more than one vertex buffer, the index does not account for the other vertex buffers. I'm probably not explaining this very well but this code should make more sense:
```
                int offset = 0;
		for (auto a : m_VertexBuffers)
		{
			offset += a->GetLayout().GetElements().size();
		}
		const BufferLayout& layout = buffer->GetLayout();
		for (int i = 0; i < layout.GetElements().size(); i++)
		{
			auto& element = layout.GetElements()[i];
			glEnableVertexAttribArray(i + offset);
			glVertexAttribPointer(i + offset, element.GetComponentCount(), GL_FLOAT, element.Normalized ? GL_TRUE : GL_FALSE, layout.GetStride(), (const void*)element.Offset);
		}
```

kcourtney77:
There was an issue where a Vertex Array only worked with one Vertex Buffer. resolves #100

### DIFF
--- a/Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
+++ b/Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
@@ -26,7 +26,7 @@ namespace Hazel {
 		return 0;
 	}
 
-	OpenGLVertexArray::OpenGLVertexArray()
+	OpenGLVertexArray::OpenGLVertexArray() : m_IndexOffset(0)
 	{
 		glCreateVertexArrays(1, &m_RendererID);
 	}
@@ -54,16 +54,12 @@ namespace Hazel {
 		vertexBuffer->Bind();
 
 		uint32_t index = 0;
-		uint32_t count = 0;
-		for(auto a : m_VertexBuffers)
-		{
-			count += a->GetLayout().GetElements().size();
-		}
+	
 		const auto& layout = vertexBuffer->GetLayout();
 		for (const auto& element : layout)
 		{
-			glEnableVertexAttribArray(index + count);
-			glVertexAttribPointer(index + count,
+			glEnableVertexAttribArray(index + m_IndexOffset);
+			glVertexAttribPointer(index + m_IndexOffset,
 				element.GetComponentCount(),
 				ShaderDataTypeToOpenGLBaseType(element.Type),
 				element.Normalized ? GL_TRUE : GL_FALSE,
@@ -73,6 +69,7 @@ namespace Hazel {
 		}
 
 		m_VertexBuffers.push_back(vertexBuffer);
+		m_IndexOffset += vertexBuffer->GetLayout().GetElements().size();
 	}
 
 	void OpenGLVertexArray::SetIndexBuffer(const std::shared_ptr<IndexBuffer>& indexBuffer)
--- a/Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
+++ b/Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
@@ -54,7 +54,6 @@ namespace Hazel {
 		vertexBuffer->Bind();
 
 		uint32_t index = 0;
-	
 		const auto& layout = vertexBuffer->GetLayout();
 		for (const auto& element : layout)
 		{
--- a/Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
+++ b/Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
@@ -26,7 +26,7 @@ namespace Hazel {
 		return 0;
 	}
 
-	OpenGLVertexArray::OpenGLVertexArray() : m_IndexOffset(0)
+	OpenGLVertexArray::OpenGLVertexArray()
 	{
 		glCreateVertexArrays(1, &m_RendererID);
 	}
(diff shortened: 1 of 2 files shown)


CleverSource:
You should add a comment saying "resolves #100" so it automatically closes that issue when this is merged (if it gets merged).

I recommend just editing your OP (original post)

Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
```diff
@@ -54,11 +54,16 @@ namespace Hazel {
 		vertexBuffer->Bind();
 
 		uint32_t index = 0;
+		uint32_t count = 0;
+		for(auto a : m_VertexBuffers)
```
LovelySanta: Mostly coding style review here:
1. Might be best to store a referece
    `auto& a`
2. Can be const
    `cont auto& a`
3. Use decent names, for example
    `const auto& vertexBuffer : m_VertexBuffers`

Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
```diff
@@ -54,11 +54,16 @@ namespace Hazel {
 		vertexBuffer->Bind();
 
 		uint32_t index = 0;
+		uint32_t count = 0;
```
LovelySanta: Cherno is quite delicate about using the words size and count.
Maybe something like `indexOffset` or `elementOffset`?

I am not a huge fan of re-calculating this every time we add a vertexBuffer, especialy when add alot of them...
Shouldn't we store this as a member and update the value after adding it to our buffers?
https://github.com/TheCherno/Hazel/blob/f52b813db65cd7275da56003c501914ad1c7515a/Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp#L70

CleverSource: Definitely agreeing with this one. Code review is good as well.

Hazel/src/Platform/OpenGL/OpenGLVertexArray.h
```diff
@@ -20,6 +20,7 @@ namespace Hazel {
 		virtual const std::shared_ptr<IndexBuffer>& GetIndexBuffer() const { return m_IndexBuffer; }
 	private:
 		uint32_t m_RendererID;
+		uint32_t m_IndexOffset;
```
CleverSource: I feel like this should have a different name such as ``m_VertexBufferIndexOffset``. Generally it should be very clear what it's used for if you were only looking at the header file.

Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
```diff
@@ -26,7 +26,7 @@ namespace Hazel {
 		return 0;
 	}
 
-	OpenGLVertexArray::OpenGLVertexArray()
+	OpenGLVertexArray::OpenGLVertexArray() : m_IndexOffset(0)
```
CleverSource: Also as general code review; I feel like this should become:
```cpp
OpenGLVertexArray::OpenGLVertexArray()
    : m_IndexOffset(0)
{
    glCreateVertexArrays(1, &m_RendererID);
}
```
mainly because Cherno prefers doing this and it would be best to stay consistent.

While this probably isn't needed it still would be nice to stay consistent with other code.

LovelySanta: I'm debating myself if this is even necessary at all, as it doesn't require constructor arguments to initialize that variable, just setting =0 in the header file should have the same effect?

CleverSource: That is true, either one of them would work but I think it would be best to set it to 0 within the header file instead. @kcourtney77 could've done this for a reason though, so clarification would be nice. I'm pretty sure they did it based on preference though.

kcourtney77: Yeah I do that just as a preference. I could definitely change it if you think I should.

LovelySanta: If you leave it as an initializer list, for sure put it on a new line to be inline with the coding style (pun not intended). I've seen cherno do both header and initialiser list, so that's totaly up to you

CleverSource: I think you should put it on a new line like my original post states but since there is no performance difference with declaring it directly on the header file (doing: `` = 0`` like Lovely stated) or doing ``: m_IndexOffset(0)`` either one of them would work. If you want to stay perfectly consistent with Cherno's style it would probably be better to do it via header but it wouldn't matter at all; unless Lovely thinks it would matter (as he is a collaborator) then you probably can keep it the same and just doing what I said.

**tl;dr** It's fine for you to keep it like that (unless Lovely says otherwise) but you should put it on a new line like my original review states.

LovelySanta: I was about to find an example of both cases, but for a constant value, I only found it defined in the header file. In all other cases I find initializer lists always depending on parameters passed to the constructor...

https://github.com/TheCherno/Hazel/blob/f52b813db65cd7275da56003c501914ad1c7515a/Hazel/src/Hazel/LayerStack.h#L25

If I would go back to the argument about the header file making stuff clear, I'm more thinking about putting it in the header... But whatever you choose I'll accept, as it doesn't impact performance in any way...

CleverSource:
With new changes it looks good for merging. Just up to Lovely or Gaztin I believe.

Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
```diff
@@ -54,11 +54,12 @@ namespace Hazel {
 		vertexBuffer->Bind();
 
 		uint32_t index = 0;
+
```
LovelySanta: Last one, remove this empty line, wasn't there before (just looked at the diff), but looks indeed good to merge!

LovelySanta:
LGTM

PR #105, opened 2019-08-24
Fix a warning introduced by PR

CleverSource:
Fix warning: '+=': conversion from 'size_t' to 'uint32_t', possible loss of data

Fixed by casting to ``uint32_t``.
image

### DIFF
--- a/Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
+++ b/Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
@@ -53,22 +53,20 @@ namespace Hazel {
 		glBindVertexArray(m_RendererID);
 		vertexBuffer->Bind();
 
-		uint32_t index = 0;
 		const auto& layout = vertexBuffer->GetLayout();
 		for (const auto& element : layout)
 		{
-			glEnableVertexAttribArray(index + m_VertexBufferIndexOffset);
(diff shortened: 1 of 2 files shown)


LovelySanta:
Convo happened in discord. Small summary what happens (after #102):
- `layout.GetElements().size()` is a `size_t`, and we're incrementing `m_VertexBufferIndexOffset `, which is a `uint32_t`, hence the warning.
- After iterating the `for element : layout`, index equals this size. Index is already a uint32_t so we could just use this so we don't have to cast
- Finaly we noticed that we can just directly increment `m_VertexBufferIndexOffset` directly instead of `index`, hence the reason to rename it to `m_VertexBufferIndex`.

LGTM

LovelySanta:
Tested by splitting the SquareVertexBuffer into one for the vertex positions and one for the texture coordinates and adding both vertexbuffers into one vertex array.
```cpp
        float squareVertices[3 * 4] = {
            -0.5f, -0.5f, 0.0f,
             0.5f, -0.5f, 0.0f,
             0.5f,  0.5f, 0.0f,
            -0.5f,  0.5f, 0.0f
        };

        float texCoordStuff[2 * 4] = {
            0.0f, 0.0f,
            1.0f, 0.0f,
            1.0f, 1.0f,
            0.0f, 1.0f
        };
```

PR #106, opened 2019-08-25
Avoid using std::function for performance

### LINKED ISSUE #85: Avoid using std::function in event dispatcher for performance
Currently Hazel is using std::function as the parameter to event dispatchers.
https://github.com/TheCherno/Hazel/blob/5dc2ee187fae6a3efa1d13b61faf5ca67e3638bd/Hazel/src/Hazel/Events/Event.h#L64-L73
But as we all know, std::function uses type erasure under the hood, which renders this approach inefficient for high performance needs:

1. The size of a class (in this case `std::function<R(Ts...)>`) is fixed at compile time, but since any kind of functor with the corresponding signature might be used to construct the std::function object, dynamic allocations may happen. Furthermore, if the functor passed is big (e.g. a lambda with several captures by value), it also needs to copy a large chunk of memory. But all these aren't necessary.

2. Modern compilers can be really good at inlining trivial function calls, like when we pass a rather simple lambda to a consumer function which calls the lambda in some way, the call might be inlined thus done without overhead. But with the type erasure mechanism used by `std::function`, the compiler in most cases cannot make sense of all these weird things thus it cannot do inlining, which makes the case worse.

Therefore I propose a simple solution to this problem, just template on the type of the functor.

```cpp
template <typename T, typename F>
bool Dispatch(const F& func)
{
    if (m_Event.GetEventType() == T::GetStaticType())
    {
        // BTW you can just simply cast the event reference to T&, no need for the addressing and dereferencing
        m_Event.Handled = func(static_cast<T&>(m_Event));  
        return true;
    }
    return false;
}
```

The `F` parameter can be deduced by the compiler and the user passes in `T` anyways.

Plus, for a better user interface, even the parameter `T` can be automatically deduced using some tricks:

```cpp
namespace detail
{
    // These thing can get the parameter type T out of the lambda type F
    // The following two functions don't need to be implemented since only the function signatures are used
    template <typename F, typename T>
    T dispatch_helper(bool (F::*)(T&));
    template <typename F, typename T>
    T dispatch_helper(bool (F::*)(T&) const);
    template <typename F>
    using DispatchEventType = decltype(dispatch_helper(&F::operator()));
}
...
// The second template parameter SFINAE's out lambdas that are not event handlers
// So if you pass in just a random functor it will be a compile error, which is neat
template <typename F, typename = decltype(detail::DispatchEventType<F>::GetStaticType())>
bool Dispatch(F&& func)
{
    using T = detail::DispatchEventType<F>;  // Use the helper to get the parameter type
    if (m_Event.GetEventType() == T::GetStaticType())
    {
        m_Event.Handled = std::forward<F>(func)(static_cast<T&>(m_Event));
        return true;
    }
    return false;
}
```

And now everything is solved!

CleverSource:
Avoiding the usage of std::function in event dispatcher for performance. This has an 20%+ improvement in speed. You can read more about this via issue #85. 

Full credit to @Chlorie for this.

This PR was never made so I've decided to make it.
This resolves #85.

### DIFF
--- a/Hazel/src/Hazel/Events/Event.h
+++ b/Hazel/src/Hazel/Events/Event.h
@@ -58,7 +58,8 @@ namespace Hazel {
 			: m_Event(event)
 		{
 		}
-
+		
+		// F will be deduced by the compiler
 		template<typename T, typename F>
 		bool Dispatch(const F& func)
 		{
--- a/Hazel/src/Hazel/Events/Event.h
+++ b/Hazel/src/Hazel/Events/Event.h
@@ -53,20 +53,18 @@ namespace Hazel {
 
 	class EventDispatcher
 	{
-		template<typename T>
-		using EventFn = std::function<bool(T&)>;
 	public:
 		EventDispatcher(Event& event)
 			: m_Event(event)
 		{
 		}
 
-		template<typename T>
-		bool Dispatch(EventFn<T> func)
+		template<typename T, typename F>
+		bool Dispatch(const F& func)
 		{
 			if (m_Event.GetEventType() == T::GetStaticType())
 			{
-				m_Event.Handled = func(*(T*)&m_Event);
+				m_Event.Handled = func(static_cast<T&>(m_Event));
 				return true;
 			}
 			return false;


LovelySanta:
You still have to deduce the F parameter, so we can just call it with a single template parameter as before? I expect it not to work as of now, since dispatch has 2 template parameters, where in all the code we only provide one

CleverSource:
> You still have to deduce the F parameter, so we can just call it with a single template parameter as before? I expect it not to work as of now, since dispatch has 2 template parameters, where in all the code we only provide one

I've tested this on a local copy of Hazel, I directly cloned the repo and just changed that one thing before making this PR. The data that the 20% improvement comes from is from someone using the code I've changed.
image

LovelySanta:
> I've tested this on a local copy of Hazel, I directly cloned the repo and just changed that one thing before making this PR. The data that the 20% improvement comes from is from someone using the code I've changed.

I am aware from where the improvement comes from, but I am talking about this
image
The template has now 2 parameters, while in every dispatcher, we're only providing 1 template parameter where the 2nd one doesn't have some default...
That's where the whole deduction in the rest of the post comes in... Or am I interpreting something wrong?

CleverSource:
Odd.. the person that says it's a 20% improvement did not test with the deduction piece. Which is why I don't think that matters.

LovelySanta:
Maybe @Chlorie can help us implement it, as we both hit a solid brick wall on trying to get the deduction working...

Chlorie:
I was busy doing something else recently so I didn't respond to this pull request, my fault.
The solution for template argument deduction mentioned in the original issue was kind of incomplete, now that you've brought up this question, I'm thinking of a more complete solution to that.

Chlorie:
After some trial and error I found that it's really hard to do the full type deduction (let the compiler to deduce T) when using generic function objects (function pointers are easy, non-generic lambdas are easy, but generic function objects like the return value of std::bind are hard). If this is not really needed, like what @ilikenoodles2 said, the solution in this PR is good enough.
I'll try to play with these templates a little more to see whether the full deduction can be achieved, but not with much hope.

LovelySanta:
Now that I look at it, it might indeed be more readable to supply T. I'll approve as it is.

Hazel/src/Hazel/Events/Event.h
```diff
@@ -53,20 +53,18 @@ namespace Hazel {
...
 	public:
 		EventDispatcher(Event& event)
 			: m_Event(event)
 		{
 		}
 
-		template<typename T>
-		bool Dispatch(EventFn<T> func)
+		template<typename T, typename F>
+		bool Dispatch(const F& func)
```
LovelySanta: I would add a small comment (maybe even with an example) mensioning F will be deduced by the compiler, as this is something that is visible for users of Hazel.

LovelySanta:
LGTM

Issue #75, opened 2019-07-05
log the OpenGL errors


LovelySanta:
As mensioned in discord, currently we're not logging any openGL errors that could possible occur. To do this, I would recommend creating the same macro as Cherno did in [his OpenGL series](https://www.youtube.com/watch?v=FBbPWSOQ0-w&list=PLlrATfBNZ98foTJPJ_Ev03o2oq3-GGOS2&index=10).

### Step 1: Creating the macro
In [src/Platform/OpenGL/](https://github.com/TheCherno/Hazel/tree/master/Hazel/src/Platform/OpenGL) create a new file `OpenGLMacro.h` where the macro is defined. We do it in the OpenGL folder as it is OpenGL specific:
```cpp
// OpenGLMacro.h
#pragma once
#include "Engine/Core.h"
#include "Engine/Log.h"

#include "glad/glad.h"

inline static void GLClearError() { while (glGetError() != GL_NO_ERROR); }

static bool GLLogCall(const char* function, const char* file, const int line)
{
    while (GLenum error = glGetError())
    {
        ENGINE_CORE_ERROR("[OpenGL] Error ({0}):", error);
        ENGINE_CORE_ERROR("[OpenGL]     in function {0}", function);
        ENGINE_CORE_ERROR("[OpenGL]     at {0} ({1})", file, line);
        return false;
    }
    return true;
}

#if defined ENGINE_DEBUG
    #define GLCall(x) GLClearError();\
                      x;\
                      ENGINE_CORE_ASSERT(GLLogCall(#x, __FILE__, __LINE__), "OpenGL Error!")
#else
    #define GLCall(x) x
#endif
```

As described in Cherno's video, the macro wraps two functions around the actual call:

- `GLClearError()` before the function call to clear all previous errors
- `GLLogCall(...)` after the function to log any errors.

_**NOTE:**_ `ENGINE_CORE_ASSERT` currently does not call the function outside the debug configuration, which is here not an issue as the GLCall macro is only defined inside the debug configuration.

When `ENGINE_DEBUG` is not defined, GLCall will just call the function without any other fuctions wrapped around it.

### Step 2: Using the macro
In every OpenGl cpp file (inside [src/Platform/OpenGL/](https://github.com/TheCherno/Hazel/tree/master/Hazel/src/Platform/OpenGL)) we include the OpenGLMacro.h and add `GLCall()` around each raw OpenGL call.

[...]

Issue #67, opened 2019-06-11
glCreateBuffers and OpenGL 4.4


Dressingoak:
Hi,
We should make it clear what OpenGL copatability the project should have. The moment we changed `glGenBuffers` to `glCreateBuffers` things went south (for me) with the exception:
```
Exception thrown at 0x0000000000000000 in Sandbox.exe: 0xC0000005: Access violation executing location 0x0000000000000000.
```
Some quick research suggested that it's sue to me running OpenGL 4.4 on my laptop. This of course means that the solution does not run out-of-the-box in this case, and based on what I've read, it's due to `GL_ARB_direct_state_access` entering the core profile only as of 4.5. So:

- Are there any reasons for not sticking with `glGenBuffers` ?
- Might it be worth differentiating the implementation based on the GL version?
- Should anything below 4.5 be unsupported all together?

WillianKoessler:
"Are there any reasons for not sticking with `glGenBuffers` ?"
I believe that he used `glCreateBuffers `just to match the same function names for generating buffers and shader. As he shown, `glCreateBuffers `does not bind the buffers as he expected.
The difference between glCreate and glGen is that glCreate returns buffers that are completely new and unused, unlike glGen, which reuse buffer names.
Somebody please correct me if i'm wrong, but there is no practical difference between glGen and glCreate


"Might it be worth differentiating the implementation based on the GL version?"
"Should anything below 4.5 be unsupported all together?"
I really don't see a reason for that. You could base your build on the GL version, but I don't think that is really necessary to limit the build to a GL version so recent. Older versions still working pretty fine

AHUMANTHATISHUMAN:
`glCreateBuffers` is useful when working with other functions in `GL_ARB_direct_state_access`, such as `glBindNamedBuffer`. Since it's not being used at the moment, there is no difference. See #66.

LovelySanta:
This was fixed in #137

PR #107, opened 2019-08-26
Change bind target inside OpenGLIndexBuffer constructor

### LINKED ISSUE #95: ELEMENT_ARRAY_BUFFER can't be bound while current VBO is 0
According to [Khronos Wiki](https://www.khronos.org/opengl/wiki/Vertex_Specification#Vertex_Array_Object),

>The core OpenGL profile makes VAO object 0 not an object at all. So if VAO 0 is bound in the core profile, you should not call any function that modifies VAO state. **This includes binding the `GL_ELEMENT_ARRAY_BUFFER` with glBindBuffer.**

`OpenGLIndexBuffer` constructor doesn't check that currently bound VAO is not 0.

rijenkii:
According to [OpenGL 4.6 Core Profile Specification](https://www.khronos.org/registry/OpenGL/specs/gl/glspec46.core.pdf) and [Khronos Wiki](https://www.khronos.org/opengl/wiki/Vertex_Specification#Vertex_Array_Object),

> [it is invalid to] modify or query buffer object state for a target bound to zero, since there is no buffer object corresponding to the name zero.
*(OpenGL Specification, page 62)*

> The core OpenGL profile makes VAO object 0 not an object at all. So if VAO 0 is bound in the core profile, you should not call any function that modifies VAO state.
*(Khronos Wiki, last paragraph of Vertex Array Object)*

Element array buffer binding is a part of VAO state *(OpenGL Specification, page 586)*.

This PR modifies the OpenGLIndexBuffer constructor, so that index buffer is initialized while bound to `GL_ARRAY_BUFFER`.

Closes #95

### DIFF
--- a/Hazel/src/Platform/OpenGL/OpenGLBuffer.cpp
+++ b/Hazel/src/Platform/OpenGL/OpenGLBuffer.cpp
@@ -39,8 +39,8 @@ namespace Hazel {
 		: m_Count(count)
 	{
 		glCreateBuffers(1, &m_RendererID);
-		glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, m_RendererID);
-		glBufferData(GL_ELEMENT_ARRAY_BUFFER, count * sizeof(uint32_t), indices, GL_STATIC_DRAW);
+		glBindBuffer(GL_ARRAY_BUFFER, m_RendererID);
+		glBufferData(GL_ARRAY_BUFFER, count * sizeof(uint32_t), indices, GL_STATIC_DRAW);
 	}
 
 	OpenGLIndexBuffer::~OpenGLIndexBuffer()


LovelySanta:
Just to mension, we're wanting to use OpenGL 4.5 core and not 4.6, but in [the specifications](https://www.khronos.org/registry/OpenGL/specs/gl/glspec45.core.pdf) it can be found at page 60.

So it is correct to 'just bind' to `GL_ARRAY_BUFFER` like we present any 'other' buffer, like done with the vertex buffer.
https://github.com/TheCherno/Hazel/blob/24fb3bbacc3661edb7ae4a5c9d946f948b9dcecd/Hazel/src/Platform/OpenGL/OpenGLBuffer.cpp#L12-L17

The only place where it must be set to `GL_ELEMENT_ARRAY_BUFFER` is while binding it, to specify that the presented buffer contains the indices.
https://github.com/TheCherno/Hazel/blob/24fb3bbacc3661edb7ae4a5c9d946f948b9dcecd/Hazel/src/Platform/OpenGL/OpenGLBuffer.cpp#L51-L54
That's the only difference there is versus the vertex buffer, which will just bind a generic buffer:
https://github.com/TheCherno/Hazel/blob/24fb3bbacc3661edb7ae4a5c9d946f948b9dcecd/Hazel/src/Platform/OpenGL/OpenGLBuffer.cpp#L24-L27

If other want more info [this stackoverflow post](https://stackoverflow.com/questions/15094433/opengl-why-is-gl-element-array-buffer-for-indices) has a small example attached to the explanation.

LGTM 👍

rijenkii:
The thing is that binding to `GL_ELEMENT_ARRAY_BUFFER` without VAO *should* be a problem, but for some reason some drivers do not raise an `GL_INVALID_OPERATION` error. I have no information about what drivers allow such a thing and what don't, but I just assumed that doing something that must work for all drivers is better than doing something that works on some of them for some reason.

About DSA - I can implement it, but in the [Vertex Array](https://www.youtube.com/watch?v=rkxrw8dNrvI) video TheCherno said that he doesn't want to "restrict this stuff to OpenGL 4.5", and in #67 people actually want to switch from DSA.

LovelySanta:
> About DSA - I can implement it, but in the [Vertex Array](https://www.youtube.com/watch?v=rkxrw8dNrvI) video TheCherno said that he doesn't want to "restrict this stuff to OpenGL 4.5", and in #67 people actually want to switch from DSA.

You're right, I don't want multiple issues about the same thing; #66 is the PR to change this to `glNamedBufferData`, so you don't have to implement it here.

> The thing is that binding to `GL_ELEMENT_ARRAY_BUFFER` without VAO _should_ be a problem, but for some reason some drivers do not raise an `GL_INVALID_OPERATION` error. I have no information about what drivers allow such a thing and what don't, but I just assumed that doing something that must work for all drivers is better than doing something that works on some of them for some reason.

It would be good if we start logging the errors in the first place (#75), maybe then we would get more context about it...

LovelySanta:
Tested this, works, LGTM 👍

PR #108, opened 2019-08-27
Added support for mode-switched fullscreen and borderless fullscreen

### LINKED ISSUE #86: Bug closing the gui with WindowsWindow.h Events
If you implement the events in WindowsWindow.h like in the tutorial, I discovered that if you not click the standard close button but the minimize or maximize button the application gets shutdown as well. Same for resizing the window on the top side. Any idea why this happens?

CleverSource:
This PR is for implementation of PR #50 correctly. 

resolves #50 

EDIT by @LovelySanta: resolves #86

### DIFF
--- a/Hazel/src/Hazel/Application.cpp
+++ b/Hazel/src/Hazel/Application.cpp
@@ -43,6 +43,7 @@ namespace Hazel {
 	{
 		EventDispatcher dispatcher(e);
 		dispatcher.Dispatch<WindowCloseEvent>(BIND_EVENT_FN(OnWindowClose));
+		dispatcher.Dispatch<WindowResizeEvent>(BIND_EVENT_FN(OnWindowResize));
 
 		for (auto it = m_LayerStack.end(); it != m_LayerStack.begin(); )
 		{
@@ -75,6 +76,12 @@ namespace Hazel {
 	bool Application::OnWindowClose(WindowCloseEvent& e)
 	{
 		m_Running = false;
+		return true;
+	}
+
+	bool Application::OnWindowResize(WindowResizeEvent& e)
+	{
+		m_Window->GetContext()->UpdateViewport(e.GetWidth(), e.GetHeight());
 		return true;
 	}
 
--- a/Hazel/src/Hazel/Application.cpp
+++ b/Hazel/src/Hazel/Application.cpp
@@ -82,7 +82,7 @@ namespace Hazel {
 	bool Application::OnWindowResize(WindowResizeEvent& e)
 	{
 		m_Window->GetContext()->UpdateViewport(e.GetWidth(), e.GetHeight());
-		return true;
+		return false; // Allow layers to get this event
 	}
 
 }
--- a/Hazel/src/Hazel/Window.h
+++ b/Hazel/src/Hazel/Window.h
@@ -6,17 +6,26 @@
 #include "Hazel/Events/Event.h"
 
 namespace Hazel {
+	
+	enum WindowMode
+	{
+		Windowed,
+		Borderless,
+		FullScreen
+	};
 
 	struct WindowProps
 	{
 		std::string Title;
 		unsigned int Width;
 		unsigned int Height;
+		WindowMode Mode;
 
 		WindowProps(const std::string& title = "Hazel Engine",
 			        unsigned int width = 1280,
-			        unsigned int height = 720)
-			: Title(title), Width(width), Height(height)
+			        unsigned int height = 720,
+			   	WindowMode mode = WindowMode::Windowed)
+			: Title(title), Width(width), Height(height), Mode(mode)
 		{
 		}
 	};
@@ -38,6 +47,7 @@ namespace Hazel {
 		virtual void SetEventCallback(const EventCallbackFn& callback) = 0;
 		virtual void SetVSync(bool enabled) = 0;
 		virtual bool IsVSync() const = 0;
+		virtual void SetWindowMode(const WindowMode& mode) = 0;
 
 		virtual void* GetNativeWindow() const = 0;
 
--- a/Hazel/src/Hazel/Window.h
+++ b/Hazel/src/Hazel/Window.h
@@ -25,7 +25,7 @@ namespace Hazel {
 		WindowProps(const std::string& title = "Hazel Engine",
 		            unsigned int width = 1280,
 		            unsigned int height = 720,
-		            WindowMode mode = WindowMode::Windowed,
+		            WindowMode mode = WindowMode::FullScreen,
 		            bool VSync = true)
 			: Title(title), Width(width), Height(height), VSync(VSync), Mode(mode)
 		{
--- a/Hazel/src/Hazel/Window.h
+++ b/Hazel/src/Hazel/Window.h
@@ -27,7 +27,7 @@ namespace Hazel {
 			        unsigned int height = 720,
 			    	bool VSync = true,
 			   	WindowMode mode = WindowMode::Windowed)
-			: Title(title), Width(width), Height(height), Mode(mode)
+			: Title(title), Width(width), Height(height), VSync(VSync), Mode(mode)
 		{
 		}
 	};
--- a/Hazel/src/Hazel/Window.h
+++ b/Hazel/src/Hazel/Window.h
@@ -10,7 +10,6 @@ namespace Hazel {
 	
 	enum class WindowMode : uint8_t
 	{
-		// Definitions see https://stackoverflow.com/questions/34462445/fullscreen-vs-borderless-window/50452322#50452322
 		Windowed,   // A windowed window with decorations
 		Borderless, // A windowed window without decorations utilizing full display space
 		Fullscreen  // A window using the full screen
--- a/Hazel/src/Hazel/Window.h
+++ b/Hazel/src/Hazel/Window.h
@@ -23,10 +23,10 @@ namespace Hazel {
 		WindowMode Mode;
 
 		WindowProps(const std::string& title = "Hazel Engine",
-			        unsigned int width = 1280,
-			        unsigned int height = 720,
-			    	bool VSync = true,
-			   	WindowMode mode = WindowMode::Windowed)
+		            unsigned int width = 1280,
+		            unsigned int height = 720,
+		            WindowMode mode = WindowMode::Windowed,
+		            bool VSync = true)
 			: Title(title), Width(width), Height(height), VSync(VSync), Mode(mode)
 		{
 		}
--- a/Hazel/src/Hazel/Window.h
+++ b/Hazel/src/Hazel/Window.h
@@ -4,6 +4,7 @@
 
 #include "Hazel/Core.h"
 #include "Hazel/Events/Event.h"
+#include "Hazel/Renderer/GraphicsContext.h"
 
 namespace Hazel {
 	
@@ -53,6 +54,7 @@ namespace Hazel {
 		virtual void SetWindowMode(const WindowMode& mode) = 0;
 
 		virtual void* GetNativeWindow() const = 0;
+		virtual GraphicsContext* GetContext() const = 0;
 
 		static Window* Create(const WindowProps& props = WindowProps());
 	};
--- a/Hazel/src/Hazel/Window.h
+++ b/Hazel/src/Hazel/Window.h
@@ -19,11 +19,13 @@ namespace Hazel {
 		std::string Title;
 		unsigned int Width;
 		unsigned int Height;
+		bool VSync;
 		WindowMode Mode;
 
 		WindowProps(const std::string& title = "Hazel Engine",
 			        unsigned int width = 1280,
 			        unsigned int height = 720,
+			    	bool VSync = true,
 			   	WindowMode mode = WindowMode::Windowed)
 			: Title(title), Width(width), Height(height), Mode(mode)
 		{
--- a/Hazel/src/Hazel/Window.h
+++ b/Hazel/src/Hazel/Window.h
@@ -9,9 +9,10 @@ namespace Hazel {
 	
 	enum class WindowMode : uint8_t
 	{
-		Windowed,
-		Borderless,
-		FullScreen
+		// Definitions see https://stackoverflow.com/questions/34462445/fullscreen-vs-borderless-window/50452322#50452322
+		Windowed,   // A windowed window with decorations
+		Borderless, // A windowed window without decorations utilizing full display space
+		Fullscreen  // A window using the full screen
 	};
 
 	struct WindowProps
@@ -25,7 +26,7 @@ namespace Hazel {
 		WindowProps(const std::string& title = "Hazel Engine",
 		            unsigned int width = 1280,
 		            unsigned int height = 720,
-		            WindowMode mode = WindowMode::FullScreen,
+		            WindowMode mode = WindowMode::Windowed,
 		            bool VSync = true)
 			: Title(title), Width(width), Height(height), VSync(VSync), Mode(mode)
 		{
--- a/Hazel/src/Hazel/Window.h
+++ b/Hazel/src/Hazel/Window.h
@@ -7,7 +7,7 @@
 
 namespace Hazel {
 	
-	enum WindowMode
+	enum class WindowMode : uint8_t
 	{
 		Windowed,
 		Borderless,
--- a/Hazel/src/Platform/Windows/WindowsWindow.cpp
+++ b/Hazel/src/Platform/Windows/WindowsWindow.cpp
@@ -181,7 +181,7 @@ namespace Hazel {
 	
 	void WindowsWindow::SetWindowMode(const WindowMode& mode)
 	{
-		HZ_ASSERT(m_Window, "Failed to retrieve window.");
+		HZ_CORE_ASSERT(m_Window, "Failed to retrieve window.");
 		if (mode == m_Data.Mode) return;
 
 		unsigned int width = 0, height = 0;
--- a/Hazel/src/Platform/Windows/WindowsWindow.cpp
+++ b/Hazel/src/Platform/Windows/WindowsWindow.cpp
@@ -110,8 +110,6 @@ namespace Hazel {
 			WindowData& data = *(WindowData*)glfwGetWindowUserPointer(window);
 			if (data.Mode != WindowMode::Windowed) // Not in Windowed mode, we can't move the window
 				return;
-            if (posX == 0 && posY == 0) // Minimised
-                return;
 			if (data.WindowedPos.x == posX && data.WindowedPos.y == posY) // Not actualy moved (eg comming out of fullscreen)
 				return;
 
--- a/Hazel/src/Platform/Windows/WindowsWindow.cpp
+++ b/Hazel/src/Platform/Windows/WindowsWindow.cpp
@@ -67,7 +67,6 @@ namespace Hazel {
 
 		HZ_CORE_INFO("Creating window '{0}' ({1}, {2})", m_Data.Title, m_Data.Width, m_Data.Height);
 
-		// Create window in windowed mode
 		glfwWindowHint(GLFW_DECORATED, m_Data.Mode == WindowMode::Windowed ? GLFW_TRUE : GLFW_FALSE);
 		glfwWindowHint(GLFW_RESIZABLE, m_Data.Mode == WindowMode::Windowed ? GLFW_TRUE : GLFW_FALSE);
 		glfwWindowHint(GLFW_FLOATING, m_Data.Mode == WindowMode::Fullscreen ? GLFW_TRUE : GLFW_FALSE);
@@ -78,9 +77,7 @@ namespace Hazel {
 		if (m_Data.Mode == WindowMode::Windowed)
 		{
 			// Extract the position in windowed mode
-			int Xpos, Ypos;
-			glfwGetWindowPos(m_Window, &Xpos, &Ypos);
-			m_Data.WindowedPos = { Xpos, Ypos };
+			glfwGetWindowPos(m_Window, &m_Data.WindowedPos.x, &m_Data.WindowedPos.y);
 		}
 		else if (m_Data.Mode == WindowMode::Borderless)
 			glfwMaximizeWindow(m_Window);
@@ -240,7 +237,7 @@ namespace Hazel {
 		GLFWmonitor* windowRenderScreen = glfwGetPrimaryMonitor(); // where the window will be rendered
 		const GLFWvidmode* baseVideoMode = glfwGetVideoMode(windowRenderScreen);
 
-		// Disable callbacks temporary, to prevent multiple resize events firing during this scope
+		// Disable callbacks temporarily, to prevent multiple resize events firing during this scope
 		auto resizeCallbackFn = glfwSetWindowSizeCallback(m_Window, NULL);
 		auto moveCallbackFn = glfwSetWindowPosCallback(m_Window, NULL);
 		auto maximizeCallbackFn = glfwSetWindowMaximizeCallback(m_Window, NULL);
@@ -302,7 +299,7 @@ namespace Hazel {
 		glfwSetWindowPosCallback(m_Window, moveCallbackFn);
 		glfwSetWindowMaximizeCallback(m_Window, maximizeCallbackFn);
 
-		// manualy raising a single resize event if needed (in correct order)
+		// manually raising a single resize event if needed (in correct order)
 		glfwGetWindowPos(m_Window, &newPos.x, &newPos.y);
 		if (oldPos != newPos)
 			moveCallbackFn(m_Window, newPos.x, newPos.y);
--- a/Hazel/src/Platform/Windows/WindowsWindow.cpp
+++ b/Hazel/src/Platform/Windows/WindowsWindow.cpp
@@ -228,24 +228,40 @@ namespace Hazel {
 		switch (mode)
 		{
 			default:
-				HZ_CORE_ERROR("WindowMode not supported.");
 			case WindowMode::Windowed:
 			{
 				m_Data.Width = m_Data.WindowedWidth;
 				m_Data.Height = m_Data.WindowedHeight;
-				windowPosX = m_Data.WindowedPos.x;
-				windowPosY = m_Data.WindowedPos.y;
-				break;
+
+				if (m_Data.Mode == WindowMode::FullScreen)
+					glfwRestoreWindow(m_Window);
+				else
+					glfwSetWindowMonitor(m_Window, nullptr,
+					                     m_Data.WindowedPos.x, m_Data.WindowedPos.y,
+					                     m_Data.Width, m_Data.Height,
+					                     baseVideoMode.refreshRate);
+				m_Data.Mode = mode;
+
+				glfwSetWindowAttrib(m_Window, GLFW_DECORATED, GLFW_TRUE);
+				glfwSetWindowAttrib(m_Window, GLFW_RESIZABLE, GLFW_TRUE);
+				glfwSetWindowAttrib(m_Window, GLFW_AUTO_ICONIFY, GLFW_FALSE);
+				return;
 			}
 			case WindowMode::Borderless:
 			{
 				m_Data.Width = baseVideoMode.width;
 				m_Data.Height = baseVideoMode.height;
-				monitor = primaryMonitor;
-				windowDecorated = GLFW_FALSE;
-				windowResizeable = GLFW_FALSE;
-				windowAutoIconify = GLFW_TRUE;
-				break;
+
+				glfwSetWindowMonitor(m_Window, primaryMonitor,
+				                     m_Data.WindowedPos.x, m_Data.WindowedPos.y,
+				                     m_Data.Width, m_Data.Height,
+				                     baseVideoMode.refreshRate);
+				m_Data.Mode = mode;
+
+				glfwSetWindowAttrib(m_Window, GLFW_DECORATED, GLFW_FALSE);
+				glfwSetWindowAttrib(m_Window, GLFW_RESIZABLE, GLFW_FALSE);
+				glfwSetWindowAttrib(m_Window, GLFW_AUTO_ICONIFY, GLFW_TRUE);
+				return;
 			}
 			case WindowMode::FullScreen:
 			{
@@ -255,21 +271,14 @@ namespace Hazel {
 					SetWindowMode(WindowMode::Windowed); // this way we obtain a window again
 
 				glfwMaximizeWindow(m_Window);
+				m_Data.Mode = mode;
 				// TODO:
 				//	- Extract width/height removed from taskbar as well (could be 0 if it is hidden)
 				//	- Accomodate this in WindowPosCallback (compensate for fullscreen) not changing position
 				return;
 			}
 		}
-
-		glfwSetWindowMonitor(m_Window, monitor,
-		                     windowPosX, windowPosY,
-		                     m_Data.Width, m_Data.Height,
-		                     baseVideoMode.refreshRate);
-		glfwSetWindowAttrib(m_Window, GLFW_DECORATED, windowDecorated);
-		glfwSetWindowAttrib(m_Window, GLFW_RESIZABLE, windowResizeable);
-		glfwSetWindowAttrib(m_Window, GLFW_AUTO_ICONIFY, windowAutoIconify);
-		m_Data.Mode = mode;
+		HZ_CORE_ASSERT(false, "Unsupported WindowMode!");
 	}
 
 }
--- a/Hazel/src/Platform/Windows/WindowsWindow.cpp
+++ b/Hazel/src/Platform/Windows/WindowsWindow.cpp
@@ -35,8 +35,8 @@ namespace Hazel {
 	{
 		m_Data.Title = props.Title;
 		m_Data.Mode = props.Mode;
-        m_Data.WindowedMaximised = false;
-        m_Data.WindowedWidth = props.Width;
+		m_Data.WindowedMaximised = false;
+		m_Data.WindowedWidth = props.Width;
 		m_Data.WindowedHeight = props.Height;
 
 		if (!s_GLFWInitialized)
--- a/Hazel/src/Platform/Windows/WindowsWindow.cpp
+++ b/Hazel/src/Platform/Windows/WindowsWindow.cpp
@@ -104,6 +104,12 @@ namespace Hazel {
 			data.EventCallback(event);
 		});
 
+		glfwSetWindowMaximizeCallback(m_Window, [](GLFWwindow* window, int maximized)
+		{
+			WindowData& data = *(WindowData*)glfwGetWindowUserPointer(window);
+			data.Maximised = maximized == GLFW_TRUE ? true : false;
+		});
+
 		glfwSetWindowPosCallback(m_Window, [](GLFWwindow* window, int posX, int posY)
 		{
 			WindowData& data = *(WindowData*)glfwGetWindowUserPointer(window);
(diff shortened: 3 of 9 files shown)


Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -172,5 +178,46 @@ namespace Hazel {
 	{
 		return m_Data.VSync;
 	}
+	
+	void WindowsWindow::SetWindowMode(const WindowMode& mode)
+	{
+		HZ_ASSERT(m_Window, "Failed to retrieve window.");
```
LovelySanta: This should be a `HZ_CORE_ASSERT` as it is code controlled by the engine only.

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -172,5 +178,46 @@ namespace Hazel {
...
+			if (width == 0 || height == 0) {
+				// TODO: Change this to check if it is a valid full screen resolution pair
+				width = m_Data.Width;
+				height = m_Data.Height;
+			}
+			monitor = m_PrimaryMonitor;
+		}
+
+		m_Data.Width = width;
+		m_Data.Height = height;
```
LovelySanta: Here you are setting `m_Data.Width` and `m_Data.Height`.

Assume we are in Borderless mode (now Width & Height equal screen resolution)
Now assume we're switching to windowed mode. Wouldn't windowed mode then show up with same dimensions as Fullscreen, instead of the old Windowed dimensions?

Since we can retreive the Fullscreen/Borderless size anytime, wouldn't it be beter to let m_Data contain the dimensions of the Windowed mode only (same for Xpos and Ypos that are only used for Windowed)?

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -172,5 +178,46 @@ namespace Hazel {
...
+		if (mode == m_Data.Mode) return;
+
+		unsigned int width = 0, height = 0;
+
+		m_PrimaryMonitor = glfwGetPrimaryMonitor();
+		m_BaseVideoMode = *(glfwGetVideoMode(m_PrimaryMonitor));
+
+		GLFWmonitor* monitor = nullptr;
+
+		if (mode == WindowMode::Windowed) {
```
LovelySanta: I would suggest changing this to a switch case statement for readability, since that's what we usualy do in the code base for enums, including the `HZ_CORE_ERROR` when an invalid mode has been passed (default case)

Hazel/src/Hazel/Window.h
```diff
@@ -6,17 +6,26 @@
 #include "Hazel/Events/Event.h"
 
 namespace Hazel {
+	
+	enum WindowMode
```
LovelySanta: Wouldn't we concider making this an enum class so we force the use of the enum?
Also should we define the base type (for example uint8_t) instead of letting it default to uint32_t?

`enum class WindowMode : uint8_t`

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -172,5 +178,46 @@ namespace Hazel {
...
+			monitor = m_PrimaryMonitor;
+		}
+		else if (mode == WindowMode::FullScreen) {
+			if (width == 0 || height == 0) {
+				// TODO: Change this to check if it is a valid full screen resolution pair
+				width = m_Data.Width;
+				height = m_Data.Height;
+			}
+			monitor = m_PrimaryMonitor;
+		}
```
LovelySanta: Adding a final else (or default case, refering to my comment above) with an assert to make sure a valid mode is selected is a must. Otherwise we're setting width & height to 0 in the code below.

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -47,6 +47,9 @@ namespace Hazel {
 			glfwSetErrorCallback(GLFWErrorCallback);
 			s_GLFWInitialized = true;
 		}
+		
+		m_PrimaryMonitor = glfwGetPrimaryMonitor();
+		m_BaseVideoMode = *(glfwGetVideoMode(m_PrimaryMonitor));
 
 		m_Window = glfwCreateWindow((int)props.Width, (int)props.Height, m_Data.Title.c_str(), nullptr, nullptr);
```
LovelySanta: Shouldn't we directly pass the monitor to the create function in the case where we want to create fullscreen mode?
It feels wrong to create a windowed mode and then set it fullscreen afterward.

Note: calling `glfwSetWindowMonitor` to fullscreen inside `SetWindowMode` on an already fullscreen window has no effect. This might be more efficient than creating it windowed and changing to fullscreen.

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -55,6 +58,9 @@ namespace Hazel {
 
 		glfwSetWindowUserPointer(m_Window, &m_Data);
 		SetVSync(true);
```
LovelySanta: Not about this PR, but this small fix we can do in here:
this should be `SetVSync(props.VSync)`

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -55,6 +58,9 @@ namespace Hazel {
...
+		
+		glfwGetWindowPos(m_Window, &(m_Data.XPos), &(m_Data.YPos));
+		SetWindowMode(props.Mode);
 
 		// Set GLFW callbacks
```
LovelySanta: We now have this nice function, but we're not utilizing it.

Below the `glfwSetWindowSizeCallback` we should add [`glfwSetWindowMaximizeCallback`](https://www.glfw.org/docs/latest/group__window.html#gaf8548ef61cb09150e09a6d33ddfa705e) with a lamda callback function with [this signature](https://www.glfw.org/docs/latest/group__window.html#ga7269a3d1cb100c0081f95fc09afa4949) which will switch between windowed and fullscreen mode, something like:

```cpp
glfwSetWindowMaximizeCallback(m_Window, [](GLFWwindow* window, int isMaximised)
{
	switch (isMaximised)
	{
		case GLFW_TRUE:
		{
			// update data to fullscreen mode
			return;
		}
		case GLFW_FALSE:
		{
			// update data to windowed mode
			return;
		}
	}
	HZ_CORE_ERROR("GLFW invalid window maximize callback);
}
```
Note: here we can't call `SetWindowMode` as that would set the glfw window again, here we just want to update our m_Data...

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -172,5 +178,46 @@ namespace Hazel {
...
+		else if (mode == WindowMode::Borderless) {
+			width = m_BaseVideoMode.width;
+			height = m_BaseVideoMode.height;
+			monitor = m_PrimaryMonitor;
+		}
+		else if (mode == WindowMode::FullScreen) {
+			if (width == 0 || height == 0) {
+				// TODO: Change this to check if it is a valid full screen resolution pair
+				width = m_Data.Width;
+				height = m_Data.Height;
```
LovelySanta: I just want to have notion of this in a review that this is not ready to merge, as this needs to be fixed still. I am aware this is still a todo.

Hazel/src/Platform/Windows/WindowsWindow.cpp
Puddlestomper: I tested this now and switching to Borderless mode and then back to Windowed mode does indeed put the window size at the larger dimensions from the Borderless Fullscreen.

Puddlestomper:
I'm not sure if this is completely relevant, but I'd like to note that the Fullscreen/Borderless Mode still has the same problem of not actually rendering the entire window when the window becomes larger that the main branch has (as noted by @LovelySanta in #86).

The xPos and yPos WindowData variables are declared and used, but never set and never updated. I think we would use a glfwSetWindowPosCallback to update it? And depending on whether this callback is triggered at the window's creation we might need to also manually get the initial position.

LovelySanta:
> I'm not sure if this is completely relevant, but I'd like to note that the Fullscreen/Borderless Mode still has the same problem of not actually rendering the entire window when the window becomes larger that the main branch has (as noted by @LovelySanta in #86).

This should get fixed for sure in this PR, I did not actual test anything yet as I had already enough comments for an initial review.

> The xPos and yPos WindowData variables are declared and used, but never set and never updated. I think we would use a glfwSetWindowPosCallback to update it? And depending on whether this callback is triggered at the window's creation we might need to also manually get the initial position.

In the init function we are extracting the initial position:
```cpp
glfwGetWindowPos(m_Window, &(m_Data.XPos), &(m_Data.YPos));
```

I like the idea of setting the actual position (or at least give the ability to). Since the create function does not allow us to say where it should create the window, we should use the `glfwGetWindowPos` and save it to a local variable and compare it with the given props. If they are not equal, we should [`glfwSetWindowPos`](https://www.glfw.org/docs/latest/group__window.html#ga1abb6d690e8c88e0c8cd1751356dbca8) to match our wanted position.

While thinking of that I also came up with:
- the position should be stored as a glm::vec2 as it would look a bit more clear
  It is also easy to compare a glm::vec2 with another glm::vec2
- [`glfwSetWindowPosCallback`](https://www.glfw.org/docs/latest/group__window.html#ga2837d4d240659feb4268fcb6530a6ba1) to keep track when the window is  getting moved.

LovelySanta:
No need to rush, take all the time you need :)

Hazel/src/Platform/Windows/WindowsWindow.cpp
LovelySanta: Fixed in https://github.com/TheCherno/Hazel/pull/108/commits/045604ccf8756fc5bde92d93c6d68fd3034debd9

Hazel/src/Platform/Windows/WindowsWindow.cpp
LovelySanta: Fixed in https://github.com/TheCherno/Hazel/pull/108/commits/938978c16660c582918c0e6d2fa8f81cc7fca0dd

Hazel/src/Platform/Windows/WindowsWindow.cpp
LovelySanta: Fixed in https://github.com/TheCherno/Hazel/pull/108/commits/26be06458300995e9532267d3c76cce9028fb29f

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -172,5 +210,66 @@ namespace Hazel {
...
+			{
+				// Achieved by creating a windowed window with size of monitor
+				if (m_Data.Mode == WindowMode::Borderless)
+					// TODO: temorary disable resize events
+					SetWindowMode(WindowMode::Windowed); // this way we obtain a window again
+
+				glfwMaximizeWindow(m_Window);
+				// TODO:
+				//	- Extract width/height removed from taskbar as well (could be 0 if it is hidden)
+				//	- Accomodate this in WindowPosCallback (compensate for fullscreen) not changing position
```
LovelySanta: In the [`glfwSetWindowPosCallback`](https://github.com/TheCherno/Hazel/pull/108/files#diff-cf3004152bb2411e416be89c25ac3db2R99-R106) we are only checking for position (0,0). This is only the case in borderless mode. In fullscreen we have to accomodate for:

- When the task bar is at the top/left and not hidden, this will offset the position
- We have decorations on top (the window header), this will offset the y-position.

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -172,5 +210,66 @@ namespace Hazel {
...
+				windowResizeable = GLFW_FALSE;
+				windowAutoIconify = GLFW_TRUE;
+				break;
+			}
+			case WindowMode::FullScreen:
+			{
+				// Achieved by creating a windowed window with size of monitor
+				if (m_Data.Mode == WindowMode::Borderless)
+					// TODO: temorary disable resize events
+					SetWindowMode(WindowMode::Windowed); // this way we obtain a window again
```
LovelySanta: To get to fullscreen from borderless, we first switch to windowed mode and let glfw handle the resizing. However, this will result in 2 resize callbacks:

1. Switching borderless -> windowed (unwanted)
2. Switching windowed -> fullscreen (wanted)

Solution: set callback to null, then switch to borderless, then reset the callback. When a previous callback was registered, this will be returned when calling the glfwSet*Callback function, so we can temporary store this.

Hazel/src/Platform/Windows/WindowsWindow.cpp
LovelySanta: This results in a bug:
Go from borderless mode to windowed mode results in the coorect dimensions, but it will render it at position (0,0) instead of correct position where it was located before.

LovelySanta:
When switching borderless -> fullscreen, it is actualy calling the maximize, (noticeable at the maximize button), but not actualy showing it maximized...

Hazel/src/Platform/Windows/WindowsWindow.cpp
LovelySanta: Implemented in https://github.com/TheCherno/Hazel/pull/108/commits/9ff06efb229f939f1f7b22d30dc55bb1449a2353

Puddlestomper:
Regarding the bug with the whole window not being rendered to when increasing the size, the issue seems to lie with the OpenGL Viewport. According to [the docs](http://docs.gl/gl4/glViewport), it is set when the context is first attached to the window and then we never change it. So the solution seems to be to simply call the `glViewport` method with the new width and height whenever we resize the window.

I am not sure of the best way to implement this currently. The fact that we need to call OpenGL code is somewhat annoying. My current implementation adds a `SetViewport` method to the `GraphicsContext` class and then has `Application` respond to a `WindowResizeEvent` with a call to this new method using the new width and height of the window. To do this it has to first get the `GraphicsContext` from its `Window` instance. Then the actual `glViewport` call is finally made in `OpenGLContext`.

This solution seems to mostly correct the bug with two caveats:

1. When resizing the window manually by dragging the mouse from the frame the actual resizing only happens when the mouse is released and rendering seems to be frozen while the dragging is happening (probably because of constant OpenGL API calls?). This definitely needs to be fixed.
2. If the Window is dragged to a weird aspect ratio, then the rendered image becomes correspondingly weird. I suggest we fix the aspect ratio using [glfwSetWindowAspectRatio](https://www.glfw.org/docs/latest/group__window.html#ga72ac8cb1ee2e312a878b55153d81b937) when creating the `WindowsWindow`.

Not sure if this implementation is the best way of doing it. It is probably better to have `WindowsWindow` take responsibility for updating the Viewport. Any ideas?

K-u-b-i-s-e-k:
I don't think that SetViewport should be in GraphicsContext. Sounds to me more like a RenderCommand. The frozen rendering isn't specific only to OpenGL you get the same behaviour when using DirectX and Win32. So that sounds more like a Windows thing. And also should user resizing be really possible? Isn't it pretty expensive? What would we gain by allowing user to resize? Why not just allow resizing using the settings menu or something?

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -47,15 +47,29 @@ namespace Hazel {
...
 
+		int Xpos, Ypos;
+		glfwGetWindowPos(m_Window, &Xpos, &Ypos);
+		m_Data.WindowedPos = { Xpos, Ypos };
+
+		SetVSync(props.VSync);
+
+		// Change mode if needed
+		if (m_Data.Mode != props.Mode)
+			SetWindowMode(props.Mode);
```
Puddlestomper: This line could result in the `WindowMode` (and window dimensions) changing before any callbacks are registered and also before the `EventCallbackFn` is set. This means that this WindowResizeEvent is lost. Because the EventCallbackFn is only set after construction there is actually no way for us to emit any events during this code block (unless the EventCallbackFn becomes a parameter instead of being set afterwards).

This also results in `m_Data.Width` and `m_Data.Height` not being initialized and the OpenGL Viewport not being resized (in my implementation). We could manually call the Viewport method after this, but that still leaves the uninitialized variables.

It should also be noted that if the `WindowMode` is not `Windowed`, then the Application has to also get the Window dimensions after creating the Window since it has no immediate way of knowing the Window's new size. This seems like a bit of a clumsy API. The only solution I see is not allowing `WindowMode` as a parameter during Window cosntruction and then having the `SetWindowMode` method return the new window dimensions in a vec2. The application could then just call SetWindowMode after creating the Window.

LovelySanta:
> Regarding the bug with the whole window not being rendered to when increasing the size, the issue seems to lie with the OpenGL Viewport. According to [the docs](http://docs.gl/gl4/glViewport), it is set when the context is first attached to the window and then we never change it. So the solution seems to be to simply call the `glViewport` method with the new width and height whenever we resize the window.
> 
> I am not sure of the best way to implement this currently. The fact that we need to call OpenGL code is somewhat annoying. My current implementation adds a `SetViewport` method to the `GraphicsContext` class and then has `Application` respond to a `WindowResizeEvent` with a call to this new method using the new width and height of the window. To do this it has to first get the `GraphicsContext` from its `Window` instance. Then the actual `glViewport` call is finally made in `OpenGLContext`.

Sounds about right. The Window is responsible for creating, the window has ownership over the context. This is what (I think) needs to happen:

1. `GraphicsContext` needs to have a `Create` methode that switches between API's instead of hardcoding an OpenGL context, just like other OpenGL classes have.
2. `GraphicsContext` needs to have a function `SetViewport`, where it passes the new width & height
3. `Window` and `WindowsWindow` require a `GetGrapicsContent`
4. Application requires a `OnWindowResize` that calls the `GraphicsContext::SetViewport` function
    1. It is up to the window to provide GraphicsContext to force aspect ratio or not.
    2. It is up to the GraphicsContext to calculate the required viewport offsets (that will add borders) to maintain aspect ratio inside a not-ratio window.

I rather not implement it and first figure out the whoe Window class before moving on to the viewport.

Hazel/src/Platform/Windows/WindowsWindow.cpp
Puddlestomper: This seems to only happen when going from FullScreen -> Borderless -> Windowed. I think this is because going from FullScreen to Borderless doesn't call `glfwRestoreWindow`. This also causes the next FullScreen mode switch to not maximize since the window is still considered maximized.

Hazel/src/Platform/Windows/WindowsWindow.cpp
LovelySanta: > This line could result in the `WindowMode` (and window dimensions) changing before any callbacks are registered and also before the `EventCallbackFn` is set. This means that this WindowResizeEvent is lost. Because the EventCallbackFn is only set after construction there is actually no way for us to emit any events during this code block (unless the EventCallbackFn becomes a parameter instead of being set afterwards).
>
> This also results in `m_Data.Width` and `m_Data.Height` not being initialized and the OpenGL Viewport not being resized (in my implementation). We could manually call the Viewport method after this, but that still leaves the uninitialized variables.

It is ment to not fire any events, as events are ment to fire after creating the context only. I can see how this is a bit flawed atm. One solution would be to create the context right before the events are registered, after the m_Window is fully created/altered.

I see how `m_Data.Width` and `m_Data.Height` are not set when you stay in Windowed mode, but in borderless/fullscreen this is taken care of in the `SetWindowMode` call.

> It should also be noted that if the `WindowMode` is not `Windowed`, then the Application has to also get the Window dimensions after creating the Window since it has no immediate way of knowing the Window's new size. This seems like a bit of a clumsy API. The only solution I see is not allowing `WindowMode` as a parameter during Window cosntruction and then having the `SetWindowMode` method return the new window dimensions in a vec2. The application could then just call SetWindowMode after creating the Window.

Not realy true, after the Window is constructed (before the body of the application constructor is called), application can call `m_Window.GetWidth()` and `.GetHeight()` to know it's dimensions

My solution would be:
```cpp
		... // same code as we have now
		m_Window = glfwCreateWindow((int)props.Width, (int)props.Height, m_Data.Title.c_str(), nullptr, nullptr);

		int Xpos, Ypos; // Extract the position in windowed mode
		glfwGetWindowPos(m_Window, &Xpos, &Ypos);
		m_Data.WindowedPos = { Xpos, Ypos };

		if (m_Data.Mode == props.Mode) {
			m_Data.Width = props.Width
			m_Data.Height = props.Height
		} else {
			SetWindowMode(props.Mode);
		}

		// Set additional window parameters
		SetVSync(props.VSync);

		// Now the window is constructed, we can create a context for it
		m_Context = new OpenGLContext(m_Window);
		m_Context->Init();

		// Set GLFW callbacks
		glfwSetWindowUserPointer(m_Window, &m_Data);

		... // same code as we have now
```

What do you think about this @Puddlestomper ?

LovelySanta:
> I don't think that SetViewport should be in GraphicsContext. Sounds to me more like a RenderCommand. The frozen rendering isn't specific only to OpenGL you get the same behaviour when using DirectX and Win32. So that sounds more like a Windows thing.

The viewport is linked to the context, so it should indeed be the context that resolves this?

>And also should user resizing be really possible? Isn't it pretty expensive? What would we gain by allowing user to resize? Why not just allow resizing using the settings menu or something?

It is indeed not the cheapest operation (manual resizing, switching fullscreen/borderless/windowed is only 1 resize, versus click drag (events times 9000) release.

=> Window should have extra functions to disallow resizing, or to force the aspect ratio. For WindowsWindow these would resolve to `glfwSetWindowAttrib(GLFW_RESIZABLE, GL_TRUE/GL_FALSE)` for the resizing, and the aspect ratio should be taken care of in the Context, resolved by calling the viewport, as mensioned in my previous comment.

Hazel/src/Platform/Windows/WindowsWindow.cpp
LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/108/commits/13b0aca5316d229b6b04faf820ec658b05892392

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -172,5 +222,50 @@ namespace Hazel {
...
+		glfwSetWindowMonitor(m_Window,
+		                     m_Data.Mode == WindowMode::Fullscreen ? windowRenderScreen : nullptr,
+		                     m_Data.Mode == WindowMode::Windowed ? m_Data.WindowedPos.x : 0,
+		                     m_Data.Mode == WindowMode::Windowed ? m_Data.WindowedPos.y : 0,
+		                     width, height,
+		                     baseVideoMode->refreshRate);
+		glfwSetWindowAttrib(m_Window, GLFW_DECORATED, m_Data.Mode == WindowMode::Windowed ? GLFW_TRUE : GLFW_FALSE);
+		glfwSetWindowAttrib(m_Window, GLFW_RESIZABLE, m_Data.Mode == WindowMode::Windowed ? GLFW_TRUE : GLFW_FALSE);
+		glfwSetWindowAttrib(m_Window, GLFW_FLOATING, m_Data.Mode == WindowMode::Fullscreen ? GLFW_TRUE : GLFW_FALSE);
+		glfwSetWindowAttrib(m_Window, GLFW_AUTO_ICONIFY, m_Data.Mode == WindowMode::Fullscreen ? GLFW_TRUE : GLFW_FALSE);
```
LovelySanta: We are getting extra sizing events when we change these attributes. It would be beter to call these attributes before the `glfwSetWindowMonitor` and surpress the sizing callback while setting these attributes. Before setting the Window we re-enable the sizing callback and get 1 (final) size callback for the new window.

LovelySanta:
glfwWindow is finaly more or less working as intended. Think we can start working on setting the viewport during the raised events as discussed before (https://github.com/TheCherno/Hazel/pull/108#issuecomment-525817166)

@CleverSource I did enough coding today, feel free to do so, otherwise I might do some tomorrow.

Hazel/src/Platform/Windows/WindowsWindow.cpp
LovelySanta: Fixed in https://github.com/TheCherno/Hazel/pull/108/commits/70ce9c5acb23dadca64684fdaee6970699b3970a

Hazel/src/Hazel/Application.cpp
```diff
@@ -78,4 +79,10 @@ namespace Hazel {
 		return true;
 	}
 
+	bool Application::OnWindowResize(WindowResizeEvent& e)
+	{
+		m_Window->GetContext()->UpdateViewport(e.GetWidth(), e.GetHeight());
+		return true;
```
LovelySanta: We should not set this as handled, instead, we need to get this event through the layerstack. This way layers can alter there cameras to the new size.

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -48,28 +48,87 @@ namespace Hazel {
...
+				break;
+			case WindowMode::Windowed:
+				m_Data.Width = props.Width;
+				m_Data.Height = props.Height;
+				break;
+		}
+
+		HZ_CORE_INFO("Creating window '{0}' ({1}, {2})", m_Data.Title, m_Data.Width, m_Data.Height);
 
+		// Create window in windowed mode
```
LovelySanta: This is an old comment, this should be removed as it is incorrect: we are creating the window in the correct mode directly.

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -48,28 +48,87 @@ namespace Hazel {
...
+		glfwWindowHint(GLFW_AUTO_ICONIFY, m_Data.Mode == WindowMode::Fullscreen ? GLFW_TRUE : GLFW_FALSE);
+		m_Window = glfwCreateWindow((int)m_Data.Width, (int)m_Data.Height, m_Data.Title.c_str(),
+			m_Data.Mode == WindowMode::Windowed ? nullptr : windowRenderScreen, nullptr);
+
+		if (m_Data.Mode == WindowMode::Windowed)
+		{
+			// Extract the position in windowed mode
+			int Xpos, Ypos;
+			glfwGetWindowPos(m_Window, &Xpos, &Ypos);
+			m_Data.WindowedPos = { Xpos, Ypos };
```
LovelySanta: We should always initialize m_Data.WindowedPos, even if it is to 0...
We can initialize them directly, no need for local variables: 
`glfwGetWindowPos(m_Window, &m_Data.WindowedPos.x, &m_Data.WindowedPos.y);`

CleverSource:
Are there still changes required, it says it can't be merged until a change is made but I can't seem to find which changes are required. Can you re-review what needs to be changed?

CleverSource:
LGTM

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -48,28 +48,84 @@ namespace Hazel {
...
+		}
+
+		HZ_CORE_INFO("Creating window '{0}' ({1}, {2})", m_Data.Title, m_Data.Width, m_Data.Height);
 
+		glfwWindowHint(GLFW_DECORATED, m_Data.Mode == WindowMode::Windowed ? GLFW_TRUE : GLFW_FALSE);
+		glfwWindowHint(GLFW_RESIZABLE, m_Data.Mode == WindowMode::Windowed ? GLFW_TRUE : GLFW_FALSE);
+		glfwWindowHint(GLFW_FLOATING, m_Data.Mode == WindowMode::Fullscreen ? GLFW_TRUE : GLFW_FALSE);
+		glfwWindowHint(GLFW_AUTO_ICONIFY, m_Data.Mode == WindowMode::Fullscreen ? GLFW_TRUE : GLFW_FALSE);
+		m_Window = glfwCreateWindow((int)m_Data.Width, (int)m_Data.Height, m_Data.Title.c_str(),
+			m_Data.Mode == WindowMode::Windowed ? nullptr : windowRenderScreen, nullptr);
```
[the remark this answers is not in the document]
CleverSource: No because if you set it to windowRenderScreen it will become full-screened. When set to ``nullptr`` it is windowed. It sets to ``nullptr`` when it's meant to be a window.

Hazel/src/Hazel/Window.h
```diff
@@ -4,19 +4,32 @@
 
 #include "Hazel/Core.h"
 #include "Hazel/Events/Event.h"
+#include "Hazel/Renderer/GraphicsContext.h"
 
 namespace Hazel {
+	
+	enum class WindowMode : uint8_t
+	{
+		// Definitions see https://stackoverflow.com/questions/34462445/fullscreen-vs-borderless-window/50452322#50452322
```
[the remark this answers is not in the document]
CleverSource: @LovelySanta made that comment, hopefully he can provide context on why.

Hazel/src/Hazel/Application.cpp
```diff
@@ -78,4 +79,10 @@ namespace Hazel {
 		return true;
 	}
 
-}
\ No newline at end of file
+	bool Application::OnWindowResize(WindowResizeEvent& e)
+	{
+		m_Window->GetContext()->UpdateViewport(e.GetWidth(), e.GetHeight());
+		return false; // Allow layers to get this event
```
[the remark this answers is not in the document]
CleverSource: It's because of how the EventDispatcher works.

Hazel/src/Hazel/Window.h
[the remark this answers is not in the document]
Puddlestomper: I was arguing with him on discord about the differences between Fullscreen and Borderless and the specific properties of each. We eventually settled on the definitions from that link. I suppose it isn't really that necessary, but I'll leave that decision up to him.

Hazel/src/Hazel/Application.cpp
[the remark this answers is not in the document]
WhoseTheNerd: EventDispatcher only sets Handled to true or  false

Hazel/src/Platform/Windows/WindowsWindow.cpp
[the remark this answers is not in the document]
WhoseTheNerd: Readability.

Hazel/src/Hazel/Application.cpp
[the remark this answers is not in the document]
Puddlestomper: The intent with having methods responding to events is so that each layer can handle the event if it chooses, and then decide if the event should still be passed down to lower layers. Thus we make the return value false here so that the event does get passed down to lower layers.
The mechanism for this is the setting of Handled which is then checked in Application's OnEvent function when it calls the layers' OnEvent functions.

Hazel/src/Platform/Windows/WindowsWindow.cpp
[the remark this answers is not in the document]
Puddlestomper: Both seem equally readable to me.

Hazel/src/Hazel/Application.cpp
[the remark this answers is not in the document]
LovelySanta: In `Application::OnEvent(...)` we dispatch the event first to application itself. Returning true in the event function will set `Event::Handled` to true. When iterating to the layerstack, after the first layer, it will be set to true and break out, resulting in distributing the event only to 1 layer, independed if it is handled or not.

We should add an if check before entering the for loop, in case application handled it (for example the close event)

Hazel/src/Platform/Windows/WindowsWindow.cpp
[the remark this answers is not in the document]
LovelySanta: Instead of saying "if we are in windowed mode we set it to nullptr" am I saying "if we are not in windowed mode, we have to provide a screen"... Both are equaly readable yes, but somehow it shows some logic behind it... Doesn't have any code impact tho...

Hazel/src/Hazel/Application.cpp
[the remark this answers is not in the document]
Puddlestomper: Alternatively we could have the loop check before propagating the event, instead of propagating before checking. This also avoids the check after the last layer which is unnecessary.

Hazel/src/Hazel/Window.h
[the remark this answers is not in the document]
LovelySanta: was more a reference while I was coding, the link can be removed...

Hazel/src/Hazel/Application.cpp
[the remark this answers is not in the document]
LovelySanta: I was about to write that, it will look nicer yes, i was thinking of the iterator incrementing first, but we can indeed fix this.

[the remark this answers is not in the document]
CleverSource: I think if it's nicer for the "end-programmer" to use then I wouldn't consider it micro optimization and more so a better experience for the developer. But I think that's for another time.

Hazel/src/Platform/Windows/WindowsWindow.cpp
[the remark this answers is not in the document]
CleverSource: I think this change is useless, both are fine and both make sense.

Hazel/src/Hazel/Window.h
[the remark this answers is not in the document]
CleverSource: I'll remove the link as the definitions are already on the other comments I guess.

Hazel/src/Hazel/Application.cpp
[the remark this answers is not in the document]
LovelySanta: Create an issue for it in the meantime, as it is a bug that it passes close events to the layers... it should never go into the for loop to begin with...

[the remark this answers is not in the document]
CleverSource: Other than that would this be available for merge?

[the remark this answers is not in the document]
LovelySanta: I first have to have time to test it fully... probably next week (as my pc will have downtime due to yearly maintenance comming up)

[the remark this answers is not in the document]
CleverSource: I'd be glad to test what is needed if you can specify what needs to be tested.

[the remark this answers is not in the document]
LovelySanta: - test switching from each mode to each mode
- test if each event is fired correctly

WhoseTheNerd:
I really doubt this can even be merged, the `SetWindowMode` is taking too long. 8000 us to be exact.
My visual studio 2019 profiler is broken right now.
https://prnt.sc/ozxyr5

Hazel/src/Hazel/Application.cpp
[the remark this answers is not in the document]
CleverSource: Can confirm you can switch between modes.

CleverSource:
@LovelySanta he should run his test on a recent clone of the Hazel repo and grab the branch from this PR. His test was done on his personal engine which he has implemented 3D in and etc.

LovelySanta:
> @LovelySanta he should run his test on a recent clone of the Hazel repo and grab the branch from this PR. His test was done on his personal engine which he has implemented 3D in and etc. Which does not give a good example for how this affects Hazel compared to his personal engine.

I'll add the testing to this branch, quickly add the timer that cherno recently made in the cpp series.

CleverSource:
> quickly add the timer that cherno recently made in the cpp series.

Do you want me to add the timer or are you saying you will?

LovelySanta:
> > quickly add the timer that cherno recently made in the cpp series.
> 
> Do you want me to add the timer or are you saying you will?

I was planning on doing it, but I didn't fine time to do yet, feel free to do so, but don't commit it, just put results from like 5 test, and between each mode...

LovelySanta:
As you saw in cherno's [Day in the Life - Hazel 2D Dev Vlog video](https://www.youtube.com/watch?v=2njnCUau3zk), cherno will fix the bug part of this PR (the resizing). I'll put this on hold until that is implemented, we can go from there about how to proceed.

PR #111, opened 2019-08-27
Made Input a Singleton

### LINKED ISSUE #97: Singleton Input class
``Input`` class is a singleton. Therefore, its copy constructor and copy operator is supposed to be ``delete`` functions to prevent from copying this class. 
However, in ``WindowsInput.cpp``, the line,

``Input* Input::s_Instance = new WindowsInput();`` 

is trying to use copy operator(which I have ``void operator=(const Input&) = delete;``) in ``Input`` class. 

If an ``Input`` constructor is explicitly declared and defined, it seems like this problem will be solved, but I not sure how this is working.

Puddlestomper:
Implemented the Singleton design pattern for the Input class by deleting copy constructor and copy-assignment operator and adding protected default constructor. Credit to @moskittle for pointing it out.

Resolves #97

### DIFF
--- a/Hazel/src/Hazel/Input.h
+++ b/Hazel/src/Hazel/Input.h
@@ -6,7 +6,12 @@ namespace Hazel {
 
 	class HAZEL_API Input
 	{
+	protected:
+		Input() = default;
 	public:
+		Input(const Input&) = delete;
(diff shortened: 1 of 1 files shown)


LovelySanta:
LGTM

PR #112, opened 2019-08-28
All Create() functions now return Ref<> instead of a pointer.

K-u-b-i-s-e-k:


Puddlestomper:
I noticed your use of `std::make_shared`. I think Cherno mentioned in the video that the point of having the `Ref` template/alias was so that in the future we could potentially replace it with our own reference counting smart pointer. So I think we should then also have a `MakeRef` or something similar so that if we do end up doing that it doesn't break every instance of a `Ref` in Hazel. You could probably implement it in the same way we implemented `Ref`.

Hazel/src/Hazel/Core.h
```diff
@@ -34,10 +34,14 @@
 
 namespace Hazel {
 
-	template<typename T>
-	using Scope = std::unique_ptr<T>;
-
-	template<typename T>
-	using Ref = std::shared_ptr<T>;
+    template<typename T>
+    const auto MakeScope = std::make_unique<T>;
```
Puddlestomper: This implementation does not currently compile for me. From my research it turns out that aliasing template functions is not the easiest thing to do. I am by no means experienced in using templates, but this seems to work:
```c++
template<typename T, typename... Args>
constexpr inline Scope<T> MakeScope(Args&&... args)
{
	return std::make_unique<T>(std::forward<Args>(args)...);
}
```
Basically instead of a straight alias, we are forwarding the function call. This might have a performance cost, but probably not a significant one. Of course this also needs to be placed after the `Scope` aliasing and then you can do the same for `Ref`, but with `MakeRef` and `make_shared`.

This compiles, but I am not experienced enough to know whether this is safe for every possible type or how it exactly works. I would like someone more experienced confirm that this is a valid design before merging.

Hazel/src/Hazel/Core.h
```diff
@@ -34,10 +34,14 @@
...
+    template<typename T>
+    using Scope = std::unique_ptr<T>;
+
+    template<typename T>
+    const auto MakeRef = std::make_shared<T>;
```
Puddlestomper: Same problem as with `MakeScope`.

Hazel/src/Hazel/Core.h
LovelySanta:
```cpp
	template<typename T>
	using MakeScope = std::make_unique<T>;
	template<typename T>
	using Scope = std::unique_ptr<T>;

	template<typename T>
	using MakeRef = std::make_shared<T>;
	template<typename T>
	using Ref = std::shared_ptr<T>;
```

Also: convert spaces to tabs to be inline with our coding style.

Hazel/src/Hazel/Core.h
```diff
@@ -34,14 +34,19 @@
...
+	{
+		return std::make_unique<T>(std::forward<Args>(args)...);
+	}
+
+	template<typename T>
+	using Ref = std::shared_ptr<T>;
+	template<typename T, typename... Args>
+	constexpr inline Ref<T> CreateRef(Args&&... args)
+	{
+		return std::make_unique<T>(std::forward<Args>(args)...);
```
WhoseTheNerd: Why is this code running make_unique instead of make_shared?

Hazel/src/Hazel/Core.h
```diff
@@ -36,8 +36,17 @@ namespace Hazel {
 
 	template<typename T>
 	using Scope = std::unique_ptr<T>;
+	template<typename T, typename... Args>
+	constexpr inline Scope<T> CreateScope(Args&&... args)
```
ReDucTor: constexpr implies inline, so potentially redundant having both

Hazel/src/Hazel/Core.h
ReDucTor: std::shared_ptr has a non-explicit constructor which accepts an std::unique_ptr so this works but will result in the control block being allocated separately, this should be using std::make_shared

Hazel/src/Hazel/Core.h
Puddlestomper: You are correct! The inline can be removed for both.

Puddlestomper:
An interesting thing I noticed is that in `Window.h` we also have a Create function which is implemented in `WindowsWindow.cpp`. For this function though we can't just blindly make it a `Ref` because `Application` currently stores the window as a `std::unique_ptr`.

We have a few options:

1. Let it return a raw pointer as it does currently.
2. Let it return a `Ref` as all the other create functions do and let Application somehow cast it?
3. Let it return a `Scope`.

I am not sure which option is best. It would depend on whether someone using Hazel could directly construct a `Window` I suppose. Currently the only `Window` is managed by Hazel and not the user.

ReDucTor:
With a window, you potentially only want a single owner (the application), so a shared pointer for it is probably not appropriate, an additional option is to return it by value and make it move-only then can transfer the ownership out of the creation function back into the application, it doesn't need to live on the heap or change ownership semantics dramatically.  

However this does increase the complexity slightly as then you need to consider what is required for a move-only type (move constructor, move assignment, implications of valid but unspecified, etc) for a beginner introduction this might just be to much noise so the 'Scope' option probably comes out on top.

LovelySanta:
> An interesting thing I noticed is that in `Window.h` we also have a Create function which is implemented in `WindowsWindow.cpp`. For this function though we can't just blindly make it a `Ref` because `Application` currently stores the window as a `std::unique_ptr`.

It should be a scope indeed.

LovelySanta:
Another thing, we create refs instead of pointers, but then, when adding it to the vertexArray, we still pass in pointers, those should also be converted, and then it starts tickling down the line... I start to think this should all be part of the code refactor instead...

PR #115, opened 2019-08-30
Implemented Missing Events

Puddlestomper:
- Implemented callbacks and Event creation for the `WindowMoved`, `WindowFocused` and `WindowFocusLost` events.
- Renamed Event names to past tense for a uniform naming scheme

I want to mention that we should consider adding the following events:
1. `WindowIconified` and `WindowRestored` for when the Window is minimised/restored.
2. `MouseDragged` which will be triggered when a mouse button is held while the mouse is moving and could optionally prevent the normal `MouseMoved` event.

These technically don't broadcast any information that couldn't otherwise be deduced from the other events, but I think they will make code easier and cleaner for anyone trying to use the Event system in their application.

The App Events I have left for when Cherno has expanded the Hazel infrastructure and made it clearer when exactly each of these should be created. I initially removed `AppUpdated` because Layers have an `OnUpdate` method anyway, and it seems like a good idea to have dedicated methods for Events that are called as fast as possible or at least multiple times per second, but I think 
that is something for later on when Cherno has made it clear what he plans for these Events.


LovelySanta:
Not merging this as cherno initially was not sure if we needed these events. When some Hazel game actually needs it, we can implement them. As of now, we have no use for it.

PR #116, opened 2019-09-01
Unified logging macro between client and engine

bagelbite:
**Proposed Change:**

HZ_LOG_* (Trace, Info, etc) is used, and the use case will determine if it is Core or App. 

**Changes Made:**

All HZ_CORE_* (Trace, Info, etc) and HZ_* (Trace, Info, etc) calls have been replaced with HZ_LOG_* (Trace, Info, etc).

This change also makes HZ_CORE_ASSERT unnecessary, so HZ_CORE_ASSERT has been removed. All HZ_CORE_ASSERT calls have been replaced by HZ_ASSERT calls.


Hazel/src/Hazel/Log.h
```diff
@@ -20,16 +20,22 @@ namespace Hazel {
...
-#define HZ_CORE_ERROR(...)    ::Hazel::Log::GetCoreLogger()->error(__VA_ARGS__)
-#define HZ_CORE_CRITICAL(...) ::Hazel::Log::GetCoreLogger()->critical(__VA_ARGS__)
-
-// Client log macros
-#define HZ_TRACE(...)         ::Hazel::Log::GetClientLogger()->trace(__VA_ARGS__)
-#define HZ_INFO(...)          ::Hazel::Log::GetClientLogger()->info(__VA_ARGS__)
-#define HZ_WARN(...)          ::Hazel::Log::GetClientLogger()->warn(__VA_ARGS__)
-#define HZ_ERROR(...)         ::Hazel::Log::GetClientLogger()->error(__VA_ARGS__)
-#define HZ_CRITICAL(...)      ::Hazel::Log::GetClientLogger()->critical(__VA_ARGS__)
+#ifdef HZ_BUILD_DLL
```
Puddlestomper: Not sure if we should be using this define still, seeing as we are statically linking now and we aren't actually building a dll. Maybe we can introduce a new define (in premake5.lua) for this? Something like just `HZ_BUILD`.

bagelbite: It still worked as intended for me. However if it makes more sense to change it then I agree.

LovelySanta: I would suggest `HZ_BUILD_ENGINE` and derive the (unused) DLL part from that.

LovelySanta: I would suggest HZ_BUILD_ENGINE and derive the (unused) DLL part from that.

LovelySanta: I would suggest HZ_BUILD_ENGINE and derive the (unused) DLL part from that.

CleverSource: Agreeing with this.

bagelbite: I’ll make this change next time I have a chance.

bagelbite:
I have changed HZ_BUILD_DLL to HZ_BUILD_ENGINE and added it to the premake. Now the define is checking to see if it is the Engine, and no longer using the DLL tag.

CleverSource:
You should remove HZ_BUILD_DLL from the defines on premake, DLL support is fully deprecated on Hazel and a DLL won't even build currently. To be honest, it wouldn't hurt to fully remove anything DLL related including HAZEL_API

bagelbite:
I agree that the remaining HZ_BUILD_DLL should be removed and removed from premake. Not 100% sure about HAZEL_API though.

LovelySanta:
> I agree that the remaining HZ_BUILD_DLL should be removed and removed from premake. Not 100% sure about HAZEL_API though.

1. HZ_BUILD_DLL should be removed since we don't build dynamicaly anymore
2. HAZEL_API isn't added in the new classes, so this isn't supported anyway, so this should be removed as well
3. Removing dll support code from hazel should be a seperate PR.

bagelbite:
> > I agree that the remaining HZ_BUILD_DLL should be removed and removed from premake. Not 100% sure about HAZEL_API though.
> 
> 1. HZ_BUILD_DLL should be removed since we don't build dynamicaly anymore
> 2. HAZEL_API isn't added in the new classes, so this isn't supported anyway, so this should be removed as well
> 3. Removing dll support code from hazel should be a seperate PR.

That’s what I was thinking as well. I feel like for the scope of my PR, it is complete.

ReDucTor:
Sometimes with debug levels you want to be able to remove them at compile time as they can make reverse engineering easier, and make it easier to create cheats. While at the same time you also want the ability to have logs which will help inform you when a user has issues, so sometimes it can be useful to have a heavier debug level for development and testing purpose, then a lower debug level which is released to players.

Hazel/src/Hazel/EntryPoint.h
```diff
@@ -7,9 +7,9 @@ extern Hazel::Application* Hazel::CreateApplication();
 int main(int argc, char** argv)
 {
 	Hazel::Log::Init();
-	HZ_CORE_WARN("Initialized Log!");
+	HZ_LOG_WARN("Initialized Log!");
```
lhurt51: I am not entirely sure if it is intentional or not but these log messages get logged on the APP side instead of the Hazel Engine side because it is compiled in the Sandbox project.

LovelySanta: Yeah, I think the intent was good, but the macros only get resolved at compile time, same goes for templated classes, those could potentialy have the same issue... I would say leave them seperated into two...

LovelySanta:
As the review https://github.com/TheCherno/Hazel/pull/116#discussion_r331773083 said, this won't realy work as it is... Also this would limit us between client and core what and what not to log when we enable/disable logging levels.

I'll close this for now, maybe in the future we might come up with a beter solution. Good first issue tho @bagelbite!

PR #121, opened 2019-09-12
File path correction for developmental builds

DaveAxiom:
Added a function to correct the file path when running with the wrong current working directory, such as when Sandbox is run outside of an IDE.
@LovelySanta

### DIFF
--- a/Hazel/src/Hazel/Application.cpp
+++ b/Hazel/src/Hazel/Application.cpp
@@ -78,4 +78,23 @@ namespace Hazel {
 		return true;
 	}
 
+	std::string Application::CorrectFilePath(const std::string& path)
+	{
+		#if defined(HZ_DEBUG) || defined(HZ_RELEASE)
+			struct stat buffer;
+			int status;
+
+			std::vector<std::string>prepath_string_vector = {"", "./Sandbox/", "../../../Sandbox/"};
+			for (auto path_iter : prepath_string_vector)
+			{
+				std::string modified_path = path_iter + path;
+				status = stat(modified_path.c_str(), &buffer);
+
+				if (status == 0)
+					return modified_path;
+			}
+		#endif
+
+		return path;
+	}
 }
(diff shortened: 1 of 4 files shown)


Hazel/src/Hazel/Application.cpp
```diff
@@ -78,4 +78,23 @@ namespace Hazel {
...
 
-}
\ No newline at end of file
+	std::string Application::CorrectFilePath(const std::string& path)
+	{
+		#if defined(HZ_DEBUG) || defined(HZ_RELEASE)
+			struct stat buffer;
+			int status;
+
+			std::vector<std::string>prepath_string_vector = {"", "./Sandbox/", "../../../Sandbox/"};
```
SeppahBaws: Doesn't look like a great idea to me to hard-code the name/directory of the application in the engine itself.
If we made a game with Hazel then we would have to change this line of engine code in order to run the game.

Hazel/src/Hazel/Application.cpp
```diff
@@ -78,4 +78,23 @@ namespace Hazel {
...
+			for (auto path_iter : prepath_string_vector)
+			{
+				std::string modified_path = path_iter + path;
+				status = stat(modified_path.c_str(), &buffer);
```
WhoseTheNerd: `stat()` is linux specfic function.

Hazel/src/Hazel/Application.cpp
DaveAxiom: That's right. Got me. Hmmm.

Hazel/src/Hazel/Application.cpp
DaveAxiom: It's a Posix C function. @LovelySanta tested it for me.

std::filesystem is giving me segfaults on my Linux system. I need to escalate that issue. GNU usually is very careful with testing though. I'll track it down.

PR #122, opened 2019-09-12
Adding VS files to gitignore

LovelySanta:
Since we've added again some VS files to the repo, I took the time to add them into the gitignore to prevent this in the future.
I also excluded the premake5 exe from the gitignore as it is required (for now). We're working on compiling premake5 from source for the future instead.


Puddlestomper:
Should we also remove the `imgui.ini` file in the Sandbox directory?

LovelySanta:
> Should we also remove the `imgui.ini` file in the Sandbox directory?

No, cherno wants to have this committed. When we start using ImGui more intensively, we will create our own default ini files.

PR #123, opened 2019-09-14
File path correction for developmental builds 2

DaveAxiom:
Added a function to correct the file path when running with the wrong current working directory, such as when Sandbox is run outside of an IDE.

New in this second attempt PR is a switch to `std::filesystem` of C++17 away from POSIX C and removal of the hard coding of _Sandbox_ in the Hazel code base.

Linux Note: This patch needs `libstdc++fs` (_"stdc++fs"_) added to the premake config for GCC versions 5.3 to 8.x support of `std::filesystem`.

### DIFF
--- a/Hazel/src/Hazel/Application.h
+++ b/Hazel/src/Hazel/Application.h
@@ -29,6 +29,7 @@ namespace Hazel {
 		inline Window& GetWindow() { return *m_Window; }
 
 		inline static Application& Get() { return *s_Instance; }
+		static std::string CorrectFilePath(const std::string&);
 	private:
 		bool OnWindowClose(WindowCloseEvent& e);
 	private:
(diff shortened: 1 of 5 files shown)


Hazel/src/Hazel/Application.h
```diff
@@ -39,9 +40,10 @@ namespace Hazel {
 		float m_LastFrameTime = 0.0f;
 	private:
 		static Application* s_Instance;
+		std::string m_base_directory;
```
ReDucTor: Why not use `std::filesystem::path`? 

Then all your appends (`p1 / p2`) can be simplified for path can be simplified, tbh with the 3 of them I'd probably unroll the loop and check them explicitly commenting the reason for each.

DaveAxiom: The parameter is turned into `std::filesystem::path` within `std::filesystem::exists()`. I don't see a prototype for `char*` so maybe C++17 automatically uses a `path` operator to do a cast-conversion? I don't see the advantage for rewriting this segment. I'm new to C++17 which is why I needed to use the new `for()` iterator.

PR #124, opened 2019-09-15
Replaced the KeyCodes and MouseButtonCodes defines with enum classes

IshoAntar:
Replaced all occurrences of "int keycode" with "KeyCode keycode", and "int button" with "MouseButtonCode button", for more type-safety and readability; also included the KeyCodes.h and MouseButtonCodes.h in the Input.h header file instead of including them separately since they are related to input.


Hazel/src/Hazel/KeyCodes.h
```diff
@@ -1,125 +1,128 @@
...
-#define HZ_KEY_LEFT_ALT           342
-#define HZ_KEY_LEFT_SUPER         343
-#define HZ_KEY_RIGHT_SHIFT        344
-#define HZ_KEY_RIGHT_CONTROL      345
-#define HZ_KEY_RIGHT_ALT          346
-#define HZ_KEY_RIGHT_SUPER        347
-#define HZ_KEY_MENU               348
\ No newline at end of file
+	/* Function keys */
+	HZ_KEY_ESCAPE        = 256,
```
LovelySanta: Since they are not macro's anymore, the `HZ_` prefix can be ommitted.

Hazel/src/Hazel/MouseButtonCodes.h
```diff
@@ -1,15 +1,18 @@
...
-#define HZ_MOUSE_BUTTON_8         7
-#define HZ_MOUSE_BUTTON_LAST      HZ_MOUSE_BUTTON_8
-#define HZ_MOUSE_BUTTON_LEFT      HZ_MOUSE_BUTTON_1
-#define HZ_MOUSE_BUTTON_RIGHT     HZ_MOUSE_BUTTON_2
-#define HZ_MOUSE_BUTTON_MIDDLE    HZ_MOUSE_BUTTON_3
\ No newline at end of file
+enum class MouseButtonCode
+{
+	// From glfw3.h
+	HZ_MOUSE_BUTTON_1      = 0,
```
LovelySanta: Since they are not macro's anymore, the `HZ_` prefix can be ommitted.

IshoAntar:
OK, i have omitted the HZ_ prefix :)

Sandbox/src/SandboxApp.cpp
```diff
@@ -139,19 +139,19 @@ class ExampleLayer : public Hazel::Layer
 
 	void OnUpdate(Hazel::Timestep ts) override
 	{
-		if (Hazel::Input::IsKeyPressed(HZ_KEY_LEFT))
+		if (Hazel::Input::IsKeyPressed(KeyCode::KEY_LEFT))
```
LovelySanta: KeyCodes and MouseButtonCodes should be moved into the Hazel namespace like all Hazel stuff.

IshoAntar:
Done, sorry i should have done this from the beginning.

LovelySanta:
Looks fine I think. More of a question, should we allign the numbers (using spaces) like it was with the macro's? That way it looks more readable (in my opinion at least), and the comments will be aligned then as well. Thoughts?

Puddlestomper:
I just want to mention that this also makes the API quite a bit more clumsy on the application's side. `HZ_MOUSE_BUTTON_LEFT` is now `Hazel::MouseButtonCode::MOUSE_BUTTON_LEFT`. Maybe we could shorten it somehow?

LovelySanta:
You could keep the macro's as `#define HZ_MOUSE_BUTTON_LEFT ::Hazel::MousebuttonCode::MOUSE_BUTTON_LEFT`.

Now that I'm writing it, they could be shortened quite a bit:
`KeyCode::KEY_L` -> `KeyCode::L`
`MousebuttonCode::MOUSE_BUTTON_LEFT` -> `MouseCode::Button_Left`
same for all of those? And ommit the all caps?

Puddlestomper:
The all caps should probably go and become camel case to keep in line with Cherno's conventions for enums as is seen in `EventType`. I agree with removing the "Mouse" and "Key" prefixes seeing as they are already present in the enum class' name.

I like the idea of keeping the macros and making them just expand to the new enums. It seems to give both the type safety advantages and the readability bonus.

IshoAntar:
i can do that, but there is a problem with removing the `KEY_` prefix in the `KeyCode` enum and that is the number keys can't be represented that way. like `KEY_0` would become `0` which isn't possible, maybe leave the `_` as a prefix?

Puddlestomper:
We could make it `ZERO` instead.

LovelySanta:
If we look, we keep `ButtonLeft` at the mouse, and tbh.. `Key0` Doesn't look that bad?

Puddlestomper:
I would prefer `ZERO`, but I think it really comes down to preference.

IshoAntar:
OK, so the todo list would be:
1. remove the `KEY_` prefix
2. use camel-case for the names
3. use the actual number name for numbers
4. keep the macro's defined but substitute the actual enum values instead of plain numbers.
did i miss something?

Puddlestomper:
We could also remove the `Button` prefix?

IshoAntar:
You mean the `MOUSE_BUTTON_*`?

Puddlestomper:
Yes, in the `MouseButtonCode` enum class members.

LovelySanta:
> OK, so the todo list would be:
> 
> 1. remove the `KEY_` prefix
> 2. use camel-case for the names
> 3. use the actual number name for numbers
> 4. keep the macro's defined but substitute the actual enum values instead of plain numbers.
>    did i miss something?

Don't use actual numbers... `ZERO` looks so bad, it's not like we write characters full out as we pronounce them.. (eg the W)...
Like @Puddlestomper says.. probably preferences, but I think keeping 3 characters for the word 'Key' aint that bad... We keep 'Button' as well...

Like I previously said, I think this looks the best between all:

> ```c++
> ::Hazel::KeyCode::Key0
> ::Hazel::KeyCode::KeyA
> ::Hazel::MouseCode::Button0
> ::Hazel::MouseCode::ButtonLast
> ```

That way, all KeyCodes have a prefix Key and all MoseCodes have a prefix Button. It also will look good when auto completing in VS for example... (but that's more preference again)

IshoAntar:
Not a problem.
One more question, should i keep the macro names as they were before, or use the same names for them as well? maybe separated by underscores?

LovelySanta:
> One more question, should i keep the macro names as they were before, or use the same names for them as well? maybe separated by underscores?

I would say keep the macro names the same as they are, as they are not bound to any enum, there is no distinction between them...

On another note: the macro's have there number nicely aligned, should we do this for the enum as well? That way the comments will become more readable (as it was with the macros)

IshoAntar:
OK, i'll do that

Hazel/src/Hazel/Events/KeyEvent.h
```diff
@@ -1,34 +1,35 @@
...
+		KeyPressedEvent(KeyCode keycode, int repeatCount)
 			: KeyEvent(keycode), m_RepeatCount(repeatCount) {}
 
 		inline int GetRepeatCount() const { return m_RepeatCount; }
 
 		std::string ToString() const override
 		{
 			std::stringstream ss;
-			ss << "KeyPressedEvent: " << m_KeyCode << " (" << m_RepeatCount << " repeats)";
+			ss << "KeyPressedEvent: " << static_cast<uint32_t>(m_KeyCode) << " (" << m_RepeatCount << " repeats)";
```
ReDucTor: Instead of this it might be worth overloading operator<<(std::ostream&,KeyCode) for the enum

Hazel/src/Hazel/MouseButtonCodes.h
```diff
@@ -1,15 +1,34 @@
...
+		Button6                = 6,
+		Button7                = 7,
+		ButtonLast             = Button7,
+		ButtonLeft             = Button0,
+		ButtonRight            = Button1,
+		ButtonMiddle           = Button2
+	};
+}
+
+#define HZ_MOUSE_BUTTON_0      ::Hazel::MouseCode::Button0
```
ReDucTor: Will there be a pass to replace these? Or are they only kept for backwards compatibility?

Hazel/src/Hazel.h
```diff
@@ -9,7 +9,6 @@
 #include "Hazel/Core/Timestep.h"
 
 #include "Hazel/Input.h"
-#include "Hazel/KeyCodes.h"
 #include "Hazel/MouseButtonCodes.h"
```
LovelySanta: Here you either: omit the inclusion of MouseButtonCodes or don't remove the KeyCodes include. I would opt for not relying on other files including it, and go with the latter solution

Hazel/src/Hazel/MouseButtonCodes.h
LovelySanta: I would opt to wrap them inside a ifdef statement for those who doesn't want them... For now I would leave them in, we can always remove them in the future...

Hazel/src/Platform/Windows/WindowsInput.cpp
```diff
@@ -8,17 +8,17 @@ namespace Hazel {
 
 	Input* Input::s_Instance = new WindowsInput();
 
-	bool WindowsInput::IsKeyPressedImpl(int keycode)
+	bool WindowsInput::IsKeyPressedImpl(KeyCode keycode)
 	{
 		auto window = static_cast<GLFWwindow*>(Application::Get().GetWindow().GetNativeWindow());
-		auto state = glfwGetKey(window, keycode);
+		auto state = glfwGetKey(window, static_cast<uint32_t>(keycode));
```
WhoseTheNerd: You are converting enum to uint32 then there is implicit conversion from uint32 to int32

IshoAntar: You're right, i should've payed attention to that.

Hazel/src/Hazel/Events/KeyEvent.h
LovelySanta: Now that the overload operator is implemented (in https://github.com/TheCherno/Hazel/pull/124/commits/e10bfd761b2b43d6540236e13c0e38f0515260ae), you don't have to static cast anymore when using the `<<` operator, so you can remove those casts as well.

Issue #47, opened 2019-03-04
Memory leak in Hazel?


CleverSource:
This is what I had to say on the Discord:
![image](https://user-images.githubusercontent.com/21996470/53761891-aa9cde80-3e94-11e9-8af0-889042037e41.png)

I don't know what causes this and am still debugging it but I have managed to experience this memory leak 2 out of 10 runs in a row. This is a definite memory leak, just I can't find the root cause. I would love to discuss more of this with people and discover possible solutions, and I want to bring this to Cherno's attention. 

Thanks.👍 

# Task list
- [x] Delete created shaders after created program (TheCherno/Hazel#126)
- [x] Delete created shaders after created program (TheCherno/Hazel#127)
- [x] Call GLFWterminate (TheCherno/Hazel#128)
- [x] Use scopes instead of raw pointers (TheCherno/Hazel#129)
- [ ] Test if there are more leaks

EDIT by @LovelySanta: Added task list

gabrielevierti:
What exactly are you experiencing? Have you tried running the program in debug mode? Did you find something interesting we might wanna take a look at? In all Honesty I haven't had any sort of problem so far :(

badryasuo:
I'm having the same problem!! Memory leaks just by moving the demo window or moving it outside of the application + it drops a lot of FPS & lags when moving the demo window as you can see.. 16 FPS; That's definitely a bug either in Hazel or ImGui docking branch. p.s I'm using the latest commit.
![image](https://user-images.githubusercontent.com/30192597/57565001-b8e80900-73b5-11e9-90c9-0d6b1c9df28b.png)

badryasuo:
Sandbox.exe has used over 7GB of my RAM when I've left the demo window outside of the viewport for some minutes. I confirm it's a **bug**.

live627:
This might be fixed in #49

Gaztin:
Can anyone who had this problem before confirm if it has been fixed or not? @CleverSource or @badryasuo or anyone else.

gold604:
@Gaztin I can confirm there's no memory leak on my machine. However there was a missing semicolon introduced in #49. I've already created new PR #72 which is waiting to be merged.

gold604:
I'd like to note that #49 introduced a missing semicolon. I've already created new PR #72 which fixes that and is ready to be merged. 

For now, you can manually add a semicolon in LayerStack.cpp at the end of the line 34.

CleverSource:
[...]

PR #125, opened 2019-09-15
A Couple of memory leaks caused by left-over GL-Objects

IshoAntar:
Since i've seen a couple of issues about memory leaks (i was never able to reproduce myself) i wanted to give it a chance to test if there is actually any leaks in the Engine, so i attached the Sandbox.exe to CodeXL and ran it, didn't notice anything unusual; But, after terminating the application, i got this:

Screenshot_1

so i started looking around to check if there is any GL-Objects left over, and there was!
in OpenGLShader.cpp (line 160) in the for-loop we are detaching the shaders from the program but never actually deleting them! and since there isn't any stored reference to them, they're leaked after the function returns, so to fix it i added a `glDeleteShader(id)` in the loop, and re-ran the test, and:

Screenshot_3

as we can see there is still 6 objects being leaked, i have no idea where from but i'm still investigating the issue; Any ideas?

### DIFF
--- a/Hazel/src/Hazel/Input.h
+++ b/Hazel/src/Hazel/Input.h
@@ -26,7 +26,7 @@ namespace Hazel {
 		virtual float GetMouseXImpl() = 0;
 		virtual float GetMouseYImpl() = 0;
 	private:
-		static Input* s_Instance;
+		static Hazel::Scope<Input> s_Instance;
 	};
 
 }
--- a/Hazel/src/Hazel/LayerStack.cpp
+++ b/Hazel/src/Hazel/LayerStack.cpp
@@ -10,7 +10,10 @@ namespace Hazel {
 	LayerStack::~LayerStack()
 	{
 		for (Layer* layer : m_Layers)
+		{
+			layer->OnDetach();
 			delete layer;
+		}
 	}
 
 	void LayerStack::PushLayer(Layer* layer)
--- a/Hazel/src/Hazel/Renderer/RenderCommand.cpp
+++ b/Hazel/src/Hazel/Renderer/RenderCommand.cpp
@@ -5,6 +5,6 @@
 
 namespace Hazel {
 
-	RendererAPI* RenderCommand::s_RendererAPI = new OpenGLRendererAPI;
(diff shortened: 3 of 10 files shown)


IshoAntar:
OK so here are the objects being leaked

Screenshot_4

Puddlestomper:
I was going to make a PR later about some of the memory leaks I know about, but this is fine too. The memory leaks and fixes I am aware of are:

1. `Input::s_Instance` is never deleted. It should be made a `Scope`.
2. `RenderCommand::s_RendererAPI` is never deleted. It should be made a `Scope`.
3. `Renderer::s_SceneData` is never deleted. It should be made a `Scope`.
4. We are not detaching layers before deleting them in `LayerStack`'s destructor meaning the imgui Shutdown methods are never called. Just call `OnDetach` before deleting each layer in the destructor of `LayerStack`.
5. The known issue of us not calling `glfwTerminate`. A temporary fix could be the the solution propsed by @LovelySanta [here](https://github.com/TheCherno/Hazel/issues/47#issuecomment-515569314).
6. `WindowsWindow::m_Context` is never deleted. It should be made a `Scope`.

I used Visual Leak Detector to find these, and after fixing them no further leaks were detected. This was done before the last two commits though and I don't fully trust Visual Leak Detector to pick every leak up.

Try implementing these and seeing if you pick anything else up?

IshoAntar:
OK, i'll do that and see if it fixes the problem

IshoAntar:
OK, so for me, i'm not getting any leaks now. looks like the problem is fixed, probably?

LovelySanta:
This look realy promissing, I like all of the changes, but please make seperate PR's for each issue:

1. Converting to scopes so they get deleted (they can all be in 1 PR)
2. Detaching the layer before deleting them in layerstack
3. Deleting the created shaders when the program was successfully created
4. GLFW window counting:
    - Leave the `glfwInit()` inside the init function
    - Move the `glfwShutdown()` inside the shutdown function
    - You check if the count is <= 0, just check == 0 as it is an `uint8_t` variable anyway...

In all of them, mension issue 47 (intentionaly not linked here)

PR #126, opened 2019-09-15
Call the layer's OnDetach before deleting it.

IshoAntar:
the leaks where discovered in #125


Puddlestomper:
I would just like to point out that this does not resolve the issue alone. Only with the other 3 PRs of yours is the issue resolved.

IshoAntar:
How do i point that out?

Puddlestomper:
I suggest you remove the closing keywords and we make a task list on the issue itself. I have suggested it there.

IshoAntar:
OK, i'll edit my comments.

PR #127, opened 2019-09-15
Delete the shaders after the shader program Creation

IshoAntar:
After successfully creating the shader program, the shaders should be deleted since they're probably not needed anymore.
the leaks where discovered in #125


LovelySanta:
It should indeed be deleted as the shader (and the id) is only present inside the scope of the`OpenGLShader::Compile(...)` function.

PR #128, opened 2019-09-15
Implemented a basic ref-counting system to terminate glfw

IshoAntar:
this was a known issue, and was fixed in #125

### DIFF
--- a/Hazel/src/Platform/Windows/WindowsWindow.cpp
+++ b/Hazel/src/Platform/Windows/WindowsWindow.cpp
@@ -9,7 +9,7 @@
 
 namespace Hazel {
 	
-	static bool s_GLFWInitialized = false;
+	static uint8_t s_GLFWWindowCount = 0;
 
 	static void GLFWErrorCallback(int error, const char* description)
 	{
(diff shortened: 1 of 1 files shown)


ReDucTor:
This seems like a the global init should be split from Window

LovelySanta:
> This seems like a the global init should be split from Window

This is how cherno did initialize it for now. I agree, we should have an abstract GLFW window base class inbetween Window and WindowWindow, but that is up to cherno to decide. For now this is a good implementation, as it was like this previously as well.

WhoseTheNerd:
I would recommend System class that initializes all required dependencies(dependencies that need to be manually initialized)

LovelySanta:
> I would recommend System class that initializes all required dependencies(dependencies that need to be manually initialized)

I agree, but that is more of a feature. Imo, this is still a valid "get things up and running" PR. When some sort of System is in place, this will have to be removed. I rather have glfw terminating correctly then not at all...

ReDucTor:
imho, I see this as technical debt over being a 'feature', this commit just compounding the technical debt.

This is an end-of-application clean-up so not really a memory leak just a false-positive with no side-effects that can show up in leak detection

LovelySanta:
> This is an end-of-application clean-up so not really a memory leak just a false-positive with no side-effects that can show up in leak detection

I know, I just want to work on closing that memory leak thread, the less results, the faster we can solve it.

PR #129, opened 2019-09-15
Use 'Scope's instead of raw pointers

IshoAntar:
s_Instance, s_RendererAPI, s_SceneData, and m_Context are now 'Scope's instead of raw pointers.
and was discovered in #125


Hazel/src/Platform/Windows/WindowsWindow.h
```diff
@@ -29,7 +29,7 @@ namespace Hazel {
 		virtual void Shutdown();
 	private:
 		GLFWwindow* m_Window;
-		GraphicsContext* m_Context;
+		Hazel::Scope<GraphicsContext> m_Context;
```
ReDucTor: Could use an std::optional here instead as GraphicsContext doesn't need to be accessed outside this class its just delay initialized (in an ideal situation you wouldn't have two-phase init)

Hazel/src/Hazel/Renderer/Renderer.h
```diff
@@ -24,7 +24,7 @@ namespace Hazel {
 			glm::mat4 ViewProjectionMatrix;
 		};
 
-		static SceneData* s_SceneData;
+		static Hazel::Scope<SceneData> s_SceneData;
```
ReDucTor: This doesn't need to be a pointer at all it can just be a normal static member variable, it requires no special ownership

Hazel/src/Hazel/Renderer/RenderCommand.h
```diff
@@ -27,7 +27,7 @@ namespace Hazel {
 			s_RendererAPI->DrawIndexed(vertexArray);
 		}
 	private:
-		static RendererAPI* s_RendererAPI;
+		static Hazel::Scope<RendererAPI> s_RendererAPI;
```
ReDucTor: This can also be a normal static member variable

Hazel/src/Hazel/Input.h
```diff
@@ -26,7 +26,7 @@ namespace Hazel {
 		virtual float GetMouseXImpl() = 0;
 		virtual float GetMouseYImpl() = 0;
 	private:
-		static Input* s_Instance;
+		static Hazel::Scope<Input> s_Instance;
```
ReDucTor: Another one which can be a normal static member variable

Hazel/src/Hazel/Renderer/RenderCommand.cpp
```diff
@@ -5,6 +5,6 @@
 
 namespace Hazel {
 
-	RendererAPI* RenderCommand::s_RendererAPI = new OpenGLRendererAPI;
+	Hazel::Scope<RendererAPI> RenderCommand::s_RendererAPI(new OpenGLRendererAPI);
```
WhoseTheNerd: instead of using `new` keyword use `make_unique`

Hazel/src/Platform/Windows/WindowsInput.cpp
```diff
@@ -6,7 +6,7 @@
 
 namespace Hazel {
 
-	Input* Input::s_Instance = new WindowsInput();
+	Hazel::Scope<Input> Input::s_Instance(new WindowsInput());
```
WhoseTheNerd: Same thing, `make_unique`

Hazel/src/Hazel/Renderer/Renderer.h
Puddlestomper: I think we should consider keeping it as a smart pointer, simply because Cherno decided to make this a pointer initially. He probably had a reason for doing that. On the other hand he might not have had a specific reason at all. I'm not sure either way.

ReDucTor: Also based on that logic could the same be said about any changes including your proposed changes?

I'm not certain that is the best logic to follow if the goal is to improve the engine, because functionally there is no difference between the two, just improved performance because your not chasing the pointer.

Hazel/src/Hazel/Renderer/RenderCommand.cpp
```diff
@@ -5,6 +5,6 @@
 
 namespace Hazel {
 
-	Hazel::Scope<RendererAPI> RenderCommand::s_RendererAPI(new OpenGLRendererAPI);
+	Hazel::Scope<RendererAPI> RenderCommand::s_RendererAPI = CreateScope<OpenGLRendererAPI>();
```
LovelySanta: Inside namespace Hazel, you can omit `Hazel::` from `Hazel::Scope<RendererAPI>`

Hazel/src/Hazel/Renderer/Renderer.cpp
```diff
@@ -5,7 +5,7 @@
 
 namespace Hazel {
 
-	Hazel::Scope<Renderer::SceneData> Renderer::s_SceneData(new Renderer::SceneData);
+	Hazel::Scope<Renderer::SceneData> Renderer::s_SceneData = CreateScope<Renderer::SceneData>();
```
LovelySanta: Same here, omit `Hazel::`

Hazel/src/Platform/Windows/WindowsInput.cpp
```diff
@@ -6,7 +6,7 @@
 
 namespace Hazel {
 
-	Hazel::Scope<Input> Input::s_Instance(new WindowsInput());
+	Hazel::Scope<Input> Input::s_Instance = CreateScope<WindowsInput>();
```
LovelySanta: Same here, omit `Hazel::`

Puddlestomper:
Should we also include the changes to Hazel's various `Create` methods in this PR (as was implemented in #112)?

Issue #130, opened 2019-09-16
Linux port progress


ppetraki:
Host is 18.04 Ubuntu x86_64
premake5 built from source as no packages exist.

Would you consider adding meson or build2 support?

https://github.com/ppetraki/Hazel

Sandbox tries build as standalone executable but it doesn't have a main().

Hacking the Sandbox app to CreateApplication in it's own main ends in fireworks.

```
Program received signal SIGSEGV, Segmentation fault.
0x000055555556d0c8 in std::__atomic_base<int>::load (__m=std::memory_order_relaxed, this=0x40) at /usr/include/c++/7/bits/atomic_base.h:396
396             return __atomic_load_n(&_M_i, __m);
(gdb) bt
#0  0x000055555556d0c8 in std::__atomic_base<int>::load (__m=std::memory_order_relaxed, this=0x40) at /usr/include/c++/7/bits/atomic_base.h:396
#1  spdlog::logger::should_log (this=0x0, msg_level=spdlog::level::info) at ../Hazel/vendor/spdlog/include/spdlog/details/logger_impl.h:333
#2  0x00005555555c96e9 in spdlog::logger::log<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, unsigned int, unsigned int> (this=0x0, lvl=spdlog::level::info, fmt=0x5555556db638 "Creating window {0} ({1}, {2})", args#0="Hazel Engine", args#1=@0x7fffffffd8e0: 1280, args#2=@0x7fffffffd8e4: 720) at vendor/spdlog/include/spdlog/details/logger_impl.h:59
#3  0x00005555555c9229 in spdlog::logger::info<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, unsigned int, unsigned int> (this=0x0, fmt=0x5555556db638 "Creating window {0} ({1}, {2})", args#0="Hazel Engine", args#1=@0x7fffffffd8e0: 1280, args#2=@0x7fffffffd8e4: 720) at vendor/spdlog/include/spdlog/details/logger_impl.h:138
#4  0x00005555555c7f36 in Hazel::WindowsWindow::Init (this=0x5555559b8f50, props=...) at src/Platform/Windows/WindowsWindow.cpp:40
#5  0x00005555555c7820 in Hazel::WindowsWindow::WindowsWindow (this=0x5555559b8f50, props=...) at src/Platform/Windows/WindowsWindow.cpp:26
[...]

PR #82, opened 2019-07-11
[DRAFT] Linux support (updated final)

LovelySanta:
"The final" PR that brings linux support to Hazel. Previously #13 and #27

###### _This needs to be done before this PR can become final_
- [ ] Add linux support to premake file in glfw
    - [ ] Update premake file in PR TheCherno/glfw#6
    - [ ] Update submodule in this PR
- [x] Add linux support to premake file in imgui
    - [ ] Update premake file in PR TheCherno/imgui#5 (optional, currently not using master branch)
    - [x] Update premake file in PR TheCherno/imgui#4 (docking branch)
    - [x] Update submodule in this PR

## Changes in this PR
- Refactored engine code to work on linux builds
    - `HZ_DEBUGBREAK()` macro instead of `__debugbreak()`
    - Update `.gitignore` to add linux build files
    - Seperate entrypoint for linux build
    - Moving `#include<glad/glad.h>` around to be included before `#include<GLFW/glfw3.h>`
    - Initializing OpenGL4.5
        - In ImGuiLayer
        - In WindowsWindow
    - Removing Visual Studio files (and adding them to `.gitignore`)
- Linux build instructions
    - Updating `premake5.lua` in all submodules and main repo
    - Added premake5.0.0-alpha14-linux binaries to vendor/premake5
- Added basic readme
    - Basic info about Hazel with links to social platforms
    - Install instructions for premake5
    - premake instructions to compile Hazel

PR #131, opened 2019-09-16
Linux port

ppetraki:
Following from https://github.com/TheCherno/Hazel/issues/130

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,16 @@
 # Directories
 .vs/
-/bin/
-bin-int/
+**/bin/
+**/bin-int
 
 # Files
 *.user
+Hazel/vendor/imgui/Makefile
+tags
+
+Hazel/Makefile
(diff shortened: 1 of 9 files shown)


LovelySanta:
instead of creating a seperate PR, could you contribute this to #82? However, I think this is all already implemented there.

ppetraki:
It looks pretty much the same sans the issues I didn't know about/run into yet (resources). What's stopping #82 from being merged into mainline? As long as it doesn't disturb the windows build it should be fine. Clearly, duplication of effort is a problem.

LovelySanta:
Currently it's just missing some filesystem solution, and cherno must mension it in his videos first. I'll close this (and #130) as duplicates of #82

ppetraki:
OK. Sounds like a plan. I'll take a closer look at #82. Thanks.

PR #132, opened 2019-09-16
Automated platform detection using predefined macro's

IshoAntar:
Since Hazel is supposed to be a multi-platform engine, i figured that this would be better than having to manually define the macro's for each platform, cluttering the premake file with needless stuff.


DaveAxiom:
Premake has lots of benefits and removing items may break premake features such as generating configuration files for unique IDEs. Merging this Pull Request may have consequences.

RsPlayer:
> Premake has lots of benefits and removing items may break premake features such as generating configuration files for unique IDEs. Merging this Pull Request may have consequences.

How does it break premake? this is for cross platform compiling. most people now use makefiles instead of IDE to compile static and shared libraries and premake was originally made to support generating makefiles

ReDucTor:
both approaches have their pros and cons
* in the header file it doesn't have a build dependency, if hazel is linked into another build system this is now reliable
* in premake it doesn't rely on the header file being included

I don't see how this could possibly break premake features, its a compiler definition not something side the premake files, it will mean that the define is not in the project file, but it still exists within the scope as the header file is included. There are some tools (e.g. very naive linters) that don't follow includes to bring in this stuff but I don't know if any legit IDE which doesn't.

DaveAxiom:
I see. For what is being removed from the Premake file, there likely isn't an IDE configuration that would be disabled. I plan to take a closer look at Premake soon. There's a Premake add-on script that exports to my favorite IDE.

CleverSource:
Looks amazing. I see no reason this couldn't be merged.

Hopefully it can get merged soon.

PR #134, opened 2019-09-20
Removing std function

ryanbbernstein:


Hazel/src/Hazel/Window.h
```diff
@@ -35,13 +33,25 @@ namespace Hazel {
...
+		virtual void SetEventCallback(void (*callback)(Event& e)) { m_Data.EventCallback = callback; };
 		virtual void SetVSync(bool enabled) = 0;
 		virtual bool IsVSync() const = 0;
 
 		virtual void* GetNativeWindow() const = 0;
 
 		static Window* Create(const WindowProps& props = WindowProps());
+
+	protected:
+		struct WindowData
```
ReDucTor: Why move this structure into Window? Seems like if thats useful it should be a separate PR

Hazel/src/Hazel/Window.h
```diff
@@ -35,13 +33,25 @@ namespace Hazel {
 		virtual unsigned int GetHeight() const = 0;
 
 		// Window attributes
-		virtual void SetEventCallback(const EventCallbackFn& callback) = 0;
+		virtual void SetEventCallback(void (*callback)(Event& e)) { m_Data.EventCallback = callback; };
```
ReDucTor: This change forces all events to be function pointers, which will not work well for anything that is not a singleton, I believe std::function is going to be the best here

Hazel/src/Hazel/Window.h
ryanbbernstein: I initially moved it since I was trying to use templates and those dont work with virtual functions. I can move it back.

Hazel/src/Hazel/Window.h
ryanbbernstein: Only for the window events though, and window events are handled by the application which is a singleton. I agree though that this isn't optimal.

ryanbbernstein: We could maybe look into a delegates library to add to the project. It would be useful on the future. I found this one which claims to be the fastest possible. https://github.com/dreamcat4/FastDelegate

Puddlestomper: Adding a library will require Cherno's approval, and for any sizable library he would probably rather use something he knows. I think adding a library for this is also maybe a bit overkill?

Puddlestomper:
Could you benchmark the performance difference? The issue was closed and it is currently not very clear why we should avoid `std::function` in this case.

Hazel/src/Hazel/Window.h
Puddlestomper: This should be in `WindowsWindow`, because different graphics APIs might want to store/handle this data in different ways and I think `Window` is meant to be a pure virtual interface so it shouldn't store any data.

Hazel/src/Hazel/Window.h
ryanbbernstein: Its a single header file

ryanbbernstein:
I did a benchmark with std::function, the fastdelegate library, function pointers, and with templates and these were the result:
`
Debug Build
Test function : std::cout << i << "\n";
Iterations: 10000
---------------------------------
std::function time:     7975.33ms
---------------------------------
fastdelegate time:     2357.67ms
---------------------------------
function ptr time:      2357.67ms
---------------------------------
template time:           2357.67ms
---------------------------------
`

`
Release Build
Test function : std::cout << i << "\n";
Iterations: 10000
---------------------------------
std::function time:     7969.86ms
---------------------------------
fastdelegate time:     2222.29ms
---------------------------------
function ptr time:      2222.29ms
---------------------------------
template time:           2222.29ms
---------------------------------
`

CleverSource:
That is a massive improvement. I think this change is viable.

ReDucTor:
That benchmark seems a little off, 
* The benchmark is doing io, this is going to blow out any benchmark and make it inconsistent
* As its I/O, and there is also no warm-up so you might find `std::function` is slower because you use it first, move it to second and try it
* Your test is combining both creation and calling, if you want to test the speed properly you must split these two things out otherwise you don't know where you pay for the cost
* Your testing based on a function pointer, there are many variants you should try (small objects, large objects, etc)

ReDucTor:
Just to show how this fails badly (tested on VS 2019 Preview)

Initial using your code
```
---------------------------------
std::function time:     5879.91ms
---------------------------------
fastdelegate time:      1803.82ms
---------------------------------
function ptr time:      1803.82ms
---------------------------------
template time:          1803.82ms
---------------------------------
```

Redirect output to a file `test.exe > output.txt`
```
---------------------------------
std::function time:	323.94ms
---------------------------------
fastdelegate time:	395.799ms
---------------------------------
function ptr time:	395.799ms
---------------------------------
template time:		395.799ms
---------------------------------
```

Move `std::function` to be the second one run

```
---------------------------------
std::function time:     1884.17ms
---------------------------------
fastdelegate time:      5480.94ms
---------------------------------
function ptr time:      5480.94ms
---------------------------------
template time:          5480.94ms
---------------------------------
```

Now lets spin that and do that again piped to a file

```
---------------------------------
std::function time:	371.055ms
---------------------------------
fastdelegate time:	313.417ms
---------------------------------
function ptr time:	313.417ms
---------------------------------
template time:		313.417ms
---------------------------------
```

So the performance tests aren't reliable, yes fast delegate will be slightly faster, but does not actually allow a callable which std::function does support, the difference between them is negligible. Keeping using it also means you don't have more third party code to support, have licenses for, etc

CleverSource:
Seeing the new tests makes me feel this is micro optimization. The reason my event PR was so substantial is due to the fact Events are used everywhere and it was a 30% optimization; this clearly isn't.

ilikenoodles2:
I would argue that there is actually a difference in performance.  Here is a test that more closely resembles the actual usage, which doesn't use io: 
[](https://gist.github.com/ilikenoodles2/07cf90d1efec598dc8aec1cdca87537e)

Feel free to suggest changes to this, if there is something I missed.  Here were the results I got with optimizations disabled.

Debug
`Average StdEventCallback time: 1520.06 ns
Average LamdaEventCallback time: 77.07 ns`

Release
`Average StdEventCallback time: 75.19 ns
Average LamdaEventCallback time: 4.78 ns`

Unfortunately, it doesn't test for templates, or the fastdelegate library, but it definitely shows a difference.  Also this doesn't benchmark actual performance differences in Hazel, so as @CleverSource says, it might be a micro optimization, that is until we actuallly benchmark real usage.

ReDucTor:
Your using `std::bind` you should be using a lambda for both of these if your wanting to make them equivalent

ReDucTor:
Even so, let's say our target is 300fps for your machine that gives you 3ms frame time 3,000,000 during this 3ms lets say your the fastest window resizer in the west and you manage to do 30 in that 3ms (10,000 per second), this even in debug mode would be 45601.8ns in your 3 million budget that's 1.5% compared to the 0.077% (remember in debug mode)

But you know what? That resizing of windows is only going to occur if glfw is polling for that, which it isn't its only going to be per-frame so really its 0.05%

And that's numbers based on debug mode with 300 fps, the difference in performance is insignificant. Your not doing 10,000 events per second to even make this 1.5% of a 300fps game in debug mode.

This is all back of napkin math, feel free to correct me if I'm wrong, remember nanoseconds are really small.

cheako:
If you are up for profiling, have a look at Hazel-rs...  I would be interested in seeing how it measures up to Hazel.  I'm in the middle of a "Can a person live without C and CPP?" hiatus, so am unable to test myself.

Hazel/src/Hazel/Window.h
LovelySanta: Since it is still here, I just want to clear it up: don't change things that are outside the scope of the PR.

As you mension, you did it to try something, failed, tried something else, it worked, congratz. The cleaning up part of a PR should include to remove unnecessary changes, so this should move back to WindowsWindow as Puddlestomper said.

Hazel/src/Hazel/Window.h
LovelySanta: I would leave std::function as well... and adding another library should be part of it's own PR?

ilikenoodles2:
I don't know how we missed this, but when you implement this PR, you replace `BIND_EVENT_FN` with the lambda.  Then you realize this macro is not only used for the window, it is also used to bind functions for event dispatching.  

std::bind takes in a function, and returns a function object anyways, which the Dispatch function then uses.  What effectively is happening, is for every call to Dispatch, a function object is created, and then called, which seems to be very inefficient. 

This then basically counters  #85 and #106.  And if it were up to me, I'd definitely want to fix this.

ReDucTor:
`std::bind` should never be used due to many of its short comings and no benefits. Also `std::bind` doesn't return a function object, it returns an unnamed callable object.

I'd love to see the following rules
 * Never use `std::bind` use a lambda, it has no use in modern C++, and has numerous short comings, and leads to bad idioms like `BIND_EVENT_FN` that have potentially dangerous ownership semantics
 * If its a callable that isn't stored use a template
 * If its a callable that is to be stored use `std::function`

Neomer:
Actual tests on g++-10 (release build with all optimisations) && Ubuntu 20.04.

I ran this algo 10 times and write all cout to the file.
avg std::function time:	0.8ms (min: 0.45ms max: 1.3ms)
avg function ptr time:	0.78ms (min: 0.44ms max: 1.23ms)
avg template time:		0.75ms (min: 0.44ms max: 1.23ms)

So it's really doesn't matter. This results shows that all implementations execution time almost the same.

PR #135, opened 2019-09-20
Fixed OpenGLShader::PreProcess Logic (issue #120)

### LINKED ISSUE #120: Contradiction In OpenGLShader::PreProcess
In `PreProcess` function, inside the while loop on the last `substr`. The second parameter checks to see if `nextLinePos == std::string::npos` and returns `source.size() - 1`. I'm not sure how this makes sense, furthermore we are passing nextLinePos as the first parameter; It seems if you pass `std::string::npos` as the first parameter of `substr` the program crashes, the second parameter checking for `nextLinePos == std::string::npos` doesn't make sense. but I've not hard anyone complaining about this so maybe I don't understand what is really going on here?

```c++
std::unordered_map<GLenum, std::string> OpenGLShader::PreProcess(const std::string& source)
{
    std::unordered_map<GLenum, std::string> shaderSources;

    const char* typeToken = "#type";
    size_t typeTokenLength = strlen(typeToken);
    size_t pos = source.find(typeToken, 0);
    while (pos != std::string::npos)
    {
        size_t eol = source.find_first_of("\r\n", pos);
        HZ_CORE_ASSERT(eol != std::string::npos, "Syntax error");
        size_t begin = pos + typeTokenLength + 1;
        std::string type = source.substr(begin, eol - begin);
        HZ_CORE_ASSERT(ShaderTypeFromString(type), "Invalid shader type specified");

        size_t nextLinePos = source.find_first_not_of("\r\n", eol);
        pos = source.find(typeToken, nextLinePos);
        shaderSources[ShaderTypeFromString(type)] = source.substr(nextLinePos, pos - (nextLinePos == std::string::npos ? source.size() - 1 : nextLinePos));
    }

    return shaderSources;
}
```

If you create a shader that specifies a tokenType like `#vertex` and the file ends on that line I think the program would crash. I would ASSERT this and throw a syntax error message for this as well.

Puddlestomper:
---
name: 'Pull Request: Implemented issue'
about: Fixes and explanatory comments for OpenGLShader::PreProcess
title: Fixed OpenGLShader::PreProcess logic (issue #120)
labels: ''
assignees: ''

---

Impact                   | Issue/PR
Relevant issue           | Resolves #120

We add two steps:
1. Assert that `nextLinePos` is not `std::string::npos`.
2. Split the final assign operation to let the substring run to the end of the file if `pos` is `std::string::npos` (if there is no next `#type` token).

We also add comments to clarify the purpose of each variable.

#### Additional context
This compiles and runs without error on the latest commit of Hazel.

### DIFF
--- a/Hazel/src/Platform/OpenGL/OpenGLShader.cpp
+++ b/Hazel/src/Platform/OpenGL/OpenGLShader.cpp
@@ -73,18 +73,18 @@ namespace Hazel {
 
 		const char* typeToken = "#type";
 		size_t typeTokenLength = strlen(typeToken);
-		size_t pos = source.find(typeToken, 0);												//Start of shader type declaration line
+		size_t pos = source.find(typeToken, 0);                                             //Start of shader type declaration line
 		while (pos != std::string::npos)
 		{
-			size_t eol = source.find_first_of("\r\n", pos);									//End of shader type declaration line
+			size_t eol = source.find_first_of("\r\n", pos);                                 //End of shader type declaration line
 			HZ_CORE_ASSERT(eol != std::string::npos, "Syntax error");
-			size_t begin = pos + typeTokenLength + 1;										//Start of shader type name (after "#type " keyword)
+			size_t begin = pos + typeTokenLength + 1;                                       //Start of shader type name (after "#type " keyword)
 			std::string type = source.substr(begin, eol - begin);
 			HZ_CORE_ASSERT(ShaderTypeFromString(type), "Invalid shader type specified");
 
-			size_t nextLinePos = source.find_first_not_of("\r\n", eol);						//Start of shader code after shader type declaration line
+			size_t nextLinePos = source.find_first_not_of("\r\n", eol);                     //Start of shader code after shader type declaration line
 			HZ_CORE_ASSERT(nextLinePos != std::string::npos, "Syntax error");
-			pos = source.find(typeToken, nextLinePos);										//Start of next shader type declaration line
+			pos = source.find(typeToken, nextLinePos);                                      //Start of next shader type declaration line
 
 			if (pos == std::string::npos) shaderSources[ShaderTypeFromString(type)] = source.substr(nextLinePos);
 			else shaderSources[ShaderTypeFromString(type)] = source.substr(nextLinePos, pos - nextLinePos);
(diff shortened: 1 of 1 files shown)


Hazel/src/Platform/OpenGL/OpenGLShader.cpp
```diff
@@ -73,18 +73,21 @@ namespace Hazel {
 
 		const char* typeToken = "#type";
 		size_t typeTokenLength = strlen(typeToken);
-		size_t pos = source.find(typeToken, 0);
+		size_t pos = source.find(typeToken, 0);                                             //Start of shader type declaration line
```
LovelySanta: I would suggest to just put the comments next to the line (single space), as due to the tabs, the indentations of the comments is wrong no matter what...

Hazel/src/Platform/OpenGL/OpenGLShader.cpp
```diff
@@ -73,18 +73,21 @@ namespace Hazel {
...
 
-			size_t nextLinePos = source.find_first_not_of("\r\n", eol);
-			pos = source.find(typeToken, nextLinePos);
-			shaderSources[ShaderTypeFromString(type)] = source.substr(nextLinePos, pos - (nextLinePos == std::string::npos ? source.size() - 1 : nextLinePos));
+			size_t nextLinePos = source.find_first_not_of("\r\n", eol);                     //Start of shader code after shader type declaration line
+			HZ_CORE_ASSERT(nextLinePos != std::string::npos, "Syntax error");
+			pos = source.find(typeToken, nextLinePos);                                      //Start of next shader type declaration line
+
+			if (pos == std::string::npos) shaderSources[ShaderTypeFromString(type)] = source.substr(nextLinePos);
+			else shaderSources[ShaderTypeFromString(type)] = source.substr(nextLinePos, pos - nextLinePos);
```
LovelySanta: Please, for readability, split these if/else up in 4 lines (you can still omit the `{}`), but seperate the statement and the actual line of code into two lines:
```cpp
			if (pos == std::string::npos)
				shaderSources[ShaderTypeFromString(type)] = source.substr(nextLinePos);
			else
				shaderSources[ShaderTypeFromString(type)] = source.substr(nextLinePos, pos - nextLinePos);
```
I don't see why you removed the ternary conditional operator? (maybe add some `()` if you feel like it)
```cpp
			shaderSources[ShaderTypeFromString(type)] = pos == std::string::npos ? source.substr(nextLinePos) : source.substr(nextLinePos, pos - nextLinePos);

Hazel/src/Platform/OpenGL/OpenGLShader.cpp
Puddlestomper: Done.

Hazel/src/Platform/OpenGL/OpenGLShader.cpp
Puddlestomper: Not sure why I did this either. Probably missed this solution. I have changed it to use the ternary conditional operator again (with brackets).

PR #137, opened 2019-09-23
Added check for OpenGL 4.5

### LINKED ISSUE #133: Unhandled  exception error
You get an error at all these call functions. 
glCreateVertexArrays(1, &m_RendererID);

Unhandled exception at 0x0000000000000000 in Sandbox.exe: 0xC0000005: Access violation executing location 0x0000000000000000. occurred

m_RendererID is not initialized but used

WhoseTheNerd:
This PR closes #133


Hazel/src/Platform/OpenGL/OpenGLContext.cpp
```diff
@@ -24,6 +24,15 @@ namespace Hazel {
 		HZ_CORE_INFO("  Renderer: {0}", glGetString(GL_RENDERER));
 		HZ_CORE_INFO("  Version: {0}", glGetString(GL_VERSION));
 
+		#if defined(HZ_ENABLE_ASSERTS)
```
Puddlestomper: We should probably standardize using `#ifdef X` vs `#if defined(X)`.

Hazel/src/Platform/OpenGL/OpenGLContext.cpp
```diff
@@ -24,6 +24,15 @@ namespace Hazel {
...
+		int versionMajor;
+		int versionMinor;
+		glGetIntegerv(GL_MAJOR_VERSION, &versionMajor);
+		glGetIntegerv(GL_MINOR_VERSION, &versionMinor);
+
+		HZ_CORE_ASSERT(versionMajor >= 4, "Hazel requires at least OpenGL version 4.5!");
+		HZ_CORE_ASSERT(versionMinor >= 5, "Hazel requites at least OpenGL version 4.5!");
```
Puddlestomper: Should probably make these one assert. Also spelling mistake.
```suggestion
		HZ_CORE_ASSERT(versionMajor >= 4 && versionMinor >= 5, "Hazel requires at least OpenGL version 4.5!");
```

Hazel/src/Platform/OpenGL/OpenGLContext.cpp
```diff
@@ -24,6 +24,15 @@ namespace Hazel {
...
 		HZ_CORE_INFO("  Version: {0}", glGetString(GL_VERSION));
 
+		#ifdef HZ_ENABLE_ASSERTS
+		int versionMajor;
+		int versionMinor;
+		glGetIntegerv(GL_MAJOR_VERSION, &versionMajor);
+		glGetIntegerv(GL_MINOR_VERSION, &versionMinor);
+
+		HZ_CORE_ASSERT(versionMajor >= 4, "Hazel requires at least OpenGL version 4.5!");
+		HZ_CORE_ASSERT(versionMinor >= 5, "Hazel requires at least OpenGL version 4.5!");
```
Puddlestomper: Once again, rather make these a single assert.

WhoseTheNerd: Oh. Forgot this.

Hazel/src/Platform/OpenGL/OpenGLContext.cpp
```diff
@@ -24,6 +24,14 @@ namespace Hazel {
 		HZ_CORE_INFO("  Renderer: {0}", glGetString(GL_RENDERER));
 		HZ_CORE_INFO("  Version: {0}", glGetString(GL_VERSION));
 
+		#ifdef HZ_ENABLE_ASSERTS
```
LovelySanta: I would remove the ifdef statement, the optimizer should take care of this as we never use the result?
If someone comes up with a reason to leave it, at least remove one indentation level on the #ifdef and #endif so they are same indented as the `{` and `}`

ReDucTor: The optimizer won't remove it as it can prove it's pure with no side affects, however it's an insignificant cost at startup so doing it even with asserts disabled should be fine.

WhoseTheNerd: The ifdef is there, because if asserts are disabled then the whole thing is useless.

LovelySanta: I know, but it only runs during initialisation/loading of the engine, I imagine this doesn't impact the code alot, like reductor mensioned, But yes, we can leave this indeed. In that case, at least remove one indentation level on the #ifdef and #endif so they are same indented as the `{` and `}`

```cpp
	void OpenGLContext::Init()
	{
		//...

	#ifdef HZ_ENABLE_ASSERTS
		int versionMajor, versionMinor;
		glGetIntegerv(GL_MAJOR_VERSION, &versionMajor);
		glGetIntegerv(GL_MINOR_VERSION, &versionMinor);
		HZ_CORE_ASSERT(versionMajor >= 4 && versionMinor >= 5, "Hazel requires at least OpenGL version 4.5!");
	#endif

**Comment by WhoseTheNerd** [comment]
> Anther note: as mensioned in [#133 (comment)](https://github.com/TheCherno/Hazel/issues/133#issuecomment-533666246), shouldn't we hint glfw as well:
> 
> ```c++
> glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 4);
> glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 5);
> glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);
> ```

That's another option to check for opengl 4.5

File: Hazel/src/Platform/OpenGL/OpenGLContext.cpp (hunk 4 above)
Comment (WhoseTheNerd) [reply Hazel/src/Platform/OpenGL/OpenGLContext.cpp]: Fixed the indentation. @LovelySanta

File: Hazel/src/Platform/OpenGL/OpenGLContext.cpp [hunk 5]
```diff
@@ -24,6 +24,14 @@ namespace Hazel {
...
+		int versionMajor;
+		int versionMinor;
+		glGetIntegerv(GL_MAJOR_VERSION, &versionMajor);
+		glGetIntegerv(GL_MINOR_VERSION, &versionMinor);
+
+		HZ_CORE_ASSERT(versionMajor >= 4 && versionMinor >= 5, "Hazel requires at least OpenGL version 4.5!");
```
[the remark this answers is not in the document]
Puddlestomper: Yes, I think it should. This won't pass version 5.0. Well spotted!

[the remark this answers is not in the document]
WhoseTheNerd: But will there be OpenGL 5.0 I don't think there will be.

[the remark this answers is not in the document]
Puddlestomper: There might be at some point? No sense in leaving it as a potential issue, even if it is very unlikely to happen.

LovelySanta:
Maybe related to #193 ? Try implementing the file logger (#196) and look in the log file to see if this is the actual error?

PR #138, opened 2019-09-23
Fixes for OrthographicCameraController Movement

Puddlestomper:
---
name: 'Pull Request: New issue'
about: Fixes the Camera Movement
title: Fixes for OrthographicCameraController Movement
labels: ''
assignees: ''

---

#### Describe the issue
The current way the camera moves is faulty when the camera is rotated. It still moves in the directions parallel to the world-space x- and y-axis rather than moving in the "left/right" (to the left/right of the window) or in the "up" direction (to the top of the window) of the camera. This results in it moving diagonally when the camera is rotated.

#### Expected behavior
Moving the camera should always move the view horizontally or vertically in the window and never diagonally (when only one key is held).

I used maths to make it move in the right direction. Not sure how much I need to explain but basically you rotate the base movement vectors by `m_CameraRotation` and then add those instead of just increasing the x and y values of position normally. Let me know if more detail/explanation is required.

I limited `m_CameraRotation` to be within the range (-180, 180] because I think (not sure where I read this) that the C++ trig functions becomes less accurate at large degrees.

I also added a comment to clarify that `m_CameraRotation` is in degrees.


ryanbbernstein:
Thanks for implementing the fix for the issue I mentioned on discord. I was gonna mention that my original math using just tan on y axis and 1 on x axis was wrong as it would translate more than intended but seems like you figured that out.

Masterninja13579:
I know this is closed but I have found another way to fix the problem that doesn't involve extra code.

In the OrthographicCamera class there is a function that is used to recalculate the view matrix.  It's there that the camera does its matrix multiplication to get the new transform:
TempCapture1

Here we are simply multiplying the translation matrix by the rotation matrix.  The fix is simply to switch the multiplication order:
TempCaptrue2

This looks like a better solution than using the trig functions and is probably what was intended in the first place

CleverSource:
I can confirm after testing that @Masterninja13579's solution provides the exact same solution as this pull request provides. It would probably be more efficient as well.

I would recommend someone (preferably @Masterninja13579 himself) makes a new pull request on this.

PR #139, opened 2019-09-27
Creates the unique_ptr by using make_unique()

Simonmicro:
See: https://stackoverflow.com/questions/22571202/differences-between-stdmake-unique-and-stdunique-ptr-with-new
...if this project uses c++17

### DIFF
--- a/Hazel/src/Hazel/Application.cpp
+++ b/Hazel/src/Hazel/Application.cpp
@@ -20,7 +20,7 @@ namespace Hazel {
 		HZ_CORE_ASSERT(!s_Instance, "Application already exists!");
 		s_Instance = this;
 
-		m_Window = std::unique_ptr<Window>(Window::Create());
+		m_Window = std::make_unique<Window>(Window::Create());
 		m_Window->SetEventCallback(BIND_EVENT_FN(OnEvent));
 
 		Renderer::Init();


WhoseTheNerd:
Sorry to say this to you, but there is another PR that does exactly like that on larger scale. Replacing all `new unique_ptr` with `std::make_unique`

Simonmicro:
Hmm, i cant find an other request doing this (maybe it's inside the hidden dev branch?) - so I've created one. But when will the PR referenced by you merged? Otherwise I wouldn't created this PR...

Puddlestomper:
I think he is referring to #129. I checked now and this change is not included there.

The reason is probably that this doesn't compile. You get:
`error C2259: 'Hazel::Window': cannot instantiate abstract class` and
`error C2664: 'Hazel::Window::Window(const Hazel::Window &)': cannot convert argument 1 from '_Ty' to 'const Hazel::Window &'`

So I think the solution is to leave this for a PR refactoring the various `Create` methods (as what #112 did) and then have the `Create` method for the `Window` return a `Scope` directly. Until then we need to use the `std::unique_ptr` constructor.

And since we do plan on using/use C++17 this change has no real effect anyway (as is said in your stackoverflow link).

LovelySanta:
As @Puddlestomper said, this is the same as we did in #112. I'll mark this as duplicate. I won't close it just yet.

I think we should make an issue that contains all the things that have to happen in an upcomming code maintenance that we can reference towards in the future.

LovelySanta:
See issue #140 that I mensioned in this https://github.com/TheCherno/Hazel/pull/139#issuecomment-535997732.

PR #142, opened 2019-10-01
Implemented the Repeat Count for KeyPressed Events

Puddlestomper:
#### Describe the issue
Currently we are simply making the repeat count 0 if it's the first press and 1 otherwise.

#### Expected behavior
You would expect the repeat count to increment for each time the same key's `KeyPressedEvent` is created.

We want to store the repeat counts of the keys somehow. We need the method of storage to satisfy the following:

1. Can store key counts for multiple keys.
2. Is relatively fast seeing as it needs to be accessed every time a `KeyPressedEvent` is created.
3. Doesn't use memory for a key that is held once and then left alone for the rest of the application's runtime. Basically it should be able to erase the key's entry relatively fast when the key is released (optional).

I have implemented this with a static `unordered_map` in `WindowsWindow.cpp` to track the repeat count of each key. I am open to other suggestions.

I also added a comment (in `KeyEvent.h`) to clarify that the repeat count is how many times the event has been repeated, not how many times it has fired in total. So on the third event it has repeated twice, not three times.

#### Additional context
This compiles and works on the latest Hazel commit (734ea18).

I also want to note some (maybe) odd behaviour when holding a key (say W) and pressing and releasing another key (say S) while still holding the first key all the while. After doing this no more `KeyPressedEvent`s will be released even though W is still being held. This seems to be due to how GLFW works and I guess how they define a `GLFW_REPEAT` (Key Repeat) event. I am not sure if we want to do something about this.

### DIFF
--- a/Hazel/src/Hazel/Events/KeyEvent.h
+++ b/Hazel/src/Hazel/Events/KeyEvent.h
@@ -23,7 +23,7 @@ namespace Hazel {
 		KeyPressedEvent(int keycode, int repeatCount)
 			: KeyEvent(keycode), m_RepeatCount(repeatCount) {}
 
-		inline int GetRepeatCount() const { return m_RepeatCount; }
+		inline int GetRepeatCount() const { return m_RepeatCount; } //The Repeat Count is number of times the key event has repeated. So on the third time the event triggers it has repeated twice.
 
 		std::string ToString() const override
 		{
--- a/Hazel/src/Platform/Windows/WindowsWindow.cpp
+++ b/Hazel/src/Platform/Windows/WindowsWindow.cpp
@@ -90,7 +90,7 @@ namespace Hazel {
 				}
 				case GLFW_RELEASE:
 				{
-					s_KeyRepeatCounts[key] = 0;
+					s_KeyRepeatCounts.erase(key);
 					KeyReleasedEvent event(key);
 					data.EventCallback(event);
 					break;
--- a/Hazel/src/Platform/Windows/WindowsWindow.cpp
+++ b/Hazel/src/Platform/Windows/WindowsWindow.cpp
@@ -10,6 +10,7 @@
 namespace Hazel {
 	
 	static bool s_GLFWInitialized = false;
+	static std::unordered_map<int16_t, unsigned int> s_KeyRepeatCounts;
 
 	static void GLFWErrorCallback(int error, const char* description)
 	{
@@ -77,24 +78,27 @@ namespace Hazel {
 		glfwSetKeyCallback(m_Window, [](GLFWwindow* window, int key, int scancode, int action, int mods)
 		{
 			WindowData& data = *(WindowData*)glfwGetWindowUserPointer(window);
-
+			
 			switch (action)
 			{
 				case GLFW_PRESS:
 				{
+					s_KeyRepeatCounts[key] = 0;
 					KeyPressedEvent event(key, 0);
 					data.EventCallback(event);
 					break;
 				}
 				case GLFW_RELEASE:
 				{
+					s_KeyRepeatCounts[key] = 0;
 					KeyReleasedEvent event(key);
 					data.EventCallback(event);
 					break;
 				}
 				case GLFW_REPEAT:
 				{
-					KeyPressedEvent event(key, 1);
+					s_KeyRepeatCounts[key]++;
+					KeyPressedEvent event(key, s_KeyRepeatCounts[key]);
 					data.EventCallback(event);
 					break;
 				}
@@ -149,6 +153,8 @@ namespace Hazel {
 
 	void WindowsWindow::Shutdown()
 	{
+		s_KeyRepeatCounts.clear();
+		
 		glfwDestroyWindow(m_Window);
 	}
 


Hazel/src/Hazel/Events/KeyEvent.h
```diff
@@ -23,7 +23,7 @@ namespace Hazel {
 		KeyPressedEvent(int keycode, int repeatCount)
 			: KeyEvent(keycode), m_RepeatCount(repeatCount) {}
 
-		inline int GetRepeatCount() const { return m_RepeatCount; }
+		inline int GetRepeatCount() const { return m_RepeatCount; } //The Repeat Count is number of times the key event has repeated. So on the third time the event triggers it has repeated twice.
```
tgsm: This comment should probably be placed above the line.

Puddlestomper: Done

Puddlestomper:
@LovelySanta mentioned on the discord that the memory object (the `unordered_map` at this point in time) should be in the `WindowsInput` class so that it can be polled or if not there, then at least part of the `WindowData`.

I agree that it would nice to put it in `WindowsInput`, however I am not sure on how to implement it. I guess it might be a member of `WindowsInput` with a getter? But that would require access to the instance. Any suggestions are welcome. I would rather not put it in the `WindowData`, because it isn't really related to the window, but rather just the input.

CleverSource:
I can understand having it in ``WindowsInput`` as it still relates to input, but I can't understand why it would go in ``WindowData``?

ReDucTor:
This seems unusual, aren't doesn't the event get sent for each repeat notification? It seems to me like a book instead of an int, so you can ignore repeats when your tracking things like WSAD not dealing with text input.

I'd even go as far to say it should be a completely separate event type to reduce mistakes on assuming it should be used for handling things like WSAD

Puddlestomper:
> This seems unusual, aren't doesn't the event get sent for each repeat notification? It seems to me like a book instead of an int, so you can ignore repeats when your tracking things like WSAD not dealing with text input.

> I'd even go as far to say it should be a completely separate event type to reduce mistakes on assuming it should be used for handling things like WSAD

Movement should be handled by polling whether the key is pressed in an `OnUpdate` function (which does indeed return a bool) which has a time step parameter anyway, not by events. I am not quite sure what the application for the `RepeatCount` would be, but @TheCherno clearly wanted it as a feature.

It should also be noted that, as far as I can tell, there is no set frequency that the event is triggered at while a key is held so there is no way to see how long the key has been held from the `RepeatCount`.

> I can understand having it in WindowsInput as it still relates to input, but I can't understand why it would go in WindowData?

I think the `WindowData` thing was so that it is stored at the GLFW user pointer and not just a static variable somewhere. I was mostly just paraphrasing LovelySanta, but I can somewhat see it being viable since GLFW ties everything to the window anyway. Even our currently implemented input polling has to first get a pointer to the window to call the GLFW methods.

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -10,6 +10,7 @@
 namespace Hazel {
 	
 	static bool s_GLFWInitialized = false;
+	static std::unordered_map<int16_t, unsigned int> s_KeyRepeatCounts;
```
LovelySanta: I really would like to see this as a static member of the input class...

###### My thoughts:
- Make this a private member
- Create a getter (to get the repeat count of a key)
- Create a setter (to set it to 0)
- Create a reset (to delete it upon release)
- Create an incrementer (to increase the repeat count, this could also act as a 2nd getter?)

###### Thinking about other platforms, code duplication
The increment/setting/resetting (aka use of the created functions) could be coded in the constructors of the `KeyPressedEvent` and the `KeyReleasedEvent` instead of the callback *(of each platform)*:
- `KeyPressedEvent` takes in a bool instead of an integer for distinguishing press/repeat
  - if not repeat then set to 0
  - if repeat then increment
- `KeyReleasedEvent` resets it to non-existing

Any other thoughts?

Puddlestomper: Moving this to the `Input` class to allow polling is a good idea.

I'm not 100% sure what the use would be for the setter? My interpretation of what you suggest would be that you only have the ability to increment it or reset it (in terms of how you are able to modify it). Resetting it would remove it from the map, so there is no need to have a setter or to ever store "0 repeats" for a key. Unless you want to be able to keep the memory for the key and not have to potentially reallocate each time?

Should we store this directly in the Input class or delegate it to each Platform's input class? Some platforms might have this sort of counter already built in and keeping another one would be wasting some memory (although it's really not a lot of memory) in that case.

I don't think it would be that much code duplication to delegate it to each platform, but it is a valid point to consider still.

I'll push all the changes once we clear up the details.

LovelySanta: > I'm not 100% sure what the use would be for the setter? My interpretation of what you suggest would be that you only have the ability to increment it or reset it (in terms of how you are able to modify it). Resetting it would remove it from the map, so there is no need to have a setter or to ever store "0 repeats" for a key. Unless you want to be able to keep the memory for the key and not have to potentially reallocate each time?

My initial idea was a getter and a setter, and when I thought about the increment, the setter made sense for 0... if it is not existing, the key press count = 0, "incrementing" then would be the ++ operator...

But yeah, I think the getter function returning 0 when it doesn't exist would have the same effect, so no real reason for the setter, as you pointed out.

> Should we store this directly in the Input class or delegate it to each Platform's input class? Some platforms might have this sort of counter already built in and keeping another one would be wasting some memory (although it's really not a lot of memory) in that case.
>
> I don't think it would be that much code duplication to delegate it to each platform, but it is a valid point to consider still.

Hazel will use only 1 set of keycodes (and converter functions to go from one platform to the other), so I don't see why we can't store it in the input class? As the user will also use the hazel keycodes (as do the events), I don't see a specific reason to put it in the platform specific code?

> I'll push all the changes once we clear up the details.

Feel free to ask any questions you have 👍

Puddlestomper:
I have changed the implementation now as @LovelySanta asked.

It still has the same issues I mentioned before (holding multiple keys means the repeat event is only fired for the last pressed key) and having an entire file (`Input.cpp`) for a single definition is maybe not ideal.

I think it is worth considering having on a thread that checks if a key is pressed in a while(running) loop that has a specific frequency or something so that the repeat count has meaning in terms of how long a key has been held down. I don't really see a use for it otherwise. This would, of course, have its own downsides and it might be better to just leave it as is.

mandar1jn:
Please fix this

Puddlestomper:
I can fix this if there's a real chance of it getting merged. I haven't been keeping up with the series, so if this won't get merged anyway I'd rather close it if you want to clean out the PR list. However, I am happy to catch up with the series and fix this properly if you want this merged.

TheCherno:
Closed (3701bbb)

PR #143, opened 2019-10-03
Fix the C26495 warning.

8infy:
Fixed the warning as per Cherno's request. There are more similar warnings in the project but they're related to spdlog, specifically the 'fmt' library.

### DIFF
--- a/Hazel/src/Hazel/Renderer/Buffer.h
+++ b/Hazel/src/Hazel/Renderer/Buffer.h
@@ -36,7 +36,10 @@ namespace Hazel {
 		uint32_t Offset;
 		bool Normalized;
 
-		BufferElement() {}
+		BufferElement()
+            		: Name(""), Type(), Size(0), Offset(0), Normalized(false)
+		{
+		}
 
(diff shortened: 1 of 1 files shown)


Jack-Punter:
Whilst this solution does work, you have essentially just explicitly made this default constructor as the compiler would have done if you had used BufferElement() = default;
The PR that I have made (to address another issue) does contain this fix. This was not done to undermine you, I had just also fixed this warning in my local repo.

8infy:
> Whilst this solution does work, you have essentially just explicitly made this default constructor as the compiler would have done if you had used BufferElement() = default;
> The PR that I have made (to address another issue) does contain this fix. This was not done to undermine you, I had just also fixed this warning in my local repo.

Yeah i saw your pr. I'm not sure if it's better to mark it as default or be more verbose and write out the default values. I think it's up to Cherno to decide here.

CleverSource:
It's better to mark as default because it is what Cherno does everywhere else. Simply put; follow what he has already done.

8infy:
> It's better to mark as default because it is what Cherno does everywhere else. Simply put; follow what he has already done.

Yeah I guess you are right.

PR #144, opened 2019-10-03
Fix Build warnings to do with BufferElement Offset Type

Jack-Punter:
This has been previously fixed, however the fix was to add a cast between the value and the (const void*) cast. This is a bad solution as it does not actually solve the problem behind the warning. just suppresses it.
The correct solution is to make the Offset member the correct type so that it is guaranteed to be the correct size to hold a pointer.
Also the Default constructor was not initializing any of the members so I made = default to get expected behavior.

### DIFF
--- a/Hazel/src/Hazel/Renderer/Buffer.h
+++ b/Hazel/src/Hazel/Renderer/Buffer.h
@@ -33,10 +33,10 @@ namespace Hazel {
 		std::string Name;
 		ShaderDataType Type;
 		uint32_t Size;
-		uint32_t Offset;
+		intptr_t Offset;
 		bool Normalized;
 
-		BufferElement() {}
+		BufferElement() = default;
 
 		BufferElement(ShaderDataType type, const std::string& name, bool normalized = false)
 			: Name(name), Type(type), Size(ShaderDataTypeSize(type)), Offset(0), Normalized(normalized)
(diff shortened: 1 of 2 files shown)


Jack-Punter:
PR#143 also addresses the constructor issue in this, however, its solution is more verbose and has the same outcome (i believe)

Jack-Punter:
Please ignore that force push, I forgot to make a new branch for this PR so when I tried to fix another issue it showed up here.

Jack-Punter:
In this case (as AFAIK we are only supporting 64bit builds) would it be better to make this uint64_t (as we can almost guarantee that all pointer types are going to be that size)

WhoseTheNerd:
`size_t` is same as `uint32_t`. So after your commit(11de26bc84b8a166e2e1cdc3fd339aaca478f2cf) this PR now is pointless.

Jack-Punter:
This is not true, size_t is 64bit.
see: https://godbolt.org/z/DI_Mko

LovelySanta:
size_t will always resolve to the biggest possible size, depending on the architecture, for a 32 bit system, this is a uint32_t, for a 64 bit system this is a uint64_t and in the future maybe even bigger... size_t is defined as the biggest size possible.

Jack-Punter:
Correct, but (currently) hazel only supports 64x so it doesn't really matter.
For most architectures, this is exactly what we want as this is the same size as pointers on those architectures. 
I know of older architectures where this is not the case but I think this will be a safe enough assumption in this case.
This can always be revisited if it causes issues if and when we move to other platforms

Issue #148, opened 2019-10-07
Promotion logo for the engine


SimonMaracine:
Hey! Maybe it would be nice for the Hazel engine to have a  **_Made with Hazel Engine_** or **_Hazel Powered_** logo, or something like that, for people to optionally include it in their games as a promotion for the engine. That is only if these assets (the main Hazel logo, this "promotion logo" and probably other things) will be included alongside the binaries in the distribution.

Also, different resolutions for the same logo wouldn't be bad, because if one would make, for example, a small resolution, 2D, pixelated game, he/she wouldn't need a large 2000x2000 logo.

SeppahBaws:
Sounds like a nice idea to me, however I think that would be something for the future as the engine in its current state is not ready to have games developed with it just yet.

LovelySanta:
> Sounds like a nice idea to me, however I think that would be something for the future as the engine in its current state is not ready to have games developed with it just yet.

I am pro/con about this, we got the assets loader in already, so we can load basic images. That said, it wouldn't be a bad idea to make the API to set the icon on the window. This would result in the icon in the top corner of the window itself, and how it shows on the taskbar. Those **could** introduce some lower res images, also with only the `H`, without the Hazel underneath.

So yes, it would already be possible to have some sort of lower res images, but indeed, it is quite early, especialy for having promotion logo's, that will come with time, probably when we'll have the loading cycle of some sorts, providing a default loading screen.

AnantTiwari-Naman:
I think you should have an option for disabling that for *free*. Let people make their choice and not have to buy a "Pro" version to remove that.

SimonMaracine:
> I think you should have an option for disabling that for _free_. Let people make their choice and not have to buy a "Pro" version to remove that.

I agree. But I said "to optionally include it in their games".

PR #151, opened 2019-10-09
Add SetIcon for Window

WhoseTheNerd:
This is feature proposed by @LovelySanta at #148

### DIFF
--- a/Hazel/src/Platform/Windows/WindowsWindow.cpp
+++ b/Hazel/src/Platform/Windows/WindowsWindow.cpp
@@ -175,7 +175,6 @@ namespace Hazel {
 	{
 		return m_Data.VSync;
 	}
-
 	void WindowsWindow::SetIcon(const std::string& path)
 	{
 		int width, height, channels;
@@ -188,5 +187,6 @@ namespace Hazel {
 		images[0].height = height;
 		images[0].pixels = data;
 		glfwSetWindowIcon(m_Window, 1, images);
+		stbi_image_free(data);
 	}
 }
--- a/Hazel/src/Platform/Windows/WindowsWindow.cpp
+++ b/Hazel/src/Platform/Windows/WindowsWindow.cpp
@@ -7,6 +7,8 @@
 
 #include "Platform/OpenGL/OpenGLContext.h"
 
+#include "stb_image.h"
+
 namespace Hazel {
 	
 	static bool s_GLFWInitialized = false;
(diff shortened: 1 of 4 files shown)


Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -174,4 +176,18 @@ namespace Hazel {
 		return m_Data.VSync;
 	}
 
+	void WindowsWindow::SetIcon(const std::string& path)
+	{
+		int width, height, channels;
+		stbi_set_flip_vertically_on_load(1);
```
L4ZZA: I'm quite sure this code is the same as the one used for texture loading. 
A good approach would be to extract a helper function (i.e. _load_image_) with parameters to flip on load and use that both here and in the texture class.

WhoseTheNerd: Yes, but I will need to create new utils.h file where I put code like this. Right now it will be copy pasted.

ilikenoodles2:
This is a nice feature to have.  What are your thoughts on making the Hazel logo default?
Hazel_Logo_Window_Icon

LovelySanta:
> This is a nice feature to have. What are your thoughts on making the Hazel logo default?
> *Hazel logo*

Since cherno did design the logo with Hazel underneath, it is made for Hazel.
Removing the word Hazel to extract the logo + using it to contribute back to Hazel would be acceptable I think?

Questions remaining:
- Do we put a max res logo in branding or not?
- Do we put fixed sized images (256x256, see [reference](https://stackoverflow.com/questions/3236115/which-icon-sizes-should-my-windows-applications-icon-include)) in the assets folder?
- This shouldn't be part of this PR, but rather a follow-up PR?

CleverSource:
I know it shouldn't be part of the PR but looking at the system behind this, it would be easy to have the Hazel logo be default, I like the idea as eventually Cherno would want to do it either way.

Hazel/src/Platform/Windows/WindowsWindow.cpp
L4ZZA: That's exactly what I would do move it to a helper file/class, a few more clicks for better maintainability.

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -174,4 +176,18 @@ namespace Hazel {
...
+		stbi_uc* data = stbi_load(path.c_str(), &width, &height, &channels, 4);
+		HZ_ASSERT(channels == 4, "Icon must be RGBA!");
```
EngineerSmith: Shouldn't this be HZ_CORE_ASSET

WhoseTheNerd:
Not maintained anymore.

PR #153, opened 2019-10-10
Added a Macro to automate excluding the Entrypoint

AhmadMamdouhEnan:
While watching the latest Game Engine Series video I noticed that the cherno just deleted the 'Entrypoint.h' include from 'Hazel.h' and Included it himself in 'Sandbox.cpp', which I thought doesn't looked weird, so I thought of adding this macro, which will make it clearer that this is the main file.
though I think this sounds like a small detail.

### DIFF
--- a/Hazel/src/Hazel.h
+++ b/Hazel/src/Hazel.h
@@ -28,5 +28,7 @@
 // -----------------------------------
 
 // ---Entry Point---------------------
-#include "Hazel/Core/EntryPoint.h"
+#ifdef __MAIN_FILE__
+  #include "Hazel/Core/EntryPoint.h"
+#endif
 // -----------------------------------
--- a/Sandbox/src/SandboxApp.cpp
+++ b/Sandbox/src/SandboxApp.cpp
@@ -1,3 +1,6 @@
+// The __MAIN_FILE__ Macro should be only defined in the file which handles
+// the Application creation (in other words, the file which contains the 'CreateApplication' function)
+#define __MAIN_FILE__
 #include <Hazel.h>
 
 #include "Platform/OpenGL/OpenGLShader.h"


Puddlestomper:
You shouldn't use names with double underscores or names starting with an underscore followed by an uppercase letter. As seen in [this stackoverflow post](https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier), these names are reserved for the compiler/linker/standard libraries to use.

I'm also not sure this change is necessary. The user only needs to include the Entrypoint header once in their project, and I think it is better to let them explicitly (and we can add this requirement to future documentation) do that rather than force them to include the entire Hazel header with a define. Putting this in the Hazel header just makes more work for the preprocessor to do since it now has to evaluate all those ifdefs in each source file that includes the Hazel header.

Both ways (yours and the current way) require one line of code, so I don't see how this is really an improvement at all.

WhoseTheNerd:
@Puddlestomper Just by counting lines of code implemented for a feature is wrong way to measure it. **THERE IS NO CORRECT WAY TO MEASURE IT**

AhmadMamdouhEnan:
@Puddlestomper Just read the stackoverflow post you attached, ok if it doesn't help at all I think I will just close the pull request, though I thought it could be kinda cooler, bec. personally I wouldn't like 
to include the entrypoint this way if I am a person who uses this engine to make a game.
also it won't increase the compile time that much, if u add a preprocessor in the project settings in visual studio it will just add a #define at the top of EACH file in the project.

Puddlestomper:
> @Puddlestomper Just by counting lines of code implemented for a feature is wrong way to measure it. **THERE IS NO CORRECT WAY TO MEASURE IT**

I merely meant it in the sense of how much effort it requires the user of the engine to define the entrypoint of his/her application. I didn't mean to measure the worth of the PR by just a line count. I see that my comment was vague on that front though, so I apologize if it sounded more criticizing that I intended.

My compile time argument isn't a huge issue, but I felt it might be worth mentioning anyway. The real impact of this will probably not be noticeable at all.

I am still open to discussing this if @AhmadMamdouhEnan wants to. We can always just change the define's name to just be something like `HZ_MAIN_FILE` instead of `__MAIN_FILE__`. I don't see any benefit to this, but that doesn't necessarily mean there isn't one.

PR #154, opened 2019-10-11
Compile premake from source

### LINKED ISSUE #90: Update Premake to compile alpha 14 frome source
The current Premake in this repo is the version 5.0.0-alpha13 that doesnt support VS2019 solutions, please update this to alpha 14. Btw in my fork of Hazel i updated this and everything works fine on VS2019 solutions.

LovelySanta:
#### Describe the issue
Github isn't made as a storage system, so storing binaries is not the goal. Hence why the reason to add premake as a submodule (like other vendor submodules) and compile premake *ONCE*, only when it is not compiled yet.

#### Expected behavior
When generating the VS project files, it first checks if premake is already build. If this is the case, it behaves as before, if premake isn't build, it will build premake and then continueu with the creation of the project files.

- Replaced the premake5 alpha14 binary with the premake submodule poining to the premake5 alpha14 branch
- Created script `Win-BuildPremake.bat` to compile premake
- Updated `Win-GenProjects.bat` to check if premake is build or not

Other issues this solves | Resolves #90


cheako:
I don't think the issue is as big as it's being made out to be.  The batch file that bootstraps premake is essentially the same as having a premake binary.  The batch is basically a blob of unintelligible mess... It would be different if this batch was invoking a script in the premake project, but it's much more involved.

I would infer that the issue of scripting building premake is more complicated than this commit could address.  Otherwise premake would include a script that does what the batch in this commit does.  Dosn't that signal to anyone else big flashing error signs?

lhurt51:
I don't believe this PR is entirely necessary because the premake5.exe file is only 1,331kb, which is very small! Plus it is very easy to go to the premake site and download the newest version and paste it into the premake directory. I believe that this is adding bloat to the building process and when the project becomes bigger this will cause significant slowdowns in rebuilding the project files. I like your thought process, but not everything needs to be automated/abstracted.

WhoseTheNerd:
@lhurt51 This allows us to have most up-to-date premake. Maybe microsoft rolls visual studio 2021 but we don't have to update the binary. 
>  but not everything needs to be automated/abstracted.
Then try to build Hazel project on command line. Good luck.
Automation is important, because it will make our job easier.

lhurt51:
@WhoseTheNerd this is why we use premake so we don't have to build using the command line. In the event that we wanted to build using the command line are you really going to check for an updated premake every time you compile? Most likely you will run into the visual studio project creation errors and need to update your premake file through the command-line anyway. Plus you wouldn't be compiling from command line if you were using visual studio and premake because premake handles project file generation and visual studio handles compilation. I believe it is a lot easier for newer coders to copy-paste one file rather than checking for an updated premake everytime they recreate their visual studio project files. Not only that but let's say that premake releases a version that breaks something within visual studio 2017. Then it would automatically pull those updates and you wouldn't know until you tried creating visual studio 2017 project files. As I said automation is great but you must be very careful how you want to use it. Maybe if the bat files were not linked that would even be a better option because then users can decide if they need to update premake.

LovelySanta:
After some discussions in Discord and the comments here, I came to the conclusion that having it build on different systems (assuming windows) could give rise to some errors, but it could be solved in most cases.

However, when we'll be moving on to other builds (for example linux), this doesn't allow you to compile from source as you require a windows build with visual studio, or a previous version of premake. At that point we could include the correct version at once...

Conclusion of this all: In order to build Hazel, it is faster, more robust to just include that 1400 kB of binaries for each system, rather than dealing with the complications/impossibilities of building premake from source. This is just a tool for us to use in order to develop Hazel.

With this in mind, i'll close this.

Issue #140, opened 2019-09-27
Code Maintenance Task List


Puddlestomper:
As suggested @LovelySanta, this issue is meant for us to make a full list of everything that needs to be done in the next code maintenance episode/PR/session.

**Please note that no PRs need to be made directly into the master branch for these issues, but rather that any PRs should be made to the [maintenance branch](https://github.com/LovelySanta/Hazel/tree/Code-Maintenance).**

## Pull Request 1: Updating code base (#165) [Merged]

- [x] Make sure all `Create` methods return a `Ref`/`Scope` as appropriate and fix all uses of these methods accordingly (originally in #112).
      *This was mentioned in the episode about Refs and Scopes.*
  - [x] `GraphicsContext`
  - [x] `VertexBuffer`
  - [x] `IndexBuffer`
  - [x] `VertexArray`
  - [x] `Shader`
  - [x] `Texture`
  - [x] `RendererAPI`
  - [x] ~~`Window`~~
  - [x] ~~`Input`~~
  - [x] ~~`Application`~~
- [x] Change other smart pointers where possible
  - [x] `Log` class
- [x] Remove all remaining dll traces like the `HZ_BUILD_DLL` and `HAZEL_API` defines.
      *This was mentioned when we converted Hazel to a static library.*
- [x] Fix include statements
  - [x] All includes should be relative to `Hazel/src` and not relative to where the file is located.
      *This was mentioned in the episode of the CameraController.*
  - [x] Include `Core.h` in `Hazel.h`
  - [x] External includes should use `<>` instead of `""`
- [x] Remove the `BIND_EVENT_FN` in Application and replace it with the `HZ_BIND_EVENT_FN` instead
- [x] Use default constructor instead of empty constructor
  - [x] `LayerStack`
  - [x] `ApplicationEvents`
- [x] Clean up code copied into `Renderer2D.cpp`
  - [x] Remove `Hazel::` namespace on line 72
  - [x] `squareVertices` array size is too big (`5 * 4` instead of `3 * 4`)
  - [x] Errors in Renderer2D due to create functions returning a ref
- [x] Remove obsolete glwf logging messages init + terminate
    *Requested by Cherno in Discord*
- [x] Remove obsolete YAML keys in PR template
- [x] Add `Shutdown` method in `Renderer` and call it. Also call the `Renderer2D` `Shutdown` method in it, since it is currently not being called and not freeing the data in `Renderer2DStorage`.
[...]

PR #158, opened 2019-10-16
Old code maintenance

AbdullahAmrSobh:
### Old code maintenance and clean up
 1. Replaced all the ints and floats with their corresponding fixed-length int_t and float_t.
 2. Replaced all std::unique_ptr with Hazel::Scope and std::shared_ptr with Hazel::Ref.

### Replaced raw pointers with smart pointers in the following files
1. Buffer.h
> VertexBuffer::Create(); //@return Ref<VertexBuffer>
> IndexBuffer::Create(); //@return Ref<IndexBuffer>

2. Application.h
> Application::Create //@return Scope<Application>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ bin-int/
 **.vcxproj
 **.vcxproj.filters
 **.vcxproj.user
+Sandbox/imgui.ini
--- a/Hazel/src/Hazel/Core/Application.cpp
+++ b/Hazel/src/Hazel/Core/Application.cpp
@@ -20,7 +20,7 @@ namespace Hazel {
 		HZ_CORE_ASSERT(!s_Instance, "Application already exists!");
 		s_Instance = this;
 
-		m_Window = std::unique_ptr<Window>(Window::Create());
+		m_Window = Scope<Window>(Window::Create());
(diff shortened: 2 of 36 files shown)


LovelySanta:
Please read #140 about how to contribute to the code maintenance.

After looking in detail:
- Adding refs/scopes
  - Window::Create is on the todo list
  - CreateApplication has been added to the todo list
  - Log is already on the todo list (make a PR for that to the code maintenance branch)
  - RendererAPI has been implemented in LovelySanta/Hazel#33
  - Buffers have been implemented in LovelySanta/Hazel#28 and LovelySanta/Hazel#29
  - VertexArray has been implemented in LovelySanta/Hazel#30
  - Shader has been implemented in LovelySanta/Hazel#31
  - Texture2D has been implemented in LovelySanta/Hazel#32
- Using fixed length types
We only use fixed length when we explicitly state the memory of the data, if it is just to hold some 'small' data, for example the window dimensions, we don't really care about the size, especialy when passing it to a function, as that will just get pushed onto the stack anyway (so all same size)..

I'll close this as duplicate. Feel free to say if I'm wrong/incomplete or mistaken.

PR #159, opened 2019-10-21
Remove all redundant 'inline' keyword usage.

8infy:
#### Describe the issue
A lot of useless ``inline`` keyword usage, which has no meaning in the context. 

> A function defined entirely inside a class/struct/union definition, whether it's a member function or a non-member friend function, is implicitly an inline function.

(from [en.cppreference.com](https://en.cppreference.com/w/cpp/language/inline))

#### PR impact
Engine code will be cleaner.

#### Proposed fix
Remove all the unnecessary 'inline' keyword usage throughout the codebase.


Jack-Punter:
I tend to agree with these updates. The biggest thing I noticed looking through this is the lack of consistency, which IMO is more important. There are a couple that don't have inline (the ones returning iterators for example) and many that do.

njchensl:
I totally agree with you. Every time I copy down Cherno's code for inline member function definitions, my Resharper tells me that inline is redundant. We should probably remove all redundant ones and keep the style consistent.

PR #161, opened 2019-10-21
Set m_CameraTranslationSpeed when m_ZoomLevel Changes

MrBrN197:
---
name: 'Pull Request'
about: Propose a fix for a (small) issue in Hazel

---

A clear and concise description of what the issue is. Explain the difference between the expected and the current behavior.
A screenshot or copy of the error could be helpful as well.

#### Additional context

### DIFF
--- a/Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
+++ b/Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
@@ -52,7 +52,6 @@ namespace Hazel {
 
 		m_Camera.SetPosition(m_CameraPosition);
 
-		m_CameraTranslationSpeed = m_ZoomLevel;
 	}
 
 	void OrthographicCameraController::OnEvent(Event& e)
@@ -67,6 +66,9 @@ namespace Hazel {
 		m_ZoomLevel -= e.GetYOffset() * 0.25f;
 		m_ZoomLevel = std::max(m_ZoomLevel, 0.25f);
 		m_Camera.SetProjection(-m_AspectRatio * m_ZoomLevel, m_AspectRatio * m_ZoomLevel, -m_ZoomLevel, m_ZoomLevel);
+		
+		m_CameraTranslationSpeed = m_ZoomLevel;
+		
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


Puddlestomper:
Now that you are highlighting this. Changing the translation speed should probably be part of the `SetZoomLevel` method and then we should use that method in the `OnMouseScrolled` method. This would ensure that the translation speed is correct no matter what changes the zoom level, since currently calling the `SetZoomLevel` method does not change the translation speed immediately (even though it will be updated in the next call of the `OnUpdate` function, which you are removing).

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
```diff
@@ -52,7 +52,6 @@ namespace Hazel {
 
 		m_Camera.SetPosition(m_CameraPosition);
```
LovelySanta: Remove this empty line
```suggestion
```

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
```diff
@@ -67,6 +66,9 @@ namespace Hazel {
 		m_ZoomLevel -= e.GetYOffset() * 0.25f;
 		m_ZoomLevel = std::max(m_ZoomLevel, 0.25f);
 		m_Camera.SetProjection(-m_AspectRatio * m_ZoomLevel, m_AspectRatio * m_ZoomLevel, -m_ZoomLevel, m_ZoomLevel);
+		
+		m_CameraTranslationSpeed = m_ZoomLevel;
```
Jack-Punter: Good update although if you do it this way, the value for m_CameraTranslationSpeed should be calculated in the constructor rather than set to a default value. I made this optimization in my code when following the video and noticed that, if the camera is moved (with WASD) before a scroll event occurs, it moves much faster than if you zoom in then back to the initial zoom level.

Jack-Punter: By calculated, I realised that I do just mean initializing it to m_ZoomLevel in the Ctor initializer list.

MrBrN197: Yeah, I did notice that it was a problem even before; we set it to 5 in initialization then immediately overwrote it. We should probably set it a multiple of zoom level and some speed that way its independent of zoom level but proportional to it.

PR #165, opened 2019-10-27
Code maintenance

LovelySanta:
Code maintenance of things Cherno did start but not finish or mensioned before but did not do in the code maintenance video.

Issues this solves       | Part 1 of #140

###### *A more detailed list can be found in https://github.com/TheCherno/Hazel/issues/140#issue-499535992*
- Convert the remaining smart pointers to Hazel wrappers
  - `std::shared_ptr` -> `Hazel::Ref` and `std::make_shared` -> `Hazel::CreateRef`
  - `std::unique_ptr` -> `Hazel::Scope` and `std::make_unique` -> `Hazel::CreateScope`
- Removing all DLL traces
  *Removed all `HAZEL_API` macros*
- Fixing include statements
  - All includes relative to `Hazel/src`
  - All external includes use `<>` instead of `""`
- Use default (de)constructor instead of an empty one
- Removed the dedicated (duplicate) `BIND_EVENT_FN` in application
  *Now uses the generic `HZ_BIND_EVENT_FN` instead*
- Removed obsolete logging messages in glfw
- Removed obsolete YAML keys in PR template

#### Additional context
Tested in debug and release, the rest of the code maintenace can be found in https://github.com/TheCherno/Hazel/issues/140#issue-499535992.


Hazel/src/Hazel/Core/Core.h
```diff
@@ -43,22 +43,6 @@
 	#error "Unknown platform!"
 #endif // End of platform detection
 
-
-// DLL support
-#ifdef HZ_PLATFORM_WINDOWS
```
WhoseTheNerd: Does Cherno really want go static-lib with his game engine?

Hazel/src/Hazel/Core/EntryPoint.h
```diff
@@ -1,8 +1,9 @@
 #pragma once
+#include "Hazel/Core/Core.h"
 
 #ifdef HZ_PLATFORM_WINDOWS
 
-extern Hazel::Application* Hazel::CreateApplication();
+extern Hazel::Scope<Hazel::Application> Hazel::CreateApplication();
```
WhoseTheNerd: I remember that Cherno wanted full control over when application gets closed.

Sandbox/src/SandboxApp.cpp
```diff
@@ -218,7 +215,7 @@ class Sandbox : public Hazel::Application
 	}
 };
 
-Hazel::Application* Hazel::CreateApplication()
+Hazel::Scope<Hazel::Application> Hazel::CreateApplication()
 {
-	return new Sandbox();
+	return Hazel::CreateScope<Sandbox>();
```
WhoseTheNerd: Same thing.

Hazel/src/Hazel/Core/Core.h
LovelySanta: Yes, since he converted to the static lib, he sometimes removes a HAZEL_API when he comes across it, and in the video about converting to a static lib he mentioned this should be removed in a maintenance.

Hazel/src/Hazel/Core/EntryPoint.h
LovelySanta: Ok, we can revert this one back to a raw pointer.

Hazel/src/Hazel.h
```diff
@@ -2,6 +2,8 @@
 
 // For use by Hazel applications
 
+#include "Hazel/Core/Core.h"
```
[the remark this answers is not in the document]
LovelySanta: Yes, it is indeed called `Base.h`. However, since we use it alot, shouldn't that be mensioned in a video instead? Tbh, I didn't even think about this one. Maybe that one should have a dedicated PR?

Sandbox/src/SandboxApp.cpp
LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/165/commits/47e7ebf8f1fe6d56aaa93456fc099b014e94f821

Hazel/src/Hazel/Core/EntryPoint.h
LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/165/commits/47e7ebf8f1fe6d56aaa93456fc099b014e94f821

Puddlestomper: I think making it a smart pointer is actually fine. If we look at how `EntryPoint` works we see that the entire lifetime of the application is spent in the `Application::Run` method which means we can never call `delete app` in the application because we are freeing memory that is still being used, which would then cause a crash.

The mechanism for closing the application is to set `Application::m_Running` to false since that ends the while loop in the run method and then the next line of code that gets executed in `EntryPoint` is to delete the Application. This is also what we currently do in `Application::OnWindowClose`. The change to a smart pointer still allows this method of closing the application and simply removes the need for that `delete` call in `EntryPoint`.

Closing the application through this mechanism also ensures that we don't close mid-render or mid-some-other-process and allow all processes to finish normally before exiting.

Also no part of the application has access to the pointer returned by this starting method anyway, since it only exists in `EntryPoint`. You might also note that we do keep `Application::s_Instance` a raw pointer, but this is because `Application` cannot have ownership of itself, which is what a `Scope` would imply.

Hazel/src/Hazel/Core/EntryPoint.h
```diff
@@ -13,7 +14,7 @@ int main(int argc, char** argv)
 
 	auto app = Hazel::CreateApplication();
 	app->Run();
-	delete app;
+    delete app;
```
LovelySanta: Convert spaces to tabs
```suggestion
	delete app;
```

CleverSource:
When will this be merged?

LovelySanta:
> When will this be merged?

Still debating if I should rename Core.h to Base.h or not... Otherwise I think tomorrow, or after cherno's next vid/upload to github

Hazel/src/Hazel.h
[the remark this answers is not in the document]
LovelySanta: I just took a look at Hazel-dev to see what plans Cherno could have with Core.h and I've decided that I'll leave the maintenance to him, as it is more than "just renaming the file". So I'll mark this as resolved as I won't implement this.

LovelySanta:
Should LovelySanta/Hazel#47 be merged into this?

CleverSource:
I don't think it should be merged into this.

LovelySanta:
> I don't think it should be merged into this.

The reason I did it lije this, only Window.cpp includes the WindowsWindow, so in that file we can add/remove the obsolete classes when they are not required. This could even work for the different renderAPI's, for example exclude DirectX in macOS...

Any particular reason why you're against it @CleverSource ? I can see how it might require a seperate PR instead...

CleverSource:
It's a good PR, but it doesn't feel very maintenance like and seeing as this PR is meant for maintenance I don't see why a PR unrelated should be merged with this.

LovelySanta:
Removed LovelySanta#47 from the TODO list.

CleverSource:
Will this be merged now?

LovelySanta:
> Will this be merged now?

There are still some things to do, see #140, such as removing the OpenGL casting in the 3D renderer...

LovelySanta:
> Will this be merged now?

Now everything is implemented. I want to have anther look in the weekend and I plan on merging it on monday if everything looks alright.

Hazel/src/Hazel/Renderer/GraphicsContext.cpp
```diff
@@ -0,0 +1,21 @@
...
+	{
+		switch (Renderer::GetAPI())
+		{
+			case RendererAPI::API::None:    HZ_CORE_ASSERT(false, "RendererAPI::None is currently not supported!"); return nullptr;
+			case RendererAPI::API::OpenGL:  return CreateScope<OpenGLContext>(static_cast<GLFWwindow*>(window));
+		}
+
+		HZ_CORE_ASSERT(false, "Unknown RendererAPI!");
+		return nullptr;
+	}
```
Puddlestomper: Should @TheCherno cover/mention this in a video? Since it's introducing a new file to the project.

Hazel/src/Hazel/Renderer/RendererAPI.cpp
```diff
@@ -1,8 +1,22 @@
...
+	{
+		switch (s_API)
+		{
+			case RendererAPI::API::None:    HZ_CORE_ASSERT(false, "RendererAPI::None is currently not supported!"); return nullptr;
+			case RendererAPI::API::OpenGL:  return CreateScope<OpenGLRendererAPI>();
+		}
+
+		HZ_CORE_ASSERT(false, "Unknown RendererAPI!");
+		return nullptr;
+	}
```
Puddlestomper: We should maybe consider having this create method take in the API as a parameter and then set `RendererAPI::s_API` along with the other code. It might, however, also be better to do this somewhere else and I think it falls outside the scope of this PR.

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -41,7 +41,6 @@ namespace Hazel {
 
 		if (s_GLFWWindowCount == 0)
 		{
-			HZ_CORE_INFO("Initializing GLFW");
```
Puddlestomper: We could keep this in. Otherwise we should probably also remove the extra logging in `EntryPoint` (the "Hello! Var = ..." thing).

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -151,10 +149,10 @@ namespace Hazel {
 	void WindowsWindow::Shutdown()
 	{
 		glfwDestroyWindow(m_Window);
+		--s_GLFWWindowCount;
 
-		if (--s_GLFWWindowCount == 0)
+		if (s_GLFWWindowCount == 0)
 		{
-			HZ_CORE_INFO("Terminating GLFW");
```
Puddlestomper: Same as with the init message.

Sandbox/src/SandboxApp.cpp
```diff
@@ -1,9 +1,10 @@
 #include <Hazel.h>
+#include <Hazel/Core/Core.h>
```
Puddlestomper: `Core.h` is now part of `Hazel.h` so we don't need to include it again.

Hazel/src/Platform/Windows/WindowsWindow.cpp
CleverSource: Cherno actually specifically said he wanted these logs gone, they aren't randomly being removed.

Sandbox/src/SandboxApp.cpp
CleverSource: Cherno actually specifically said he wanted these logs gone, they aren't randomly being removed.

Hazel/src/Platform/Windows/WindowsWindow.cpp
CleverSource: Cherno actually specifically said he wanted these logs gone, they aren't randomly being removed.

Sandbox/src/SandboxApp.cpp
CleverSource: We should probably look for other spots in the codebase where this was done.

Hazel/src/Hazel/Renderer/GraphicsContext.cpp
CleverSource: I agree, is this change necessary? I feel like Cherno should do it himself but it's so small that it could be fine.

Hazel/src/Platform/Windows/WindowsWindow.cpp
LovelySanta: image

The reasoning for this: this is pure something the engine needs to do, the user doesn't have any knowledge of GLFW or it's existence. So having a message telling it is initializing is just cluttering the logging. This was good for testing the PR about the shutdown and the windowcount. But not something to leave in the code.

However, we do have an assert to notify when it goes wrong, to give feedback to Hazel about why the application is not launching.

I'll mark this as resolved.

Sandbox/src/SandboxApp.cpp
LovelySanta: Yes, sandbox should indeed only include `Hazel.h` and the `entrypoint.h`. The only exception is for example to include some OpenGL file when we don't have it interfaced in Hazel yet.

LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/165/commits/f1676f4378d5c57018bc5c598a8e7437200e6ba4, also removed the OpenGL specific includes

Hazel/src/Platform/Windows/WindowsWindow.cpp
LovelySanta: See same as the init message https://github.com/TheCherno/Hazel/pull/165#discussion_r344480639. Marking as resolved

Hazel/src/Hazel/Renderer/RendererAPI.cpp
LovelySanta: The way I was thinking about it was have 1 function that actualy sets `RendererAPI::s_API`. This would usualy be read in from some config settings that are saved on the user side, if not, use a default (probably OpenGL).

I bet cherno will have a solution by that time. Whatever happens, it is not part of this PR as you said. Marking it as resolved.

Hazel/src/Hazel/Renderer/GraphicsContext.cpp
LovelySanta: The only reason this is here is to not have to include OpenGL in the Window class. It is yet another create function that Cherno already rushes over in his videos as it is a million times the same code over and over. Switch over the API's and create the correct one..

As cherno skips over it in his videos, does it really need to be in the videos?

LovelySanta:
> One other thing that I can think of that we might want to include, seeing as we have the platform detection now, is the include guards for the API-specific files depending on platform. The `Windows*` files would only be on Windows and the `OpenGL*` files can probably work on both Windows and Linux. Do you think we should do that? And, if so, should it be part of this PR?

That is exactly what I asked some days ago: https://github.com/TheCherno/Hazel/pull/165#issuecomment-547148507.

The only place Windows* files are included are within the `Window` and `Input` class for the `WindowsWindow` and `WindowsInput` respectively. By adding a create function to them (as done in LovelySanta/Hazel#47), we can use the include guards to include the headers (if needed) and use them to return the correct implementation.

I agree with the reply from CleverSource https://github.com/TheCherno/Hazel/pull/165#issuecomment-548185614 on the matter: It is a good PR, but not really maintenance, it should have its own PR into Hazel instead.

So I think this is indeed a nice suggestion, but also outside the scope of this PR?

CleverSource:
It most likely is outside the scope of this PR to be quite honest.

beewyka819:
I know this is an old PR that was merged over a year ago, but Cherno mentioned in the Smart pointer episode (20 minutes in) that the spdlog::loggers should remain in std::shared_ptrs instead of Hazel Refs since spdlog::loggers aren't Hazel assets?

LovelySanta:
I do remember that yes, but as long as it is just a wrapper around shared ptrs, this is equivalent. When we implement our own ref system, it has to change indeed.

PR #168, opened 2019-11-05
Added file reading sanity check

Paul1365972:
---
name: 'Added file reading sanity check'
about: Missing fail check in OpenGLShader::ReadFile

---

[std::ifstream::tellg](http://www.cplusplus.com/reference/istream/istream/tellg/) can return -1 on fail, which leads to a fatal memory allocation, as -1 is treated as an unsigned value.
Should print error and return instead.

Added check on the critical tellg(), so std::string::resize will not be called with -1 anymore. 
Treats this the same way as if the file was missing in the first place.

#### Additional context
Compiles and runs without error on all currently relevant machines (Windows x64).


LovelySanta:
LGTM 👍

PR #169, opened 2019-11-08
Missing Iterator Wrappers in LayerStack.h

Jack-Punter:
about: The way Application.cpp was sending events to each of the layers was using forward iterators to reverse iterate through the layer stack.
---

#### Describe the issue
The way Application.cpp was sending events to each of the layers was using forward iterators to reverse iterate through the layer stack. This is a little unclear and could be more explicitly a reverse loop using reverse iterators.

#### Proposed fix
Added more wrapper functions to the LayerStack Class to access both forwards, reverse and const iterators from the vector of layers.

#### Additional context
Validated that code works on windows. This code is adding wrappers to STL containers which are platform independent.


csp256:
Personally, I prefer using a simple iterator wrapper that allows me to use range-for loops in reverse order. (I actually got this as an interview question once!) Example implementation:

https://stackoverflow.com/a/28139075

This allows you to write

    for (auto& layer : reverse(m_LayerStack)) {
        layer->OnEvent(e);
        // ...
    }

Which I find so much cleaner and more expressive than `for (auto it = m_LayerStack.rbegin(); it != m_LayerStack.rend(); ++it)` or `(*it)->OnEvent(e);`. You can use Godbolt to verify there is no runtime overhead, too.

Jack-Punter:
@csp256 Interesting, I have never seen this adaptor mechanism used before to reverse a ranged for loop. Either way, the main addition to this PR was the iterator wrappers, so if we were to create this adaptor in a util's directory or something it should just work as you posted. But I don't think I want to go so far as to add it for this current use case.

If it comes up more often later It may be worth considering.

LovelySanta:
Marked as bug for not implementing the const iterator, adding suggestion for the reverse iterators

ReDucTor:
I'm not a huge fan of those iterator wrappers as they can hide unnecessary copying or dangling references depending on the implementation

csp256:
@ReDucTor Avoiding unnecessary copies is easy to get right once, use forever.

"Dangling references" doesn't make sense. Are you talking about iterators getting invalidated out from underneath you? None of the STL containers offer stable iterators to my knowledge.

ReDucTor:
Even the code in the stack overflow post has potential use-after-free
https://cppinsights.io/s/4468c2f1

ReDucTor:
What do you expect to happen to `std::vector<int>` here, copied, a reference kept to it?
```
std::vector<int> test();
for(auto && a : reverse(test()) cout << a;
```

> Avoiding unnecessary copies is easy to get right once, use forever.

I think you'll find this is not the case, avoiding unnecessary copies is not easy, make the above code safe then make it avoid an unnecessary copy from

```
const std::vector<int> & test2();
for(auto && a : reverse(test2()) cout << a;
```

csp256:
Yes, lifetime extention only applies to the final expression in a range-for. That's not at all unique to iterator wrappers. If you want that type of safety you're better off going ahead and paying for proper stable iterators.

ReDucTor:
Having them in a code base is extremely error prone, people misuse them, especially in a project like this where there are a lot of new people to C++.

I'm not saying that these wrappers are unique, they exist however I would suggest avoiding them for situations like this, they just introduce hidden bugs and catch out developers.

Jack-Punter:
I'm not sure that I'm following this feed. Is your problem with the iterators themselves or just the wrappers used in the PR... because I can't see how your points are unique to the wrappers added by this.

ReDucTor:
Iterators are fine, they are a concept which C++ developers should become familiar with, however wrappers like @csp256 suggested should be avoided imho

PR #170, opened 2019-11-10
Move 3D Layer to Separate Files 

cxgslegend:
3D Layer

---
name: 'Move 3D Layer to Separate Files '
about: Propose moving the Sandbox 3D layer implementation example code to separate files, so that the code is consistent between the 2D and 3D layer examples. 

---

#### Describe the issue
The 3D layer example and 2D layer example are not represented the same way to the user. The 2D layer example is presented with its own .h and .cpp files, but the 3D layer example is not. 

#### PR impact 

#### Proposed fix 
This is only a simple file move. It will only help with code consistency. 

#### Additional context
* I tested the code after I moved the code to ensure it produced the 3D example in the same way as before. 
* I also tested the 2D example to ensure there were no side effects. 

All tests were performed on Windows with Visual Studio Code 2019


CleverSource:
There is no 3D currently supported..?

That layer you're moving is just an example of what Hazel can do, nothing more, Cherno wants it to stay how it is according to his videos meaning this should not be merged unless Cherno wants it to happen.

cxgslegend:
I understand there is no 3D supported. But the code he started was intended to be 3D eventually. At least that was my understanding. 

I assumed that is why he didn't remove it when he started the 2D layer.

CleverSource:
The code he started was an example, he said the example layer would be added on upon when 3D comes so that there is an example of how 3D works. 

If anything you should at least name it appropriately to be named ExampleLayer.

cxgslegend:
I would, of course, be willing to rename it. I do think being named Sandbox3D is better for when he does add 3D to it. But it's his call. 

At the very least, I think it helps clean up the SandboxApp.cpp file in the meantime.

LovelySanta:
If all goes well, I'll merge a maintenance PR, which alters a lot in those files already. Afther that, you can move the `ExampleLayer` to it's own file, but don't call it `Sandbox3D`, just call it `ExampleLayer`. But this might indeed be a good idea.

cxgslegend:
Thanks LovelySanta! I'll do another commit when you merge your maintenance PR.

LovelySanta:
> Thanks LovelySanta! I'll do another commit when you merge your maintenance PR.

#165 has been merged ;)

cxgslegend:
Thanks @LovelySanta! The PR has been updated for upstream changes.

Sandbox/src/ExampleLayer.cpp
```diff
@@ -0,0 +1,198 @@
...
+	ImGui::Begin("Settings");
+	ImGui::ColorEdit3("Square Color", glm::value_ptr(m_SquareColor));
+	ImGui::End();
+}
+
+void ExampleLayer::OnEvent(Hazel::Event& e) 
+{
+	m_CameraController.OnEvent(e);
+}
+
```
CleverSource: You should remove these extra spaces.

cxgslegend: Thanks, for pointing that out @CleverSource. The extra spaces have been removed.

CleverSource:
Look's ready to be merged! @LovelySanta hopefully you can merge this soon. 👍

Sandbox/src/SandboxApp.cpp
```diff
@@ -1,210 +1,16 @@
...
+#include "ExampleLayer.h"
 
 class Sandbox : public Hazel::Application
 {
 public:
 	Sandbox()
 	{
-		// PushLayer(new ExampleLayer());
-		PushLayer(new Sandbox2D());
+		PushLayer(new ExampleLayer());
```
CleverSource: One last thing, you should have the ExampleLayer commented out since when Cherno pulls next time he needs to be on ``Sandbox2D`` since that is what the series is currently on, it will make it easier for Cherno so he doesn't get confused on a video. Didn't notice this until now. 😛 

You should just uncomment Sandbox2D and have ExampleLayer commented out.

Sandbox/src/ExampleLayer.cpp
LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/170/commits/ddeb98cc54ade9ca965f144823cf28f496283475

Sandbox/src/SandboxApp.cpp
LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/170/commits/be8290d630af3aef401ae3a6b1fa46088398675d

LovelySanta: https://github.com/TheCherno/Hazel/pull/170/commits/be8290d630af3aef401ae3a6b1fa46088398675d

cxgslegend: I am so sorry about that. When I updated the PR with upstream changes, I tested both layers again, but I forgot to return to the previous layer state after testing. Good catch!

PR #174, opened 2019-11-21
created __FUNCSIG__ definition for unix compilers

### LINKED ISSUE #173: __FUNCSIG__ is not defined on all target platforms.
On both GCC and Clang, `__FUNCSIG__` is not defined. Inside Instrumentor.h, there should be a macro defining `__FUNCSIG__` as `__PRETTY_FUNCTION__` which I've found to be exclusive to *Unix targeted compilers (GCC, Clang, ARM GCC, Arduino's compilers etc.).

since `__FUNCSIG__` isn't defined on all the platforms that `__PRETTY_FUNCTION__` is, the following would fix the issue on all Unix targeted compilers

```cpp
#ifdef __FUNCSIG__
#define __FUNCSIG__ __PRETTY_FUNCTION__
#endif
```

proof it compiles on all target platforms compilers:
https://godbolt.org/z/wtudtn

GabeRundlett:
Pull request which closes #173

Issues this solves       | Resolves #173

Fixes cross-platform compatibility inside of `Instrumentor.h`

This PR was tested on all platforms as detailed in the issue linked above (#173)

EDIT by @LovelySanta: Added [closing keywords](https://help.github.com/en/articles/closing-issues-using-keywords)


CleverSource:
Make sure to add a closing keyword so that it actually closes issue #173, just put "closes #173" somewhere. 👍

Hazel/src/Hazel/Debug/Instrumentor.h
```diff
@@ -1,129 +1,116 @@
...
-#include <chrono>
 #include <algorithm>
+#include <chrono>
 #include <fstream>
+#include <string>
 
 #include <thread>
 
 namespace Hazel {
-	struct ProfileResult
```
LovelySanta: In this whole file, please convert this back to tabs instead of spaces.

Hazel/src/Hazel/Debug/Instrumentor.h
```diff
@@ -1,129 +1,116 @@
...
+      private:
+        const char *m_Name;
+        std::chrono::time_point<std::chrono::high_resolution_clock> m_StartTimepoint;
+        bool m_Stopped;
+    };
+} // namespace Hazel
 
 #define HZ_PROFILE 1
 #if HZ_PROFILE
+	#ifndef __FUNCSIG__
```
LovelySanta: What about creating a macro `HZ_FUNC_SIG` (in core.h) as follows:

```cpp
#if defined(__FUNCSIG__)
	#define HZ_FUNC_SIG __FUNCSIG__
#elif defined (__PRETTY_FUNCTION__)
	#define HZ_FUNC_SIG __PRETTY_FUNCTION__
#else
	#error "No function signature function defined!"
#endif
```

_The macro name can be changed, the message here is about how the macro is defined (and where)._

ReDucTor: This might not work outside of a function as those macros are designed for usage inside a function, so besure to fully test it if going with this

GabeRundlett: This is a nice and formalized method to solve this issue, however, it may be overkill to put it in core.h considering Cherno probably will only use HZ_FUNC_SIG inside of the instrumentor implementation. I like the idea of including a #error though and will update the code (as well as fix the whitespace)

Hazel/src/Hazel/Debug/Instrumentor.h
```diff
@@ -1,129 +1,116 @@
...
-
-			m_Stopped = true;
-		}
-	private:
-		const char* m_Name;
-		std::chrono::time_point<std::chrono::high_resolution_clock> m_StartTimepoint;
-		bool m_Stopped;
-	};
-}
+    struct ProfileResult {
```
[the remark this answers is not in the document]
GabeRundlett: You're commenting on the first commit of the file. I already pushed a version that has reverted the formatting. For me when I go to `Files Changed` I see the comparison between the newest commit of my branch. Is there more I must do to set the newest commit as the one to be contributed?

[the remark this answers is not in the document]
Puddlestomper: Don't worry about it. The latest commit on the branch is always used in the pull request. You can see at the top of this comment box that it says "Outdated" in yellow to mark that the comment is on an older commit.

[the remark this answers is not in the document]
DrOptix: I know. I've seen the updated changes and I approved the code review that I started. But at first I did not expect PR comment to have a pending period. :)

Hazel/src/Hazel/Debug/Instrumentor.h
LovelySanta: > Let me know what you think I should change

@GabeRundlett, now you define a HZ_FUNC_SIG when profile is set to 1, I would also add it in the else statement to be empty (like the other ones)? `	#define HZ_FUNC_SIG`
https://github.com/TheCherno/Hazel/blob/3aff8a00f063efb429d5e0e766820b9df9243207/Hazel/src/Hazel/Debug/Instrumentor.h#L127-L133

> This might not work outside of a function as those macros are designed for usage inside a function, so besure to fully test it if going with this

@ReDucTor, the place where this is used is literaly called `HZ_PROFILE_FUNCTION()`. I think it is up to the user to correctly utilize it? We have an alternative scope profiler as well in that case:
https://github.com/TheCherno/Hazel/blob/3aff8a00f063efb429d5e0e766820b9df9243207/Hazel/src/Hazel/Debug/Instrumentor.h#L136-L137

Hazel/src/Hazel/Debug/Instrumentor.h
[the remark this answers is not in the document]
LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/174/commits/3aff8a00f063efb429d5e0e766820b9df9243207

Hazel/src/Hazel/Debug/Instrumentor.h
LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/174/commits/3aff8a00f063efb429d5e0e766820b9df9243207

Hazel/src/Hazel/Debug/Instrumentor.h
ReDucTor: It's not about where it's used but where you check those macros, which is outside a function.

LovelySanta: ohhhh, now I understand

Hazel/src/Hazel/Debug/Instrumentor.h
```diff
@@ -124,10 +124,17 @@ namespace Hazel {
 
 #define HZ_PROFILE 1
 #if HZ_PROFILE
+	#if defined(__FUNCSIG__)
+		#define HZ_FUNC_SIG __FUNCSIG__
+	#elif defined(__PRETTY_FUNCTION__)
+		#define HZ_FUNC_SIG __PRETTY_FUNCTION__
+	#else
+		#define HZ_FUNC_SIG __FUNCTION__
+	#endif
```
Puddlestomper: Should we first check to see if `__func__` is available and use it before using `__FUNCTION__`? See [this stackoverflow answer/post](https://stackoverflow.com/a/4384825) for the reasoning.

LovelySanta: According to [the language references](https://docs.microsoft.com/en-us/cpp/cpp/func?view=vs-2019), cpp says __func__ is mandated since C++11 (been mandated in C99 as well).

With that note, do we need any macro definitions then? Should we just use __func__ instead as this should be cross platform, as we are compiling with C++17 anyway?

Puddlestomper: I believe the string returned by `__func__` is less detailed than `__FUNCSIG__` and `__PRETTY_FUNCTION__` ([on gcc at least](https://godbolt.org/z/edtdmm)) so we should use those when available. We can remove the case for `__FUNCTION__` though, since (as you said) we are compiling with C++17.

LovelySanta: I guess now we have to do some research about how to find out what compiler we're using and such, what functions are available...

LovelySanta: After alot of research I came up with this solution, mainly based on [Boost's assert function](https://www.boost.org/doc/libs/1_66_0/libs/assert/doc/html/assert.html#current_function_macro_boost_current_function_hpp):

```cpp
#if defined(__GNUC__) || (defined(__MWERKS__) && (__MWERKS__ >= 0x3000)) || (defined(__ICC) && (__ICC >= 600)) || defined(__ghs__)
	# define HZ_FUNC_SIG __PRETTY_FUNCTION__
#elif defined(__DMC__) && (__DMC__ >= 0x810)
	# define HZ_FUNC_SIG __PRETTY_FUNCTION__
#elif defined(__FUNCSIG__)
	# define HZ_FUNC_SIG __FUNCSIG__
#elif (defined(__INTEL_COMPILER) && (__INTEL_COMPILER >= 600)) || (defined(__IBMCPP__) && (__IBMCPP__ >= 500))
	# define HZ_FUNC_SIG __FUNCTION__
#elif defined(__BORLANDC__) && (__BORLANDC__ >= 0x550)
	# define HZ_FUNC_SIG __FUNC__
#elif defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901)
	# define HZ_FUNC_SIG __func__
#elif defined(__cplusplus) && (__cplusplus >= 201103)
	# define HZ_FUNC_SIG __func__
#else
	# define HZ_FUNC_SIG "HZ_FUNC_SIG unknown!"
#endif
```

What do you think @Puddlestomper ?

Puddlestomper: Looks fine to me. I don't know all (or even most) of these macros, but it seems to work for gcc and clang on Godbolt and I confirmed it works for Visual Studio and gcc/g++ in test projects.

If there are any issues with this then we can always change it. For now it will at least work for the main compilers so I think it's good. Thanks for doing the research to figure this out.

LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/174/commits/2b06109be382a435724c782d9f69ef8e228ec672

Hazel/src/Hazel/Debug/Instrumentor.h
LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/174/commits/2b06109be382a435724c782d9f69ef8e228ec672

PR #175, opened 2019-11-22
Fixed Instrumentor generating the same start time more than once.

Emilundpixeln:
The InstrumentationTimer weren't synchronized and multiple of the same start time were generated. This lead to overlapping as seen in https://youtu.be/qiD39bB7DvA?t=1535. 

Now when starting an InstrumentationTimer it checks and potentially adjusts the starting time a bit, so that no two starting times are identical. This fixes the overlapping:
trtr

#### Additional context
~~I assumed that the Instrumentor will be used from multiple threads. If this is not the case, the mutex can be removed.~~ The mutex is not needed anyway. The thread_local keyword works as well.
 I only tested this on Windows, but there is no platform specific code used.

### DIFF
--- a/Hazel/src/Hazel/Debug/Instrumentor.cpp
+++ b/Hazel/src/Hazel/Debug/Instrumentor.cpp
@@ -1,0 +1,3 @@
+#include "hzpch.h"
+#include "Instrumentor.h"
+thread_local long long Hazel::Instrumentor::s_LastTimeStart = 0;
--- a/Hazel/src/Hazel/Debug/Instrumentor.h
+++ b/Hazel/src/Hazel/Debug/Instrumentor.h
@@ -137,13 +137,13 @@ namespace Hazel {
 
 #define HZ_PROFILE 1
 #if HZ_PROFILE
-#define HZ_PROFILE_BEGIN_SESSION(name, filepath) ::Hazel::Instrumentor::Get().BeginSession(name, filepath)
-#define HZ_PROFILE_END_SESSION() ::Hazel::Instrumentor::Get().EndSession()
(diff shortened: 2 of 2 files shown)


LovelySanta:
I'll wait before I decide if this is good or bad, as cherno mentioned several ways of improving this. I'm also unsure how much he will still do himself. For that manner I'll leave this open until he (more or less) finishes this mini series or until he explicitly states some wanted solution.

ReDucTor:
This seems like potentially it's an issue with the wrong time unit used which causes overlap in times, potentially should be using nanoseconds.  Using the approach in this PR would be expensive and lead to incorrect results

Emilundpixeln:
> This seems like potentially it's an issue with the wrong time unit used which causes overlap in times, potentially should be using nanoseconds. Using the approach in this PR would be expensive and lead to incorrect results

The problem with using nanoseconds, is that chrome://tracing/ doesn't support nanoseconds as far as I know. Just outputting in nanoseconds anyway, would mean that an event that is at 1 second will be displayed at 16.7 minutes, which I feel like isn't really an obvious scaling factor.

LovelySanta:
Both of you are right, the scaling factor isn't obvious, so for sure we need to have a 'fix' for this precision. On the other hand, I rather have a semi working version than no version at all. On the other had, coming up with a better fix might also be a good idea, hence why I didn't want to merge it in (yet).

Puddlestomper:
I agree with letting this sit for now. @TheCherno seemed to have a solution in mind (and definitely knows of a few solutions for this) so we should give him an opportunity to implement his fix (hopefully soon).

PR #178, opened 2019-11-28
More compact made the movement of the camera.

KennyProgrammer:
###### Describe the issue #######
I just fixed some code in ``OrthographicCamera.h`` and ``OrthographicCamera.cpp``, namely made the camera movement code more compact and readable than it was before. (in my opinion.)

###### Proposed fix #######
That is, why always write ``cos(glm:: radians(...))`` and ``sin(glm:: radians(...))`` if it can be much simplified to just ``dX`` and ``dY``. Theta is simply the name of the angle around which the camera will rotate. ``dTranslationSpeed`` and ``dRotationSpeed`` are speeds only multiplied by ``delta``. Well, ``increase*()`` and ``decrease*()`` methods are just auxiliary. (they can not be used, but it seems to me that the code looks cleaner).


Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
```diff
@@ -13,41 +13,32 @@ namespace Hazel {
...
-			m_CameraPosition.x += -sin(glm::radians(m_CameraRotation)) * m_CameraTranslationSpeed * ts;
-			m_CameraPosition.y += cos(glm::radians(m_CameraRotation)) * m_CameraTranslationSpeed * ts;
-		}
-		else if (Input::IsKeyPressed(HZ_KEY_S))
-		{
-			m_CameraPosition.x -= -sin(glm::radians(m_CameraRotation)) * m_CameraTranslationSpeed * ts;
-			m_CameraPosition.y -= cos(glm::radians(m_CameraRotation)) * m_CameraTranslationSpeed * ts;
-		}
+		float dTranslationSpeed = m_CameraTranslationSpeed * delta;
+		float dRotationRot      = m_CameraRotationSpeed * delta;
```
Puddlestomper: Not sure why this variable is called `dRotationRot` and not just `dRotation` or something.

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
```diff
@@ -13,41 +13,32 @@ namespace Hazel {
...
+		float theta = m_CameraRotation;
```
Puddlestomper: I don't think there is much point to this variable because there is no processing happening here so we might as well use `m_CameraRotation` rather than having the extra variable.

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
Puddlestomper: `delta` should be `ts`. `delta` is not a variable that exists (unless you plan to make it, but `ts` is perfectly fine).

Hazel/src/Hazel/Renderer/OrthographicCameraController.h
```diff
@@ -21,6 +21,11 @@ namespace Hazel {
 
 		float GetZoomLevel() const { return m_ZoomLevel; }
 		void SetZoomLevel(float level) { m_ZoomLevel = level; }
+		
+		const void increasePosition(float x, float y) { m_CameraPosition.x += x; m_CameraPosition.y += y; }
+		const void decreasePosition(float x, float y) { m_CameraPosition.x -= x; m_CameraPosition.y -= y; }
+		const void increaseRotation(float rotation) { m_Rotation += rotation; }
+		const void decreaseRotation(float rotation) { m_Rotation -= rotation; }
```
Puddlestomper: The names of the functions don't follow the Hazel convention of starting with the uppercase letter. These functions also don't actually update the camera's position. They only modify the controller's variables.

I am also not sure if these functions are really necessary. If we do want to include these we should maybe also make it clear whether this change is in camera space or in world space (like whether calling `IncreasePosition(0.0f, 1.0f)` would move in the camera's up direction or just add 1 to the y coordinate).

I also think we should rather have only one `ChangePosition` method (if we are adding this), since floats can be negative anyway and having both an increase and decrease method isn't needed. Even if we want these methods, should it be part of this PR?

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
```diff
@@ -13,41 +13,32 @@ namespace Hazel {
...
+		if (Input::IsKeyPressed(HZ_KEY_W)) increasePosition(-dY * dTranslationSpeed, dX * dTranslationSpeed);
+		if (Input::IsKeyPressed(HZ_KEY_S)) decreasePosition(-dY * dTranslationSpeed, dX * dTranslationSpeed);
+		
 		if (m_Rotation)
 		{
-			if (Input::IsKeyPressed(HZ_KEY_Q))
-				m_CameraRotation += m_CameraRotationSpeed * ts;
-			if (Input::IsKeyPressed(HZ_KEY_E))
-				m_CameraRotation -= m_CameraRotationSpeed * ts;
+			if (Input::IsKeyPressed(HZ_KEY_Q)) decreateRotation(dRotationRot);
```
Puddlestomper:
```suggestion
			if (Input::IsKeyPressed(HZ_KEY_Q)) decreaseRotation(dRotationRot);
```
Spelling mistake.

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
```diff
@@ -13,41 +13,32 @@ namespace Hazel {
...
+			if (Input::IsKeyPressed(HZ_KEY_E)) increateRotation(dRotationRot);
```
Puddlestomper:
```suggestion
			if (Input::IsKeyPressed(HZ_KEY_E)) increaseRotation(dRotationRot);
```
Spelling mistake.

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
```diff
@@ -13,41 +13,32 @@ namespace Hazel {
...
-			else if (m_CameraRotation <= -180.0f)
-				m_CameraRotation += 360.0f;
+			if (m_CameraRotation  >= 360)
+			{
+				cm_CameraRotation = 0;
+			}
+			else if (m_CameraRotation  <= -360)
+			{
+				m_CameraRotation  = 0;
+			}
```
Puddlestomper: I am biased, but I prefer my way of doing this. It limits the range to (-180; 180] rather than your (-360; 360) and if someone is lagging a lot and the timestep is big enough that the angle jumps a significant amount over the boundary your way will noticeably cause them to jump back to an angle of 0.

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
```diff
@@ -13,41 +13,32 @@ namespace Hazel {
...
 
-			m_Camera.SetRotation(m_CameraRotation);
+			camera.setRotation(m_CameraRotation);
```
Puddlestomper:
```suggestion
			m_Camera.setRotation(m_CameraRotation);
```
The variable is `m_Camera`, not `camera`.

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
KennyProgrammer: Sorry, i forgot it change form my project.

Hazel/src/Hazel/Renderer/OrthographicCameraController.h
KennyProgrammer: Yea, ``ChangePosition'' its better name. I originally wanted to call it that but then I thought that ``Increase, Decrease`` would be clearer.

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
KennyProgrammer: Of course) Really thanks, im from Russia, and can make spelling mistakes(

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
KennyProgrammer: Again, thanks!

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
KennyProgrammer: I want to ``dRotationSpeed``. Im fix it now.

KennyProgrammer: d* means that the speed is already multiplied by delta (ts)

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
KennyProgrammer: Bring back)

Hazel/src/Hazel/Renderer/OrthographicCameraController.h
```diff
@@ -21,6 +21,18 @@ namespace Hazel {
...
+		{ 
+			if(action) { m_CameraPosition.x += x; m_CameraPosition.y += y; }
+			if(!action) { m_CameraPosition.x -= x; m_CameraPosition.y -= y; }
+		}
+		
+		const void ChangeRotation(float action, float rotation) 
+		{ 
+			if(action) m_CameraRotation += rotation; 
+			if(!action) m_CameraRotation -= rotation; 
+		}
```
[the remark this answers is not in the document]
KennyProgrammer: Okay

Hazel/src/Hazel/Renderer/OrthographicCameraController.h
```diff
@@ -21,6 +21,24 @@ namespace Hazel {
 
 		float GetZoomLevel() const { return m_ZoomLevel; }
 		void SetZoomLevel(float level) { m_ZoomLevel = level; }
+		
+		
+
```
LovelySanta: These 3 newlines should become 1 new line

Hazel/src/Hazel/Renderer/OrthographicCameraController.h
```diff
@@ -21,6 +21,24 @@ namespace Hazel {
...
+			m_Camera.SetPosition(m_CameraPosition);
+		}
+		
+		inline void ChangeRotation(float rotation) 
+		{ 
+			m_CameraRotation += rotation;
+
+			m_Camera.SetRotation(m_CameraRotation);
+		}
```
LovelySanta: Maybe put the implementation in the cpp file?

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
```diff
@@ -13,41 +13,28 @@ namespace Hazel {
...
-			m_CameraPosition.x += -sin(glm::radians(m_CameraRotation)) * m_CameraTranslationSpeed * ts;
-			m_CameraPosition.y += cos(glm::radians(m_CameraRotation)) * m_CameraTranslationSpeed * ts;
-		}
-		else if (Input::IsKeyPressed(HZ_KEY_S))
-		{
-			m_CameraPosition.x -= -sin(glm::radians(m_CameraRotation)) * m_CameraTranslationSpeed * ts;
-			m_CameraPosition.y -= cos(glm::radians(m_CameraRotation)) * m_CameraTranslationSpeed * ts;
-		}
+		float dTranslationSpeed = m_CameraTranslationSpeed * ts;
+		float dRotationSpeed    = m_CameraRotationSpeed * ts;
```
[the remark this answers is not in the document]
LovelySanta: That should indeed be fixed, this shouldn't have all those spaces

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
```diff
@@ -13,34 +13,21 @@ namespace Hazel {
...
+		float theta = m_CameraRotation;
+		float dX = cos(glm::radians(theta));
+		float dY = sin(glm::radians(theta));
 
+		if (Input::IsKeyPressed(HZ_KEY_A)) ChangePosition(-(dX * dTranslationSpeed), -(dY * dTranslationSpeed));
+		if (Input::IsKeyPressed(HZ_KEY_D)) ChangePosition(dX * dTranslationSpeed, dY * dTranslationSpeed);
+		if (Input::IsKeyPressed(HZ_KEY_W)) ChangePosition(-dY * dTranslationSpeed, dX * dTranslationSpeed);
+		if (Input::IsKeyPressed(HZ_KEY_S)) ChangePosition(-(-dY * dTranslationSpeed), -(dX * dTranslationSpeed));
```
LovelySanta: As @Puddlestomper said, I'm not a fan of this implementation:
1. The name `ChangePosition` would imply you would give it an absolute value, not a delta (`movePosition` would be more appropriate for example?)
2. I don't really like the functions `ChangePosition` and `ChangeRotation`, I would rather take a compromise between both implementations: calculate the sine & cosine before and use the `dCos`, `dSin` and `dRot` in the if statements. (the names `dx` `dy` and `dtheta` don't realy read nicely, for example check the W and S keys...)

This is just my opinion... I don't really see any improvements over the current implementation other than maybe some readability...

Puddlestomper: I agree with dropping the methods. This PR is really just for readability, so introducing the new methods isn't really needed.

KennyProgrammer: I agree with @LovelySanta that ``Change *`` is not really needed. I just roughly suggested how to implement this. Calculation of sine and cosine, I originally wanted to make it calculated above as one line (or two). But I just write my engine in a slightly different way, and just relying on the lessons of Cherno and I don’t know if this will work correctly in Hazel. I'll see if I can still find better ways to do this.  

This is just my suggestion, and I'm almost 99% sure that Cherno will somehow implement it in its own way. Especially when GameObject's system appears.

Hazel/src/Hazel/Renderer/OrthographicCameraController.h
KennyProgrammer: Fixed.

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
Puddlestomper: Resolved in c336b21.

Puddlestomper: Resolved in c336b21.

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
Puddlestomper: Resolved in c336b21.

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
Puddlestomper: Resolved in c336b21.

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
Puddlestomper: Resolved in 644dbf1.

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
Puddlestomper: Resolved in 157ddda.

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
```diff
@@ -13,41 +13,28 @@ namespace Hazel {
...
+			if (Input::IsKeyPressed(HZ_KEY_Q)) ChangeRotation(0, dRotationSpeed);
+			if (Input::IsKeyPressed(HZ_KEY_E)) ChangeRotation(1, dRotationSpeed);
 
 			if (m_CameraRotation > 180.0f)
 				m_CameraRotation -= 360.0f;
 			else if (m_CameraRotation <= -180.0f)
 				m_CameraRotation += 360.0f;
 
-			m_Camera.SetRotation(m_CameraRotation);
+			m_camera.setRotation(m_CameraRotation);
```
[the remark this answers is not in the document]
Puddlestomper: Resolved in 480a1ab.

Hazel/src/Hazel/Renderer/OrthographicCameraController.h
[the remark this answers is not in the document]
Puddlestomper: Resolved in 9e545fa.

Emilundpixeln:
This potentially recalculates the MVP matrix four times, which is a bit wasteful.

LovelySanta:
I've updated this PR to the current commit of Hazel. I'll start now reworking some of the changes mentioned above.

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/178/commits/fb258084c6f1397579435b42d8daf5377d4dcfb7.

Hazel/src/Hazel/Renderer/OrthographicCameraController.h
LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/178/commits/fb258084c6f1397579435b42d8daf5377d4dcfb7.

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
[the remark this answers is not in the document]
LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/178/commits/fb258084c6f1397579435b42d8daf5377d4dcfb7.

LovelySanta:
Now that I'm looking back on this, let's evaluate the performance.

I assume that most of the time you do not move around, and if you move around, you don't press both directions, resulting in not moving at all. If you look at the current code:
https://github.com/TheCherno/Hazel/blob/fb258084c6f1397579435b42d8daf5377d4dcfb7/Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp#L18-L31
Then no matter if you move or you don't, you always calculate the `dCos` and `dSin`. If we look at how the code was before:
https://github.com/TheCherno/Hazel/blob/7fa595072c32abf1770b82954cfcc92519006b3a/Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp#L18-L27
It did only calculate it once (yes I know it has priority on one direction over the other, this is the smallest edge case, which should not occur at all, worst case, you calculate it twice in this edge).

This said, I think calculating it once every frame, versus calculating it when moving, that calculating it when moving is better.

Taking the W and S movement, we do calculate it another time, but I would say that calculating this twice has the same reasoning as I just made above.

So I feel that this PR does only give some extra readability without providing any functionality. Hence why I would not merge this in.

PR #179, opened 2019-11-29
Instrumentation fixes for #176 and #181

### LINKED ISSUE #176: Instrumentation is not thread safe
Not sure if this is your intention, but your Instrumentation is not thread safe.

The way I solved it is by adding a lock around `WriteProfile`.

```cpp
// ...
std::mutex _mtx;  

// ...
void WriteProfile(const ProfileResult& result)
{
  _mtx.lock();
  try
  {
    // ...
  }
  catch( ... )
  {
  }
  _mtx.unlock();
}
```
Let me know if you would like me to submit a PR but I suspect you already have your own approach to this.

Of course the price of the lock is kind of rendering the profiling pointless in multi-threading cases.

### LINKED ISSUE #181: Mismatched Instrumentor::BeginSession()/EndSession() breaks result file and leaks memory
#### Describe the bug
Instrumentor::BeginSession() does not check to see if there is already an instrumentation session active.
The existing session (if it exists) is not ended properly which a) means the results json file wont load nicely into chrome://tracing because it is not complete, and b) leaks an InstrumentationSession instance.

#### To Reproduce
The following example is somewhat contrived (because the instrumentation code is for Hazel devs only, so wouldn't normally be found in "client" code.  Nevertheless, it does illustrate the problem:

If/When Hazel becomes multi-threaded, there becomes more ways to reproduce the problem.

Add the following to SandboxApp.cpp, which you might think could be used to profile the creation of SandBox in a separate instrumentation session.

`Hazel::Application* Hazel::CreateApplication()
{
	Hazel::Instrumentor::Get().BeginSession("innocent mistake", "mistake.json");  // <- -Add this line
	Hazel::Application* app = new Sandbox();
	Hazel::Instrumentor::Get().EndSession();  // <-- Add this line
	bool result = handle.get();
	return app;
}
`

#### Expected behavior
All profile result .json files should be valid json, and load cleanly into chrome://tracing
InstrumentationSession objects should not be leaked.

- OS: N/A  problem is not OS specific

freeman40:
Some small tidy ups to Instrumentor:
- Added mutex to Instrumentor to avoid jumbled output if profiling multiple threads.
- ~~Removed~~ ~~unused~~ ~~InstrumentationSession.~~  ~~Note~~ ~~that~~ ~~InstrumentationSession~~ ~~was~~ ~~also~~ ~~a~~ ~~potential~~ ~~memory~~ ~~leak~~ ~~if~~ ~~BeginSession()/EndSession()~~ ~~are~~ ~~not~~ ~~perfectly~~ ~~paired.~~
- Used InstrumentationSession to ensure that new a BeginSession() will cleanly end any previously existing session.
- Removed std::hash<> hack on ThreadId.  The code is cleaner without it (and still works as far as I can tell)

#### PR impact
Fixes #176 
Fixes #181 

#### Additional context
- There is still a memory leak if an instrumentation session is begun but not ended when app exits.
This could be easily fixed (by making m_CurrentSession into a Hazel::Scope, but it comes at the cost of having to add a constructor for InstrumentationSession  (Hazel::CreateScope<InstrumentationSession>({name}) will not work)

- One could remove InstrumentationSession (this was my first idea), and use m_OutputStream.is_open() in place of checking for non-null m_CurrentSession.  However, I've left InstrumentationSession in there as it could come in useful later (for example: If it is desired to create instrumentation sessions that only capture particular categories of profiling output - you will need to track which categories are active in the InstrumentationSession instance)

- I have not addressed the concurrent start timestamp issue, as Cherno has said he is going to address it in an upcoming episode, and he probably has his own idea for how to fix it.
(my suggestion would be a thread_local static counter that is incremented/decremented automatically via InstrumentationTimer RAII.  Just add this counter on to the timestamp.  (yes, it will result in timestamps that are a few ticks off, but a tick here is 0.001 of a millisecond... it shouldn't matter)


WhoseTheNerd:
Why does code commited have whitespace inconsistency?

freeman40:
There are some problems with BeginSession() / EndSession() as currently designed.
1) If/when multiple threads are introduced then there will be no guarantee that Begin/End session calls are nicely paired up - there probably needs to be one instance of Instrumentor per thread (instead of global singleton).
2) Even within one thread, there is nothing stopping someone from accidentally BeginSession()'ing without EndSession()'ing first...  with the result that the output .json files are broken.

freeman40:
> 
> 
> Why does code commited have whitespace inconsistency?

Whoops, sorry.  I use spaces.  Cherno uses tabs.   Should there be a .editorconfig?

WhoseTheNerd:
Doesn't .editorconfig do nothing?

Puddlestomper:
Either way, we have noted in our [contribution guidelines](https://github.com/TheCherno/Hazel/blob/master/.github/CONTRIBUTING.md#coding-conventions) that we use tabs and not spaces.

freeman40:
> 
> 
> Doesn't .editorconfig do nothing?

I was under the impression that it could be used to control things like indentation style.
I have not actually used one myself. [EditorConfig](https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2019)

@Puddlestomper  yes, I know.  I missed that I'd used spaces.  Sorry.

freeman40:
> 
> 
> Doesn't .editorconfig do nothing?
> Also you still didn't fix the white-space inconsistencies.

@WhoseTheNerd   Didn't I?  Can you give me a hint, because I'm not seeing it.

Puddlestomper:
We should consider leaving the `InstrumentationSession` struct in (even if we never use it). I think @TheCherno probably put it there in order to support multiple sessions in the future.

We definitely should fix the memory leak though. We can remove the code that uses the struct (as you did), or just delete the pointer it in the `Instrumentor` destructor if it's not a `nullptr`.

freeman40:
Yes, thinking about this some more, I agree that InstrumentationSession should be left in there, as it can be used for multiple session support.  I will play around with this a bit more, and update the PR.

WhoseTheNerd:
> > Doesn't .editorconfig do nothing?
> > Also you still didn't fix the white-space inconsistencies.
> 
> @WhoseTheNerd Didn't I? Can you give me a hint, because I'm not seeing it.

Now it is fixed.

freeman40:
I think this is good to go now.  I have updated the PR description to more accurately describe what I ended up doing.
It might all be a bit overkill actually, given that the profiling stuff is intended only for internal use by Hazel devs.  We could just keep the Instrumentor simple, and mandate that Hazel devs don't do stupid stuff with Begin/End session.   (the minimum that needs to be kept is the mutex in WriteProfile() - assuming that Hazel will be multi-threaded one day)

What do others think?

Hazel/src/Hazel/Debug/Instrumentor.h
```diff
@@ -34,41 +35,63 @@ namespace Hazel {
...
-			m_OutputStream << "{";
-			m_OutputStream << "\"cat\":\"function\",";
-			m_OutputStream << "\"dur\":" << (result.End - result.Start) << ',';
-			m_OutputStream << "\"name\":\"" << name << "\",";
-			m_OutputStream << "\"ph\":\"X\",";
-			m_OutputStream << "\"pid\":0,";
-			m_OutputStream << "\"tid\":" << result.ThreadID << ",";
-			m_OutputStream << "\"ts\":" << result.Start;
-			m_OutputStream << "}";
+			std::lock_guard lock(m_Mutex);
```
ReDucTor: This looks like an extremely costly time to hold onto a lock

freeman40: Agreed.  It could be improved by formatting the json to a local stringstream first, and then locking (or not, see comments on #176) only while the stringstream is sent to m_OutputStream.

However, first need to agree on the design of WriteProfile() in general, per your other comment on #176

freeman40: I've committed an update to hold the lock for less time.

freeman40:
@ReDucTor   Thanks, I missed that.  Reason why code reviews are good.

LovelySanta:
Another thought of mine, shouldn't we split into .h and .cpp file as the file grows?

Emilundpixeln:
I think there is a possibility this will try to write to an already closed stream. 
For this thread 1 has to be writing a `ProfileResult` and be past the `m_HasSession` check while thread 2 calls `EndSession`, locks the mutex and closes `m_OutputStream`. Then thread 2 gets the mutex and tries to write to `m_OutputStream`, which has been closed.

freeman40:
@Emilundpixeln  Good catch, thanks.  Need to check that there still is a session after getting the lock in WriteProfile().  After that it seems like the code is a bit cleaner to just check m_CurrentSession and not bother with the atomic<bool> m_HasSession at all - so I've stripped that out.  It means that if one thread is still profiling when another thread has ended the session, then the still-profiling thread will waste a bit of time formatting json which is then never used.  I don't think we should get too hung up on that - its kind of an edge case error situation anyway.

freeman40:
> 
> 
> Another thought of mine, shouldn't we split into .h and .cpp file as the file grows?

I've left as-is for now.  Hopefully the file wont get any bigger for a while (maybe some day when a better profiling sub system is needed).

Hazel/src/Hazel/Debug/Instrumentor.h
```diff
@@ -23,55 +22,77 @@ namespace Hazel {
...
+				InternalEndSession();
+			}
 			m_OutputStream.open(filepath);
-			WriteHeader();
-			m_CurrentSession = new InstrumentationSession{ name };
+			if (m_OutputStream.is_open()) {
+				m_CurrentSession = new InstrumentationSession({name});
+				WriteHeader();
+			} else {
+				HZ_CORE_ERROR("Instrumentor could not open results file '{0}'.", filepath);
```
Emilundpixeln: Shouldn't the `if (Log::GetCoreLogger())` check be here aswell?

LovelySanta: Good catch, it should indeed!

Issue #182, opened 2019-12-01
Profile result .json files should not be under version control


freeman40:
#### Is your feature request related to a problem? Please describe
Not really a "feature request", but more of a suggestion.

The result files from profiling (HazelProfile-[Startup|Runtime|Shutdown].json) shouldnt be under version control.

In order to not to have to keep remembering to not-commit these files, it would be better if they were created in a directory outside of the source tree.

#### Describe the solution you'd like
Remove these .json files from version control.
When creating instrumenation session, create the files in temporary directory instead of the current directory (which is likely to be within the hazel source tree)

#### Describe alternatives you've considered
I considered just adding the files to .gitignore.  But this doesn't seem like the right solution
(e.g. what if someone runs Sandboix app from a different working directory.  What if someone adds a new session (you'd then need to remember to update the .gitignore)

PR #183, opened 2019-12-01
Remove profile result json files from version control.

0xworks:
Implements #182


Puddlestomper:
Also, adding on to what FFMG said, where exactly is this temporary directory created? I definitely agree that these profiling files shouldn't be on the GitHub repo, but adding them to the gitignore or making a dedicated folder for them or both seems like a better idea than creating files outside of the solution folder to me.

freeman40:
@Puddlestomper  The temporary directory isn't created - it is a guaranteed already-exists location.  It's %TEMP% on windows, or $TMPDIR on mac/linux (or failing that /tmp).

@FFMG  The profile result files weren't going to the exe directory.  They were going to the current working directory which may or may not be the exe directory.  For example, if you just run the project from within Visual Studio on Windows, then the profile result files end up in the Sandbox project directory (could be why they ended up getting committed)

I did consider just adding the filenames to .gitignore, but decided against that because the actual problem is that the files are created in the wrong place to begin with.  So adding to .gitignore seems like a workaround rather than a fix.
 
I chose the temp directory as the location to store these files as that seemed a natural place for files that we only want to look at for a little bit while doing profiling runs.

I'm totally fine with the exe folder being the location for these files though, (and this wouldn't require any change to .gitignore, since "bin" is already ignored).  If someone can tell me a cross-platform std-c++ way to get the current exe directory, then I can update the PR.

Puddlestomper:
Looking at %TEMP% on my windows machine, it's really cluttered (~300 items) and I don't want to have to go looking through that every time I just want to check my profile result for Hazel. I would rather we keep it as is for now and add it to the .gitignore or make a dedicated folder for it as you suggested.

LovelySanta:
What about just outputting it in a profiling folder or something?

freeman40:
Yes, it would be more convenient to have the result files in the exe directory.  I'll update the PR.

TheCherno:
Profile results should definitely not be commited (that's a mistake), however I don't think they should be explicitly moved to the .exe directory. They should always be relative the working directory, whether or not we put that into a separate folder is a more realistic question. It's nice to have them inside the project hierarchy in Visual Studio, so I'll leave them here for now.

HerrDingenz:
Simply just add the profile files to .gitignore

LovelySanta:
> Simply just add the profile files to .gitignore

We could rename them to `*.profiling.json` or something, and then we can indeed add them to the .gitignore, just adding *.json would be a bad idea, and hardcoding the names is only a temporary measure...

HerrDingenz:
You are right @LovelySanta rename the files and then add them to .gitignore.

@freeman40 if you run the exe file from your bin folder they will be created right there next to the exe file. They only get created inside the sandbox project folder if you run it inside VS.

TheCherno:
Closed (obsolete).

PR #185, opened 2019-12-13
Fixed Instrumentor generating same start time

Penuvil:
Instrumentor clock resolution is too low and can create duplicate start times for closely executed functions.

Other PRs this solves    | closes #175

Change start and stop times to nanoseconds, static cast to long double and divide by 1000 so Chrome tracing will accept the input. Chrome tracing will not accept nanoseconds directly but will accept floating point microsecond input.

#### Additional context
This will fail once nanoseconds since epoch becomes too large to fit in a double.
If higher resolution than nanoseconds is needed Chrome Tracing might not be the best solution because you would have to manually manipulate start times to avoid duplicates after that.


Hazel/src/Hazel/Debug/Instrumentor.h
```diff
@@ -4,14 +4,15 @@
...
+#include <iomanip>
 
 #include <thread>
 
 namespace Hazel {
 	struct ProfileResult
 	{
 		std::string Name;
-		long long Start, End;
+		long double Start, End;
```
WhoseTheNerd: Would suggest `uint64_t`.

Penuvil:
It needs to be written to the json file in a floating point format for
tracing to process it properly.

LovelySanta:
Similar note as I wrote in https://github.com/TheCherno/Hazel/pull/175#issuecomment-557837056: Cherno has something in mind, he might fix it himself. If it still takes too long, we can always merge a PR to fix this issue.

Penuvil:
Copy, no worries.  I only threw it out there because of this comment...

[Timer Comment](https://youtu.be/FtsehM4xoDw?t=2211)

Lecrapouille:
It can also be nice to replace high_resolution_clock by steady_clock since high_resolution_clock is not recommended. For example, system_clock may suffer from time forward/backward.

Hazel/src/Hazel/Debug/Instrumentor.h
Lecrapouille: why long double ? Is this really archi independent ? Why not simply double ? https://en.cppreference.com/w/cpp/language/types say extended precision floating point type. Does not necessarily map to types mandated by IEEE-754. Usually 80-bit x87 floating point type on x86 and x86-64 architectures.

Hazel/src/Hazel/Debug/Instrumentor.h
```diff
@@ -107,8 +109,8 @@ namespace Hazel {
 		{
 			auto endTimepoint = std::chrono::high_resolution_clock::now();
 
-			long long start = std::chrono::time_point_cast<std::chrono::microseconds>(m_StartTimepoint).time_since_epoch().count();
-			long long end = std::chrono::time_point_cast<std::chrono::microseconds>(endTimepoint).time_since_epoch().count();
+			long double start = static_cast<long double>(std::chrono::time_point_cast<std::chrono::nanoseconds>(m_StartTimepoint).time_since_epoch().count()) / 1000;
```
Lecrapouille: Should not you explicitly cast 1000 (integer) into long double ? Maybe try to compile with flags -Wconversion.
I'm a noob concerning chrono but can nanosecond may overflow for long durations? May you try to measure/profile a sleep(5 minutes) to see if this is ok. Can we imagine a case of profiling a game for one hour or more?

Hazel/src/Hazel/Debug/Instrumentor.h
```diff
@@ -107,8 +109,8 @@ namespace Hazel {
 		{
 			auto endTimepoint = std::chrono::high_resolution_clock::now();
```
Lecrapouille: auto endTimepoint = std::chrono::steady_clock::now();

Lecrapouille:
Here my modified version https://gist.github.com/Lecrapouille/1341f9392fb79831efe5ff7f42374c54 (sorry for the moment I changed namespace and method names because I tested for my personal project I will tomorrow restore Hazel's names. I this version I also removed InstrumentationSession and ProfileResult to avoid useless copying data.

Hazel/src/Hazel/Debug/Instrumentor.h
LovelySanta: Even for such a rare case, my solution would be to start a new session (from time 0) with a name like '.... (2)' or something when it overflows...

Lecrapouille: For long durations (> minutes) I think you will do not care about nanoseconds ;) my idea would be:
- try casting from nanoseconds
- check if overflowed (if possible)
- try casting from milliseconds

But I'm not an expert of chrono.

Penuvil:
Been pretty busy with holidays and sick kids, I will look over all of this as soon as I get a chance.

Hazel/src/Hazel/Debug/Instrumentor.h
Lecrapouille: I watched CppCon 2016: Howard Hinnant “A chrono Tutorial" https://www.youtube.com/watch?v=P32hvk8b13M I understood now how this lib is working I think we can simplify code

HerrDingenz:
"Replaced m_OutputStream.flush(); by std::endl (some text editor failed when lines are too long)"

didn't you watch Jason Turners video about NOT using std::endl and instead calling std::flush on your own once work is done?

https://www.youtube.com/watch?v=GMqQOEZYVJQ&list=PLs3KjaCtOwSZ2tbuV1hx8Xz-rFZTan2J1&index=7

Lecrapouille:
@HerrDingenz I know that forcing a flush takes more time than letting filling the buffer until completely full which makes an automatic flush (~1024 to 4096 bytes). But you are right we can try to optimize this part. This document http://www.dil.univ-mrs.fr/~regis/prog/pdf/c-sys.4b.stdio-bufferisation.pdf describes it better than your video (sorry in french but it's the best I found yet). I summarized pages 9 and 11:

There are 3 modes of buffering: 
- unbuffered (such as stderr)
- fully-buffered (such as stdout when not tty). Fully-buffered is flushed when one of the following conditions is met: the internal buffer is full or flush() is called or when close() is called.
- and line-buffered (such as stdout when used as tty). Line-buffered is flushed when '\n' is met or when one of three conditions upper is met.

According to your video std::ofstream looks to be fully-buffered which is in accordance to my document. Anyway, the original code already called flushes, I just add an extra carriage return to avoid making some text editors crying when opening the file. I did nothing more! But yes, we may replace std::endl by '\n' but this not Windows compliant.

Another solution would be to store in RAM ProfileResult (with a std::dqueue ?!) and when WriteFooter() is called, convert all ProfileResults as JSON format and write lines in the file.

Hazel/src/Hazel/Debug/Instrumentor.h
Penuvil: Changed to use a custom chrono::duration with double for Start and changed End to chrono::microseconds ElapsedTime.  End was only used to determine elapsed time anyway.

Hazel/src/Hazel/Debug/Instrumentor.h
Penuvil: Got rid of the static_cast and /1000, using chrono to do the conversion now.  Also changed duration back to chrono:::microseconds since we only really care about the nanoseconds on the start time to prevent duplicates...

Hazel/src/Hazel/Debug/Instrumentor.h
Penuvil: Switched to steady_clock

Penuvil:
Update...
Switched to steady_clock
Got rid of static_cast
Made better use of chrono features

As pointed out there are some other issues but this PR was intended to fix the duplicate start times.  I'm personally not a big fan of trying to address too many issues in one PR.

LovelySanta:
This looks fine as it is to solve the issue, should we open new issues that are mentioned here, and link to this PR?

Penuvil:
Looks like the memory leak and multiple BeginSession() issues, m_ProfileCount and ThreadID are cleaned up in PR #179.  I could update this PR to include the minor change to m_Stoppped if you want and the suggested change to std::replace should probably be it's own issue.

LovelySanta:
Make the two seperate PR's one for m_Stopped instead of m_CurrentSession and one for the std::replace. This will be easier to resolve merge conflicts

Lecrapouille:
I did not notice PR 179 when I made my suggestion. It looks fine.

LovelySanta:
I merged in #179, and I'm trying to resolve the merge conflicts, but whatever I do, I keep getting duplicated start times and some weird outputs.. (which is not caused by #179).

Can you resolve the merge conflicts? Or want me to push what I currently have and you go from there?

Penuvil:
I will work on the conflicts and test again

Penuvil:
It should be good now, test it on your rig an let me know if you have any problems.

LovelySanta:
> It should be good now, test it on your rig an let me know if you have any problems.

Works for me! Found my mistake: I misplaced the `std::setprecision(3)`. I left it in the m_OutputStreaminstead of moving it over to the json stream. Tested it multiple times, not a single duplicate starting time found, and all profilers are loading and are fine (startup, runtime and shutdown).

PR #171, opened 2019-11-19
Add key mapping and Update CameraController using improved Input system.

3x3y3z3t:
#### Describe the issue (if no issue has been made)
This PR doesn't fix any issue.
This PR will improve the existing `Input` and `OrthographicCameraController` described below.

#### PR impact _(Make sure to add [closing keywords](https://help.github.com/en/articles/closing-issues-using-keywords))_
List of related issues/PRs this will solve:

 Impact                  | Issue/PR
------------------------ | ------
Other PRs this solves    | #161, #124  

#### Proposed fix _(Make sure you've read [on how to contribute](https://github.com/TheCherno/Hazel/blob/master/.github/CONTRIBUTING.md) to Hazel)_
This PR adds/improves the following:
- KeyCodes/MouseButtonCodes:
  - Add `enum class` `Keys` and `MouseButtons` (#124).
  - In the future version, macros will be replaced completely by those enum classes.
- Input: 
  - Add `KeyAlternative` struct which represent a single key press or mouse button press.
  - Input checks now checks over the said `KeyAlternative` instead of raw keycodes/mousebuttoncodes `int` values.
  - Add a map for key mapping: each `std::string` key is assigned with one `KeyAlternative`, and are remappable at any time/any place.
- OrthographicCameraController:
  - Update Controller to take advantage of the new Keymapping system.
  - **[minor]** Zoom Level and Enable Rotation can now be manually set at any time/any place. There is a min/max zoom level limitation.
  - **[minor]** Camera translation speed now scales with the zoom level (#161).

This PR is backward-compatible with current version of `master` branch Hazel (need minor variable rename though).

#### Additional context
None.

PR #186, opened 2019-12-19
Add key mapping

3x3y3z3t:
This PR will improve the existing `Input` and described below.

Other PRs this solves    | closing #171 

This PR adds/improves the following:
- Input: 
  - Add `InputKey` struct which represent a single key press or mouse button press.
  - Input checks now checks over the said `InputKey` instead of raw KeyCodes/MouseCodes enum values.
  - Add an enum `InputName` for holding the name of an input to be mapped.
  - Add a map for key mapping: each `InputName` key is assigned with one `InputKey`, and are remappable at any time/any place.
- OrthographicCameraController:
  - Update Controller to take advantage of the new Keymapping system.

#### Additional context
User can also add their own `InputName` value at user-level code.


Hazel/src/Hazel/Core/Input.h
```diff
@@ -14,12 +36,39 @@ namespace Hazel {
...
 		inline static bool IsKeyPressed(KeyCode key) { return s_Instance->IsKeyPressedImpl(key); }
 
 		inline static bool IsMouseButtonPressed(MouseCode button) { return s_Instance->IsMouseButtonPressedImpl(button); }
 		inline static std::pair<float, float> GetMousePosition() { return s_Instance->GetMousePositionImpl(); }
 		inline static float GetMouseX() { return s_Instance->GetMouseXImpl(); }
 		inline static float GetMouseY() { return s_Instance->GetMouseYImpl(); }
+
+		/* Map a Key/Mouse Button to an InputKey, and overwrite the old value if existed. */
+		template <typename T>
+		static void Remap(T name, InputKey inputKey) { s_Instance->m_Mapping[static_cast<uint16_t>(name)] = inputKey; }
```
Emilundpixeln: Why not simply `static void Remap(InputName  name, InputKey inputKey)` ?

3x3y3z3t: The user may want to use their own custom input name, for example `InputName::JumpAndShoot` for a Right Click, which is not existed in the engine's `InputName` enum. If the function only accept `InputName` parameter then the user has to modify the engine code, which is what I want to avoid.
By making it a template function like that, user can define their own enum as long as its underlying type match the engine's one, which is `uint16_t` currently.

Line 70 should be `template <typename T : uint16_t>`.

PR #194, opened 2020-01-21
Premake organization

lhurt51:
More of a feature or optimization, it will help mitigate the bloated example code within our main sandbox application.

N/A                         |   N/A

The fix would be that our project will be broken into two separate visual studio files but both files access the same ghost engine project files.

#### Additional context
Recently Cherno made a new OpenGL demo test-bed(https://github.com/TheCherno/OpenGL) and I thought that some of the premake optimizations he made could be really useful within our game engine. So I have optimized our projected generation by having two project solutions within the file. Both visual studio solutions reference the same game engine so both will be updated if either makes changes to the engine. This allows us to have a place to store important examples that future users may want to reference.


LovelySanta:
The reason that I'm not merging this, is that I think that the examplelayer and Sandbox2D should be in the same project. ExampleLayer is basicaly Sandbox3D, but we did not name it like that...

For now it could make sense to split them, but after (a long time) when we finish the 2D renderer, the 2D renderer will also be part of the 3D renderer (it is just a subrenderer), so then it wouldn't make sense.

I think it would make more sense to have a seperate project for the rederer example and for example another one for the ECS or something like that, the big chunks of the engine..

Any thoughts?

lhurt51:
@LovelySanta I believe that you are confusing the examples with the renderer. The reason I moved the example layer to a separate project was just to clean up our current working directory but still have a reference to our old examples! I believe as this project further progresses this system of premake would be very beneficial, such as Engine Examples for new developers that want to learn the engine quickly. Also, I believe that we won't be working from those old examples in the future and will be creating new and improved examples with the current tech. Just my thoughts.

LovelySanta:
> @LovelySanta I believe that you are confusing the examples with the renderer.

If you look in the [`ExampleLayer.cpp`](https://github.com/TheCherno/Hazel/blob/master/Sandbox/src/ExampleLayer.cpp) file, you'll see that this contains the initial render code to render textures and let us move around it in 3d (using the Renderer)

In the series after that, we moved to the Renderer2D, and made a new layer Sandbox2D (see SandboxApp).

We have discussed this in other threads already to change the name of examplelayer to Sandbox3D
We'll come back to the example layer for sure, we're just focussing on the 2D renderer first.

lhurt51:
> If you look in the [`ExampleLayer.cpp`](https://github.com/TheCherno/Hazel/blob/master/Sandbox/src/ExampleLayer.cpp) file, you'll see that this contains the initial render code to render textures and let us move around it in 3d (using the Renderer)
> 
> In the series after that, we moved to the Renderer2D, and made a new layer Sandbox2D (see SandboxApp).

I agree we will be coming back to that code but I figured this would be a good step in preparing for the future since it is just organizational stuff. So would you suggest that I leave the premake set up so that it is easily extendable and we have a separate premake build for the engine and the project? Then ill get rid of the example app for now and add the previous code back as Sandbox3DLayer or 3DExampleLayer!

LovelySanta:
I'll leave this PR here, but until we have some decent example I don't really see a point in this/requirement...

lhurt51:
That is entirely up to you! I agree that there probably isn't a point in this pull request yet other than organization. However, it does break up our premake files making it a bit easier to read. It also clarifies the ExampleLayer as Sanbox3D, which, as you say is a suggested change.

PR #196, opened 2020-01-31
fix issue 193

### LINKED ISSUE #193: Hazel closes when OpenGl < 4.5
#### Describe the bug
As of now, we check the OpenGL version and create an error when we do not meet the requirements:
https://github.com/TheCherno/Hazel/blob/0716731c8e274f2e3d83aefef2222fe189a33728/Hazel/src/Platform/OpenGL/OpenGLContext.cpp#L30-L35

However, depending on the VS install, the default settings might close the debugger when the program exits (after the error), so the user has no way of noticing the error (let alone read it)

#### To Reproduce
Start Hazel with an OpenGL version < 4.5 and you won't be able to read the error

#### Expected behavior
Have a way of notifying the user that they do not meet the minimal requirements, in this case we (discord conversation) where thinking of implementing the log file(s) so we can see the errors, even if Hazel closes unexpectedly (or unintended).

HerrDingenz:
As stated in Issue #193 there is a need for the implementation of file loggers.

Issues this solves       | Closes #193

I implemented a file logger for both Hazel Core and Hazel Client.
The Hazel Core file logger will create a file named "hazel-log.txt".
The Client file logger will create a file named "client-log.txt".

The console loggers will only be available in Debug builds while the file loggers will be available in Debug and Release builds.

NOTE:
I changed an if-statement in Instrumentor.h --- It now checks if the file logger is initialized.

#### Additional context
I tested and ran both Debug and Release build at it seems to work like a charm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 **/bin/
 !vendor/bin/premake/
 bin-int/
+
+# Hazel files
 *.log
 
 # Visual Studio files and folder
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 **/bin/
 !vendor/bin/premake/
 bin-int/
+*.log
 
 # Visual Studio files and folder
 .vs/
--- a/Hazel/src/Hazel/Core/Log.cpp
+++ b/Hazel/src/Hazel/Core/Log.cpp
@@ -25,7 +25,6 @@ namespace Hazel {
 		s_ClientLogger = std::make_shared<spdlog::logger>("APP", begin(logSinks), end(logSinks));
 		spdlog::register_logger(s_ClientLogger);
 		s_ClientLogger->set_level(spdlog::level::trace);
-
 	}
 
 }
--- a/Hazel/src/Hazel/Core/Log.cpp
+++ b/Hazel/src/Hazel/Core/Log.cpp
@@ -1,5 +1,5 @@
 #include "hzpch.h"
-//#include "Hazel/Core/Log.h"		included in hzpch.h already
+#include "Hazel/Core/Log.h"
 
 #include <spdlog/sinks/basic_file_sink.h>
 
@@ -14,11 +14,11 @@ namespace Hazel {
 		s_ConsoleSink = std::make_shared<spdlog::sinks::stdout_color_sink_mt>();
 
 		std::vector<spdlog::sink_ptr> coreSinks;
-		coreSinks.emplace_back(s_ConsoleSink); // VS debug console
+		coreSinks.emplace_back(s_ConsoleSink);
 		coreSinks.emplace_back(std::make_shared<spdlog::sinks::basic_file_sink_mt>("Hazel-Core.log", true));
 
 		std::vector<spdlog::sink_ptr> clientSinks;
-		clientSinks.emplace_back(s_ConsoleSink); // VS debug console
+		clientSinks.emplace_back(s_ConsoleSink);
 		clientSinks.emplace_back(std::make_shared<spdlog::sinks::basic_file_sink_mt>("Hazel-Client.log", true));
 
 		s_CoreLogger = std::make_shared<spdlog::logger>("HAZEL", begin(coreSinks), end(coreSinks));
--- a/Hazel/src/Hazel/Core/Log.cpp
+++ b/Hazel/src/Hazel/Core/Log.cpp
@@ -6,37 +6,32 @@
 
 namespace Hazel {
 
-	#ifdef HZ_DEBUG
-		Ref<spdlog::logger> Log::s_CoreDebugLogger;
-		Ref<spdlog::logger> Log::s_ClientDebugLogger;
-	#endif // HZ_DEBUG
+	Ref<spdlog::logger> Log::s_CoreLogger;
+	Ref<spdlog::logger> Log::s_ClientLogger;
 
-	Ref<spdlog::logger> Log::s_CoreFileLogger;
-	Ref<spdlog::logger> Log::s_ClientFileLogger;
 
 	void Log::Init()
 	{
-		#ifdef HZ_DEBUG
-			s_CoreDebugLogger = spdlog::stdout_color_mt("HAZEL");
-			HZ_ASSERT(s_CoreDebugLogger, "Could not initialize Core Debug Logger!");
-			s_CoreDebugLogger->set_pattern("%^[%T]: %v%$");
-			s_CoreDebugLogger->set_level(spdlog::level::trace);
+		// create the sinks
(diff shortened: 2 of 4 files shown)


HerrDingenz:
Thanks for your review.

The idea using sinks is pretty nice but:

I didn't use sinks because i wanted to use different patterns for both loggers. I think all that timestamp information just inflates the file logs - and we really don't need that information in there.

For now the macros do the trick - even though we have 4 different loggers now (at least in debug builds). In release builds there are only 2 loggers (the file loggers obviously)

Regarding the filenames you are right - they should fit the other style of hazel filenames.

HerrDingenz:
Ok .. I changed it to use sinks.

But i am not really happy with the result, because now we have the timestamps AND the logger name inside the log files.

Not only increases this the file size - it is useless information!
We shouldn't care about what time the log message occured - cause the files get created per session anyway.
And since we have 2 different files (one for core and one for client) the logger name inside the files is useless in there, too.

For output to the console this information was useful (i mean the name of the logger) and we need that preset there because of color formatting. But for the file loggers it is just data garbage nobody needs inside those files.

And as i stated in my last comment already: we can't use different presets for the loggers anymore when switching to multiple sinks for one logger.

Hazel/src/Hazel/Core/Log.cpp
```diff
@@ -1,21 +1,36 @@
 #include "hzpch.h"
-#include "Hazel/Core/Log.h"
+//#include "Hazel/Core/Log.h"		included in hzpch.h already
```
CleverSource: This should just be removed, someone can easily find out that it's included in the pch without this here.

Hazel/src/Hazel/Core/Log.cpp
```diff
@@ -1,21 +1,36 @@
...
 	Ref<spdlog::logger> Log::s_CoreLogger;
 	Ref<spdlog::logger> Log::s_ClientLogger;
+	Ref<spdlog::sinks::stdout_color_sink_mt> Log::s_ConsoleSink;
 
 	void Log::Init()
 	{
+		s_ConsoleSink = std::make_shared<spdlog::sinks::stdout_color_sink_mt>();
+
+		std::vector<spdlog::sink_ptr> coreSinks;
+		coreSinks.emplace_back(s_ConsoleSink); // VS debug console
```
CleverSource: This isn't necessarily required, but it won't always be the Visual Studio debug console, which somewhat makes this comment useless.

Hazel/src/Hazel/Core/Log.cpp
LovelySanta: Cherno mensioned this in one of his videos, we should not rely in Hazel files being included in the precompiled header. The reason we added it to the pch in the first place was because we won't update it often. And especialy in this case where the cpp file should (always) include its header file, so can you please put it back?

HerrDingenz: that is done already .. the include parts of the cpp file are as they have been before my pull request

HerrDingenz:
Thanks for your review again, LovelySanta...

I made some changes to the code.

Now we are using just one sink and only one log file (for both - client and engine log messages).
The log file will simply just be named "Hazel.log" (for now).

I changed the names of the loggers to "ENGINE" and "CLIENT" and added "%L" to the log patterns. I did that to to make it look more straight inside the log file...

ENGINE I: This is an info.
CLIENT C: This is critical!

is more readable than:

HAZEL: This is an info.
APP: This is critical!

or:

HAZEL info: This is an info.
APP critical: This is critical!

The output to the console stays the same as before.

Hazel/src/Hazel/Core/Log.cpp
HerrDingenz: i already put that back yesterday ;)

HerrDingenz:
trying to optimize it as much as i can .. this is my last commit to this PR.

if you're not happy with it then just close it

CleverSource:
I think it should be formatted ``[Severity] Level: Text`` so it could be ``[Critical] Hazel: Some critical text here.``

HerrDingenz:
Done.

HerrDingenz:
I added "*.log" to the ".gitignore" file to make sure the log file doesn't get pushed to the repository.

Hazel/src/Hazel/Core/Log.cpp
```diff
@@ -10,12 +11,22 @@ namespace Hazel {
...
+
+		s_CoreLogger = std::make_shared<spdlog::logger>("HAZEL", begin(logSinks), end(logSinks));
+		spdlog::register_logger(s_CoreLogger);
 		s_CoreLogger->set_level(spdlog::level::trace);
 
-		s_ClientLogger = spdlog::stdout_color_mt("APP");
+		s_ClientLogger = std::make_shared<spdlog::logger>("APP", begin(logSinks), end(logSinks));
+		spdlog::register_logger(s_ClientLogger);
 		s_ClientLogger->set_level(spdlog::level::trace);
+
```
LovelySanta: Remove this last empty line on the function

.gitignore
```diff
@@ -2,6 +2,7 @@
 **/bin/
 !vendor/bin/premake/
 bin-int/
+*.log
```
LovelySanta: Maybe make a new group (aka a comment line above) for hazel files?
```.gitignore
# Hazel files
*.log
```

Hazel/src/Hazel/Core/Log.cpp
HerrDingenz: sure

.gitignore
HerrDingenz: done

HerrDingenz:
well then i guess we are ready to merge :wink:

TheCherno:
Thank you to everyone for their contributions to this, I'd love to merge this in, however, I have a few issues:
- Logged file doesn't seem to contain any timestamps, like the output log does. This would be useful to have
- Flushing the file seems to be tied to the engine shutting down cleanly, which isn't always the case - I would love to have a log file of what happened if the engine crashed, but currently I would just have a blank Hazel.log file

Could we get these issues resolved? I'm open to discussion.

HerrDingenz:
Adding the timestamps back in can be done by just adding them back to the pattern of the file sink. We stripped them out because the log file gets created per session anyway so the timestamps might not be really that necessary in the log file.

For the log file not being filled if the engine crashes that is really an issue. In my first version i did the flushing myself inside the macros everytime a message got put to the file.

HerrDingenz:
Timestamps are added back to the log file.
Flush level for the loggers is set to "trace" - so now spdlog will flush the contents at every message and we will have data inside the log file even if Hazel crashes.

PR #198, opened 2020-02-02
Fixed memory leak in LayerStack

HerrDingenz:
When we pop a layer or an overlay the layer/overlay will get removed from the layer stack but they won't get deleted when the destructor of the layer stack is called (because obviously they are not on the layer stack anymore).

#### Proposed fix 
Delete the layer/overlay inside PopLayer/PopOverlay functions.

#### Additional context
I talked with @TheCherno about this issue back in october 2019 already and here is what he said:

"Cherno 15.10.2019
Just took a look at the layer stack issue... yes, delete should definitely be called from LayerStack::PopLayer
And PopOverlay"

### DIFF
--- a/Hazel/src/Hazel/Core/LayerStack.cpp
+++ b/Hazel/src/Hazel/Core/LayerStack.cpp
@@ -30,6 +30,7 @@ namespace Hazel {
 		{
 			layer->OnDetach();
 			m_Layers.erase(it);
+			delete layer;
 			m_LayerInsertIndex--;
 		}
 	}
@@ -40,6 +41,7 @@ namespace Hazel {
 		if (it != m_Layers.end())
 		{
 			overlay->OnDetach();
+			delete overlay;
 			m_Layers.erase(it);
 		}
 	}


HerrDingenz:
Another approach might be to put the "popped" layers and overlays onto another stack and clean that up on destruction of the layer stack. This way we could keep them in memory in case they are just turned off for a little while and are being reused later again. But that might complicate things too much. And I like this solution here :wink:

HerrDingenz:
But right now we are pushing the layers to the layerstack like this:

		PushLayer(new Sandbox2D());

in the constructor of the SandboxApp.

So we don't have a pointer to that layer at all, because we're not storing it. In this case the LayerStack should handle deletion of the pointer, because we give ownership of that layer to the layerstack - as cherno stated in his video about the layerstack. :thinking:

LovelySanta:
When you push it, you transfer ownership to the layerstack, and the layerstack will delete it when the layerstack itself gets deleted (in deconstructor).
When you pop it, you get the ownership back and you're responsible of deleting it. If we know we'll never pop it, there is no need to hold a pointer to the layer, and we can pass it in directly. If we do know we'll be popping it, then we should hold a reference to, otherwise we can't pop it to begin with.

Issue #200, opened 2020-02-07
ImGUI doesn't swallow input events


HerrDingenz:
#### Describe the bug
ImGUI doesn't swallow input events and forwards them to underlaying layers.

If you have an ImGUI window open inside which you can scroll up and down the mouse scroll event gets forwarded to our current render layer and changes camera zoom level.
If you type in an ImGUI text box the keys Q, W, E, A, S, D get used by the OrthographicCameraController to move the camera around.

#### To Reproduce
1. set the ImGUIDemoWindow to show
2. uncollapse all the trees until you get a scroll bar inside the ImGUIDemo window
3. Scroll your mouse inside that window
4. Camera zoom level changes, too

1. Open a Textbox input inside an ImGUIWindow and type in there (you can do that inside the ImGUIDemo window)
2. If you type the letters Q, W, E, A, S, D the camera will be moved around or change its rotation

#### Expected behavior
The camera should not react at all to input to an ImGUI window.

PR #201, opened 2020-02-07
fixing ImGUI not swallowing mouse & key input

HerrDingenz:
this fixes the mouse scrolling bug inside issue #200 but doesn't close the issue.

i've been playing with around with fixing the keyinput issue but didn't come to any solution yet.

The problem with the key input issue lies within our OrthographicCameraController which checks for pressed keys inside its OnUpdate() function.

I tried solving that by moving the key input checks from there to an OnKeyPressed() function. 
That works nice so far and ImGUI swallows those inputs, too. But new issues come up then:

- as long as an ImGUI window is active (focused) it will swallow key inputs and they won't get forwarded to the camera - even if you don't type into a textbox
- we don't have access to timestep inside the OnKeyPressed() function .. thus the camera movement looks very clunky and not as smooth as we do it right now inside the OnUpdate() function of the OrthographicCameraController.

I am open for hints and tipps on how to solve the key input issue.

### DIFF
--- a/Hazel/src/Hazel/Core/Application.cpp
+++ b/Hazel/src/Hazel/Core/Application.cpp
@@ -9,6 +9,8 @@
 
 #include <glfw/glfw3.h>
 
+#include <imgui.h>
+
 namespace Hazel {
 
 	Application* Application::s_Instance = nullptr;
@@ -58,6 +60,12 @@ namespace Hazel {
 		EventDispatcher dispatcher(e);
 		dispatcher.Dispatch<WindowCloseEvent>(HZ_BIND_EVENT_FN(Application::OnWindowClose));
 		dispatcher.Dispatch<WindowResizeEvent>(HZ_BIND_EVENT_FN(Application::OnWindowResize));
+
+		ImGuiIO& io = ImGui::GetIO();
+		if (io.WantCaptureMouse)
+		{
+			e.Handled = true;
+		}
 
 		for (auto it = m_LayerStack.rbegin(); it != m_LayerStack.rend(); ++it)
 		{
--- a/Hazel/src/Hazel/ImGui/ImGuiLayer.cpp
+++ b/Hazel/src/Hazel/ImGui/ImGuiLayer.cpp
@@ -69,6 +69,10 @@ namespace Hazel {
 		ImGui_ImplOpenGL3_NewFrame();
 		ImGui_ImplGlfw_NewFrame();
 		ImGui::NewFrame();
+
+		// this is just added to test the behaviour
+		bool showDemo = true;
+		ImGui::ShowDemoWindow(&showDemo);
 	}
 
 	void ImGuiLayer::End()
--- a/Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
+++ b/Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
@@ -4,6 +4,8 @@
 #include "Hazel/Core/Input.h"
 #include "Hazel/Core/KeyCodes.h"
 
+#include <imgui.h>
+
 namespace Hazel {
 
 	OrthographicCameraController::OrthographicCameraController(float aspectRatio, bool rotation)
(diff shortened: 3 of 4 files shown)


HerrDingenz:
This now fixes the key input issue, too --- but i am not really happy with the result and the approach to that fix.

There is the issue now that we have to unfocus any ImGUI window to pass key input through the layers to the CameraController.

ThomasJowett:
Would it not be better to put the io.WantCaptureMouse check in the ImGuiLayer OnEvent(Event& event)?
Then you would not have to include imGui in the application, or if imGui had not been initialized does not cause an error

LovelySanta:
I already had a private conversation with HerrDingenz about this PR. I am not merging this since I really do no like the way the input polling is handling the capturing, and we're still looking for a good way for that. For the same reasons I don't like imgui to be included in the camera controller (or in any place we want to use the input polling).

I do like the proposal to have the ImGuiLayer to set handled = true instead of the application. There is also a bug with the current implementation as it will set any event to `handled = true` when it wants to capture, instead of only the mouse and keyboard events.

I can suggest multiple ways of solving this. The cleanest one would be an event dispatcher where we can have a lamda function as callback function to return the capture status such that `handled = io.WantCaptureMouse` for the relevant events.
The alternative option is to check for EventCategory::EventCategoryInput and return `handled = EventCategory::EventCategoryInput ? io.WantCaptureMouse : false`.

From these two options, I think the first one (with the dispatcher) will have the best overal performance as it will just not run anything when the event type is not one we dispatch, versus the alternative option where it will always have a check, that will run for any event?

Any thoughts on this? And any suggestions for the input polling?

ilikenoodles2:
I would suggest not even using input polling for the CameraController, instead, hooking it up to the event system, and tracking the key states with something like this:
`enum KeyState : uint8_t
{
	A = BIT(0),
	D = BIT(1),
	W = BIT(2),
	S = BIT(3),
	Q = BIT(4),
	E = BIT(5)
};

uint8_t m_KeyState = 0;`
Therefore you can just replace this
`if(Input::IsKeyPressed(HZ_KEY_A))`
with
`if(m_KeyState & KeyState::A)`
Additionally, Imgui capturing doesn't need to be in the application, instead you could override Layer's OnEvent function:
`void ImGuiLayer::OnEvent(Event& e)
{
	ImGuiIO& io = ImGui::GetIO();
	auto flags = e.GetCategoryFlags();

	e.Handled = (flags & EventCategoryMouse) && io.WantCaptureMouse;
	e.Handled |= (flags & EventCategoryKeyboard) && io.WantCaptureKeyboard;
}`

LovelySanta:
Ok, so what about this?

1. The `ImGuiLayer` is the top layer anyway, so this layer can set the event as handled, as @ilikenoodles2 posted:
```cpp
void ImGuiLayer::OnEvent(Event& e)
{
	ImGuiIO& io = ImGui::GetIO();
	auto flags = e.GetCategoryFlags();

	e.Handled |= (flags & EventCategoryMouse) && io.WantCaptureMouse;
	e.Handled |= (flags & EventCategoryKeyboard) && io.WantCaptureKeyboard;
}
```

2. Then the controller can check if the event is already handled, and if this is the case, it can ignore it:
```cpp
bool OrthographicCameraController::OnMouseScrolled(MouseScrolledEvent& e)
{
	HZ_PROFILE_FUNCTION();

	if (e.Handled)
		return true;

	m_ZoomLevel -= e.GetYOffset() * 0.25f;
	m_ZoomLevel = std::max(m_ZoomLevel, 0.25f);
	m_Camera.SetProjection(-m_AspectRatio * m_ZoomLevel, m_AspectRatio * m_ZoomLevel, -m_ZoomLevel, m_ZoomLevel);
	return false;
}
```
```cpp
bool OrthographicCameraController::OnWindowResized(WindowResizeEvent& e)
{
	HZ_PROFILE_FUNCTION();

	if (e.Handled)
		return true;

	m_AspectRatio = (float)e.GetWidth() / (float)e.GetHeight();
	m_Camera.SetProjection(-m_AspectRatio * m_ZoomLevel, m_AspectRatio * m_ZoomLevel, -m_ZoomLevel, m_ZoomLevel);
	return false;
}
```
3. About the input polling in the camera controller, storing the state of the keys should be part of the input class instead of everything that wants to use it. For now we used the glfw polling, but this may change in the future? Not sure how efficient it is, we will see in the future once we start using it more often I guess?

ilikenoodles2:
@LovelySanta I forgot there is already a function implemented for this, OnEvent can look like this:
```
void ImGuiLayer::OnEvent(Event& e)
	{
		ImGuiIO& io = ImGui::GetIO();
		
		e.Handled |= e.IsInCategory(EventCategoryMouse) && io.WantCaptureMouse;
		e.Handled |= e.IsInCategory(EventCategoryKeyboard) && io.WantCaptureKeyboard;
	}
```
Also, you don't have to check if an event is handled, because once it is, nothing else can touch it.

Regarding input polling, I agree storing the key state isn't pretty, but it seems like there isn't any other way to do it, because Camera Controller needs to act like a layer.  You could tie ImGui and input polling together, which is one way to do it, but probably not the best way.  Any suggestions?

HerrDingenz:
I agree - we should do the setting of e.Handled inside ImGui layer.
And maybe we should push the cameracontroller code that uses input polling in OrthographicCameraController::Update to OrthographicCameraController::OnEvent and handle the camera controls based on events instead of input polling.

LovelySanta:
> Also, you don't have to check if an event is handled, because once it is, nothing else can touch it.

Oh yes, forgot about that, however, my thoughts still were still on the handler as 'not working'. This is semi-fixed, but this PR could also fix another issue about the events:

https://github.com/TheCherno/Hazel/blob/3d5068b9bd665bc0030de4c6bb5b9745a2dd3171/Hazel/src/Hazel/Core/Application.cpp#L54-L68

As of now, `Application::OnWindowClose(WindowCloseEvent& e)` will always return true, so if you then look at the layer stack, it will always propagate to the first layer (ImGuiLayer in this case)  before breaking out of it. A simple solution would be to first check if it is handled, and breaking before passing it through that layer:

```cpp
 	for (auto it = m_LayerStack.rbegin(); it != m_LayerStack.rend(); ++it) 
 	{ 
 		if (e.Handled) 
 			break;
 		else
 			(*it)->OnEvent(e); 
 	} 
```

When this is fixed, the layers will work correctly and such we can rely on handled events to not pass through, such that we only have to check if ImGui wants to capture it.

```cpp
void ImGuiLayer::OnEvent(Event& e)
{
	ImGuiIO& io = ImGui::GetIO();
	e.Handled |= e.IsInCategory(EventCategoryMouse) && io.WantCaptureMouse;
	e.Handled |= e.IsInCategory(EventCategoryKeyboard) && io.WantCaptureKeyboard;
}
```

> Regarding input polling, I agree storing the key state isn't pretty, but it seems like there isn't any other way to do it, because Camera Controller needs to act like a layer. You could tie ImGui and input polling together, which is one way to do it, but probably not the best way. Any suggestions?

The camera controller is part of a layer, and we manually call the controller's onEvent in the layer that has ownership over the controller:

https://github.com/TheCherno/Hazel/blob/3d5068b9bd665bc0030de4c6bb5b9745a2dd3171/Sandbox/src/Sandbox2D.cpp#L82-L85

I would assume that we would be able to rebind the keys of the controller at some point, so I would not hardcode the controls to begin with. In any case, as of now, it would be up to Sandbox2D to decide if it is handled or not. I would really leave this out as of now, maybe wait till we have an ECS that maybe might have a camera component that handles the controls for this eventually? To be fair, that also seems a bit out of the scope of this PR at this point and maybe even in the future.

How do you feel about only implementing the parts mentioned in this post then? This is more than enough to close/merge this PR.

HerrDingenz:
> As of now, `Application::OnWindowClose(WindowCloseEvent& e)` will always return true, so if you then look at the layer stack, it will always propagate to the first layer (ImGuiLayer in this case) before breaking out of it

Returning true from any event will set e.Handled inside the event to true so it won't be handled furthermore.

ilikenoodles2:
My bad, I shouldn't have labelled the states by key, I really meant something like this
```
enum MoveState : uint8_t
{
	Up = BIT(0),
	Down = BIT(1),
	Left = BIT(2),
	Right = BIT(3),
	Clockwise = BIT(4),
	CClockwise = BIT(5)
};
```
Because like @HerrDingenz said, the movement should be done in OnUpdate in order to be smooth, OnEvent would set the movestates.

HerrDingenz:
this might be the proper solution.

we have both a OnKeyPressed and OnKeyReleased function that sets the state when a key is pressed or released and the rest is handled by the OnUpdate function.

unfortunately - during my last cleanup of my github - i deleted the fork and we can't merge from there anymore.
so maybe @ilikenoodles2 can create a new PR with his suggestions and we close this one.

LovelySanta:
To be fair, those are already 2 different things, and should use two different PRs..
1. A PR to close this one and the related issue about ImGui swallowing the inputs
2. A new PR for the camera movement

However, I still don't see how saving the states inside the camera movement is different than using the input polling. During a frame you'll get the events of a button press, and you'll save the state, only the next frame you'll start updating the translation/rotation depending on the saved state, how is this different than polling of a key is pressed inside the input class itself?

As of now, the camera already updates inside the `OnUpdate` loop, using the timestep. The only discussion we should be having here is if we rather have the movement keys still propagate to other layers or not? I don't see why we should need to save our own data so we can poll our own saved data if a key is pressed over the current implementation that is polling the saved data in glfw that the Input class is polling already? The only reason I hesitate at this point is that this seems to be duplicating an existing behaviour elsewhere? Please correct me if I'm wrong.

ilikenoodles2:
@LovelySanta, I think we are misunderstanding each other.  Storing the movement state is not a duplication of data.  Just because GLFW says the 'w' key is pressed, it doesn't mean the camera will move up.  This is because of Imgui, or any layer for that matter, handling the keypress. Therefore the only other options are modifying input polling, or something else no one has thought of yet.
> The only discussion we should be having here is if we rather have the movement keys still propagate to other layers or not?
I'm confused by what you mean here, I thought the layer propagation was fine?

LovelySanta:
> Storing the movement state is not a duplication of data. Just because GLFW says the 'w' key is pressed, it doesn't mean the camera will move up. This is because of Imgui, or any layer for that matter, handling the keypress. Therefore the only other options are modifying input polling, or something else no one has thought of yet.

So what I understand you are trying to tell with this is as follows.

You want to store how the camera is moving inside the controller, so for example by pressing down 'W' (event), you store (in the controller) that the camera is moving upwards, when releasing 'W' again (event), you say it is not moving up anymore (so standing still again, if no left/right movement). Then in the update loop, you check, the camera is moving up, then move it by some amount that is defined by the speed and the deltatime.

How is this different that what is happening now in the update loop? Now in the update loop we're asking, is the key 'W' pressed down? If it is, it means the camera is moving up, so we move the camera a bit, defined by the speed and the deltatime.

If I'm wrong, maybe instead of keeping going back and forth, maybe an implementation (PR?) could clear things up?

ilikenoodles2:
That would be a big pull request, would you like me to split them into 2, then we can discuss further?

LovelySanta:
@HerrDingenz did make a branch already, and after some discussion and looking at how the Hazel-dev is looking, it would be better to leave it as is. If you look in the videos where he showed the 3D renderer, you'll see that we'll be rendering to ImGui directly in Hazelnut. And that will solve all our issues that we're having now.

The only thing I consider adding is the part where ImGui swallows the event when it wants to capture them, which should have it's own PR as mentioned multiple times already above. It seems that this also has some issues that, without clicking outside an imgui window, imgui wants to keep capturing the keyboard, but we can tackle that issue once the PR has been made.

ilikenoodles2:
Sound good

LovelySanta:
As mentioned before https://github.com/TheCherno/Hazel/pull/201#issuecomment-616880700, closing this as we have 219 as replacement of what we do want to merge.

PR #204, opened 2020-02-07
added imgui.ini to .gitignore

### LINKED ISSUE #202: imgui.ini should not get pushed to the repository
As the title says: the imgui.ini file should be pushed to the repository since it is changed from user to user and everytime you move ImGUI windows around anyway.

We should add "imgui.ini" to the .gitignore file

HerrDingenz:
Added "imgui.ini" to ".gitignore" ... Closes #202

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,9 @@
 !vendor/bin/premake/
 bin-int/
 
+# ImGUI files
+imgui.ini
+
 # Hazel files
 *.log
 


LovelySanta:
We do want an imgui.ini file committed, we just don't update it, as cherno mensioned for example on discord:

image

HerrDingenz:
But this can lead to several problems, like:

What if somebody who commits changes to the repository and has his ImGUI windows moved over to his 2nd screen and somebody with just one screen clones it? then the ImGUI windows will be shown outside his screen and will be "unreachable" unless he deletes the imgui.ini file :thinking:

LovelySanta:
> What if somebody who commits changes to the repository and has his ImGUI windows moved over to his 2nd screen ...

We check the files before merging a PR in. Even then, it's easy to revert. We can also add our own ini files to imgui, but we're just letting imgui handle it for now.

PR #206, opened 2020-02-16
Exposed Window Props to the sandbox application

ThomasJowett:
Application could not change the name of the window without changing the default window Props.

#### Proposed fix

Expose the window props through the application constructor

### DIFF
--- a/Hazel/src/Hazel/Core/Application.cpp
+++ b/Hazel/src/Hazel/Core/Application.cpp
@@ -13,13 +13,13 @@ namespace Hazel {
 
 	Application* Application::s_Instance = nullptr;
 
-	Application::Application()
+	Application::Application(const WindowProps& props)
 	{
 		HZ_PROFILE_FUNCTION();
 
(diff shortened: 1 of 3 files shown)


PhilipKatis:
I think it would also be a good idea to set the default WindowProperties to the Application constructor.

WhoseTheNerd:
If we want to rename the title then exposing a function would be better.
Something like this:
`SetTitle(const std::string& title)`

ThomasJowett:
It would be a good idea to have a set title and set any other window props as well, but i think it would be best to initialize the application with the correct title to begin with rather than a default and then change it immediately after.

LovelySanta:
Cherno did allow you to set the title when creating an application in https://github.com/TheCherno/Hazel/commit/9ddc33488e50da9550023077227a4b973867c1c2. It doesn't make sense to allow you to create custom sized windows yet, as that depends on the monitor(s) it will be displayed on. For this we need some sort of System handler first. Since the part that is logical is now implemented, I'll close this for now. Once we have some notion of the system this is running on, we can revise the current implementation.

PR #207, opened 2020-02-19
Fix minor issues in OpenGL buffer creation

### LINKED ISSUE #184: Indexbuffer is using wrong bufferdata type
I noticed that the Indexbuffer is using ```GL_ARRAY_BUFFER``` instead of ```GL_ELEMENT_ARRAY_BUFFER``` at ```glBindBuffer(GL_ARRAY_BUFFER, m_RendererID);```

MrLever:
In OpenGLBuffer.cpp I noticed the following issues:
 - The vertex buffer reports it's size in bytes as the vertex count, and not count * sizeof(float)
 - ~~The constructor of OpenGLIndexBuffer used GL_ARRAY_BUFFER instead of GL_ELEMENT_ARRAY_BUFFER like it does in the same object's Bind() and Unbind() functions.~~

#### PR impact 
Issues this solves: closes #184 

#### Proposed fix
- [ ] Change OpenGLVertexBuffer to report the correct buffer size
- [ ] ~~Change OpenGLIndexBuffer to bind the correct buffer type in its constructor.~~ *Not bug

### DIFF
--- a/Hazel/src/Platform/OpenGL/OpenGLBuffer.cpp
+++ b/Hazel/src/Platform/OpenGL/OpenGLBuffer.cpp
@@ -49,8 +49,11 @@ namespace Hazel {
 		HZ_PROFILE_FUNCTION();
 
 		glCreateBuffers(1, &m_RendererID);
-		glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, m_RendererID);
-		glBufferData(GL_ELEMENT_ARRAY_BUFFER, count * sizeof(uint32_t), indices, GL_STATIC_DRAW);
+		
+		// GL_ELEMENT_ARRAY_BUFFER is not valid without an actively bound VAO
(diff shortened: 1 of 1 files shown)


LovelySanta:
The second bullet point is not a bug, as described in #95, and we have an open issue #184 to add a comment to it, to let people know this is how it should be.

MrLever:
> The second bullet point is not a bug, as described in #95, and we have an open issue #184 to add a comment to it, to let people know this is how it should be.

Reverted the change and added the comment mentioned in #184.

MrLever:
Oh, it seems that this was intentional. According to the video the argument to the VertexBuffer (size) is supposed to be the size in bytes.

LovelySanta:
Oh yes, I forgot (again) about that! As you see, the index buffer is taking an argument 'count', which is the number of elements, where that the vertex buffer is taking in the size (in bytes)! Can you edit it, we still can merge in the comment ;)

MrLever:
Fixed, only change is now the comment

PR #210, opened 2020-04-10
Renderer2D: Replace vertex setup code with a for loop.

8infy:
#### Describe the issue
Renderer2D had a lot of code duplication, which could be trivially replaced with a for loop.

#### PR impact
A lot less duplicated code.

#### Proposed fix
Replace with loops where possible.

### DIFF
--- a/Hazel/src/Hazel/Renderer/Renderer2D.cpp
+++ b/Hazel/src/Hazel/Renderer/Renderer2D.cpp
@@ -153,33 +153,25 @@ namespace Hazel {
 		glm::mat4 transform = glm::translate(glm::mat4(1.0f), position)
 			* glm::scale(glm::mat4(1.0f), { size.x, size.y, 1.0f });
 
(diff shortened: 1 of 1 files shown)


IshoAntar:
There is still some duplication though.
The code that you added is copy-pasted 4 times for the different overloads.
wouldn't it be better to just declare a function (static/internal) to setup the buffer?
something like:
```c++
static void SetupQuadVertexBuffer(Ref<VertexBuffer> buffer)
{
    ...
}
```
and call the function instead.

8infy:
> There is still some duplication though.
> The code that you added is copy-pasted 4 times for the different overloads.
> wouldn't it be better to just declare a function (static/internal) to setup the buffer?
> something like:
> 
> ```c++
> static void SetupQuadVertexBuffer(Ref<VertexBuffer> buffer)
> {
>     ...
> }
> ```
> 
> and call the function instead.

It would be, but it feels like too big of a rework and something that cherno would want to design himself.

PR #212, opened 2020-04-12
General Cleanup

ardarsha:
Cleanup for multiple thing:
- GLFW_INCLUDE_NONE was defined only for Windows, though it isn't a Windows exclusive issue.
=> Moved to global defines.
- Removed obsolete `HZ_BUILD_DLL` define
- In 'OpenGLContext.cpp' the header file `GL/GL.h` isn't used, but yet is included. - Removed.
- Includes should match the case of files included. In this cast the GLFW directory is all uppercase.
=> Modified to fit directory name.
- __debugbreak is an Windows exclusive function.
=> Wrapped it in `HZ_DEBUGBREAK` macro with an appropriate ifdef statement for windows and added the linux equivalent.
=> `HZ_ASSERT` is now platform agnostic.

EDIT by @LovelySanta: Making PR description up to date

### DIFF
--- a/Hazel/src/Hazel/Core/Core.h
+++ b/Hazel/src/Hazel/Core/Core.h
@@ -50,6 +50,8 @@
 		#error "Platform doesn't support debugbreak yet!"
 	#endif
 	#define HZ_ENABLE_ASSERTS
+#else
+	#define HZ_DEBUGBREAK()
 #endif
 
 #ifdef HZ_ENABLE_ASSERTS
--- a/Hazel/src/Hazel/Core/Core.h
+++ b/Hazel/src/Hazel/Core/Core.h
@@ -48,8 +48,12 @@
 #endif
 
 #ifdef HZ_ENABLE_ASSERTS
-	#define HZ_ASSERT(x, ...) { if(!(x)) { HZ_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
-	#define HZ_CORE_ASSERT(x, ...) { if(!(x)) { HZ_CORE_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
+	#ifdef HZ_PLATFORM_WINDWOS
+		#define HZ_ASSERT(x, ...) { if(!(x)) { HZ_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
+		#define HZ_CORE_ASSERT(x, ...) { if(!(x)) { HZ_CORE_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
+	#else
+		#error "Platform doesn't support asserts yet!"
+	#endif
 #else
 	#define HZ_ASSERT(x, ...)
 	#define HZ_CORE_ASSERT(x, ...)
--- a/Hazel/src/Hazel/Core/Core.h
+++ b/Hazel/src/Hazel/Core/Core.h
@@ -44,8 +44,11 @@
 #endif // End of platform detection
 
 #ifdef HZ_DEBUG
-	#ifdef HZ_PLATFORM_WINDOWS
+	#if defined(HZ_PLATFORM_WINDOWS)
 		#define HZ_DEBUGBREAK() __debugbreak()
+	#elif defined(HZ_PLATFORM_LINUX)
+		#include <signal.h>
+		#define HZ_DEBUGBREAK() raise(SIGTRAP)
 	#else
 		#error "Platform doesn't support debugbreak yet!"
 	#endif
--- a/Hazel/src/Hazel/Core/Core.h
+++ b/Hazel/src/Hazel/Core/Core.h
@@ -44,16 +44,17 @@
 #endif // End of platform detection
 
 #ifdef HZ_DEBUG
+	#ifdef HZ_PLATFORM_WINDOWS
+		#define HZ_DEBUGBREAK() __debugbreak()
+	#else
+		#error "Platform doesn't support debugbreak yet!"
+	#endif
 	#define HZ_ENABLE_ASSERTS
 #endif
 
 #ifdef HZ_ENABLE_ASSERTS
-	#ifdef HZ_PLATFORM_WINDWOS
-		#define HZ_ASSERT(x, ...) { if(!(x)) { HZ_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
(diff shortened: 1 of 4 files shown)


8infy:
I believe some of theses things have already been fixed in #82

premake5.lua
```diff
@@ -83,8 +84,7 @@ project "Hazel"
 
 		defines
 		{
-			"HZ_BUILD_DLL",
-			"GLFW_INCLUDE_NONE"
+			"HZ_BUILD_DLL"
```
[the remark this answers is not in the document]
ardarsha: Makes sense to me. Didn't realize it was still there.

Hazel/src/Hazel/Core/Core.h
```diff
@@ -48,8 +48,12 @@
...
 
 #ifdef HZ_ENABLE_ASSERTS
-	#define HZ_ASSERT(x, ...) { if(!(x)) { HZ_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
-	#define HZ_CORE_ASSERT(x, ...) { if(!(x)) { HZ_CORE_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
+	#ifdef HZ_PLATFORM_WINDWOS
+		#define HZ_ASSERT(x, ...) { if(!(x)) { HZ_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
+		#define HZ_CORE_ASSERT(x, ...) { if(!(x)) { HZ_CORE_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
+	#else
+		#error "Platform doesn't support asserts yet!"
+	#endif
```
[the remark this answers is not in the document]
ardarsha: That is indeed a cleaner way of doing things.

Hazel/src/Hazel/Core/Core.h
```diff
@@ -44,12 +44,17 @@
 #endif // End of platform detection
 
 #ifdef HZ_DEBUG
+	#ifdef HZ_PLATFORM_WINDOWS
+		#define HZ_DEBUGBREAK() __debugbreak()
+	#else
+		#error "Platform doesn't support debugbreak yet!"
+	#endif
```
LovelySanta, line 51: I would already implement the linux equivalent since quite a few are on linux, and it is not the only macro that is already implemented for linux.

```suggestion
 	#if defined HZ_PLATFORM_WINDOWS 
 		#define HZ_DEBUGBREAK() __debugbreak() 
 	#elif defined HZ_PLATFORM_LINUX 
 		#include <signal.h> 
 		#define HZ_DEBUGBREAK() raise(SIGTRAP)
	#else
		#error "Platform doesn't support debugbreak yet!"
	#endif
```

Hazel/src/Hazel/Core/Core.h
```diff
@@ -44,12 +44,17 @@
...
 	#define HZ_ENABLE_ASSERTS
 #endif
 
 #ifdef HZ_ENABLE_ASSERTS
-	#define HZ_ASSERT(x, ...) { if(!(x)) { HZ_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
-	#define HZ_CORE_ASSERT(x, ...) { if(!(x)) { HZ_CORE_ERROR("Assertion Failed: {0}", __VA_ARGS__); __debugbreak(); } }
+		#define HZ_ASSERT(x, ...) { if(!(x)) { HZ_ERROR("Assertion Failed: {0}", __VA_ARGS__); HZ_DEBUGBREAK(); } }
+		#define HZ_CORE_ASSERT(x, ...) { if(!(x)) { HZ_CORE_ERROR("Assertion Failed: {0}", __VA_ARGS__); HZ_DEBUGBREAK(); } }
```
LovelySanta, line 57: Remove the extra tab indentation that you added.
```suggestion
	#define HZ_ASSERT(x, ...) { if(!(x)) { HZ_ERROR("Assertion Failed: {0}", __VA_ARGS__); HZ_DEBUGBREAK(); } }
	#define HZ_CORE_ASSERT(x, ...) { if(!(x)) { HZ_CORE_ERROR("Assertion Failed: {0}", __VA_ARGS__); HZ_DEBUGBREAK(); } }
```

Hazel/src/Hazel/Core/Core.h
ardarsha: The only concern I might have is the inconsistency with regards to brackets as used here:
https://github.com/TheCherno/Hazel/blob/eb3f705b6097d81d8c743d19ecb239d4dfb94c0c/Hazel/src/Hazel/Core/Core.h#L35-L44

LovelySanta: if we can, we leave the brackets out of the macro. However for debugbreak, on linux you need an argument between the brackets, so it's not something we can leave out:

```
raise(SIGTRAP)(); // Error when just removing the brackets
raise(); // If brackets are not part of it, it won't break correct, missing argument
```

Also, similar to the asserts, there the brackets are also part of the macro as it has an argument. A check never has brackets, and for function macros, the rule is to always add brackets to the call, but in the macro definition, the brackets can be omitted. So the definition can look inconsistent, in the code base, where it is called, it is always consistent with the brackets always present.

Hazel/src/Hazel/Core/Core.h
```diff
@@ -44,12 +44,17 @@
 #endif // End of platform detection
 
 #ifdef HZ_DEBUG
+	#ifdef HZ_PLATFORM_WINDOWS
+		#define HZ_DEBUGBREAK() __debugbreak()
+	#else
+		#error "Platform doesn't support debugbreak yet!"
+	#endif
 	#define HZ_ENABLE_ASSERTS
 #endif
```
LovelySanta, line 53:
```suggestion
#else
	#define HZ_DEBUGBREAK()
#endif

File: Hazel/src/Hazel/Core/Core.h (hunk 3 above)
Comment (ardarsha) [reply Hazel/src/Hazel/Core/Core.h:51]: I meant something like:
```suggestion
	#if defined(HZ_PLATFORM_WINDOWS)
		#define HZ_DEBUGBREAK() __debugbreak()
	#elif defined(HZ_PLATFORM_LINUX)
		#include <signal.h>
		#define HZ_DEBUGBREAK() raise(SIGTRAP)
	#else
		#error "Platform doesn't support debugbreak yet!"
	#endif
```

LovelySanta: oh yes, that's perfectly fine

PR #219, opened 2020-04-21
Imgui Swallowing

### LINKED ISSUE #200: ImGUI doesn't swallow input events

ilikenoodles2:
#### Describe the issue
ImGui sometimes wants mouse and/or key inputs, when it does,
stop events from reaching the rest of the layers.

#### PR impact

Issues this solves       |  Closes #200 
Other PRs this solves    | #201

#### Proposed fix
Provide an OnEvent function for ImGui, so it can set events to handled when it wants input.

#### Additional context
This is not the complete solution to #201.

### DIFF
--- a/Hazel/src/Hazel/Core/Application.cpp
+++ b/Hazel/src/Hazel/Core/Application.cpp
@@ -63,8 +63,7 @@ namespace Hazel {
 		{
 			if (e.Handled) 
 				break;
-			else
-				(*it)->OnEvent(e);
+			(*it)->OnEvent(e);
 		}
 	}
 
--- a/Hazel/src/Hazel/Core/Application.cpp
+++ b/Hazel/src/Hazel/Core/Application.cpp
@@ -61,9 +61,10 @@ namespace Hazel {
 
 		for (auto it = m_LayerStack.rbegin(); it != m_LayerStack.rend(); ++it)
 		{
-			(*it)->OnEvent(e);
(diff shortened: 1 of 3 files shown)


LovelySanta:
As mentioned in https://github.com/TheCherno/Hazel/pull/201#issuecomment-616785426, we should also fix the `e.Handled` check:

```cpp
 	for (auto it = m_LayerStack.rbegin(); it != m_LayerStack.rend(); ++it) 
 	{ 
 		if (e.Handled) 
 			break;
 		else
 			(*it)->OnEvent(e); 
 	} 
```

PR #223, opened 2020-04-23
Fixed Incorrect Cases in GetComponentCount

### LINKED ISSUE #94: Incorrect cases in GetComponentCount
The [OpenGL specification](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glVertexAttribPointer.xhtml) states that the `size` argument of the `glVertexAttribPointer`  function must be 1, 2, 3, or 4. However: https://github.com/TheCherno/Hazel/blob/3c147806f0b253e4a288b64e65a81948d63f112e/Hazel/src/Hazel/Renderer/Buffer.h#L46-L65

There are cases in which 9 or 16 may be returned, and this method is used to generate the `size` argument as described above in: https://github.com/TheCherno/Hazel/blob/3c147806f0b253e4a288b64e65a81948d63f112e/Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp#L61-L66

muthubro:
#### Description of issue
Basically, the `size` argument of glVertexAttribPointer function can at most be 4, but for mat3 and mat4 attributes, the size is currently being set to 9 and 16 respectively.

#### PR impact

Issues this solves       | Fixes #94

#### Proposed fix
This fix passes the matrix as 3 vec3(or 4 vec4) vectors and uses the glVertexAttribDivisor function so that the matrix is read per instance.

#### Additional context
This was tested by passing a `mat4` translation matrix and a `mat3` scale matrix as attributes in Debug and Release builds on Windows x64. It performs as expected.

### DIFF
--- a/Hazel/src/Hazel/Renderer/Buffer.h
+++ b/Hazel/src/Hazel/Renderer/Buffer.h
@@ -51,8 +51,8 @@ namespace Hazel {
 				case ShaderDataType::Float2:  return 2;
 				case ShaderDataType::Float3:  return 3;
 				case ShaderDataType::Float4:  return 4;
-				case ShaderDataType::Mat3:    return 3;
-				case ShaderDataType::Mat4:    return 4;
+				case ShaderDataType::Mat3:    return 3; // 3* float3
+				case ShaderDataType::Mat4:    return 4; // 4* float4
 				case ShaderDataType::Int:     return 1;
 				case ShaderDataType::Int2:    return 2;
 				case ShaderDataType::Int3:    return 3;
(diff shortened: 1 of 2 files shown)


LovelySanta:
Shouldn't a switch statement look better here, also for when more types would be added?

muthubro:
I guess it could, but since the code only differs for matrix type attributes I just used an if-else.

I did think of adding a flag for matrix types so that we could just check for the flag, but I don't see where we would use anything other than `mat3` or `mat4`.

If you think a switch statement would be better, I would be glad to add it though :smile:

LovelySanta:
Yes plz, make it a switch.

Also, I've been looking into it, I'm not 100% sure about the component count on the matrices. Could be nice to add a comment like `// 3x float3` and similar to the other one?

I suppose it works if you pass in twice a mat after each other that the offset is correct.

LovelySanta:
How does this look to you?

muthubro:
Looks good to me.

I got busy with some work and didn't get time to do this, so thanks for the help 😅

PR #225, opened 2020-04-24
Added Random Utility

BaderEddineOuaich:
Added Utility Folder which can hold useful utils in the future such as Random.h

### DIFF
--- a/Hazel/src/Hazel/Utility/Random.h
+++ b/Hazel/src/Hazel/Utility/Random.h
@@ -1,0 +1,53 @@
+#pragma once
+#include <random>
+
+namespace Hazel {
+
+	class Random
+	{
+	public:
+		/*
+		* Returns a Random Real between min and max
+		*/
+		template<typename T>
+		static inline T Real(T min, T max)
+		{
+			HZ_CORE_ASSERT(min < max, "min is > max");
+			std::uniform_real_distribution<T> dist(min, max);
+			auto& mt = Random::GetEngine();
+			return dist(mt);
+		}
+
+		/*
+		* Returns a Random Integer between min and max
+		*/
+		template<typename T>
+		static inline T Int(T min, T max)
+		{
+			HZ_CORE_ASSERT(min < max, "min is > max");
+			std::uniform_int_distribution<T> dist(min, max);
+			auto& mt = GetEngine();
+			return dist(mt);
+		}
+
+		/*
+		* Returns a Random bool value, either 'true' or 'false'
+		*/
+		static inline bool Bool()
+		{
+			//fifty fifty chance
+			std::bernoulli_distribution dist(0.5);
+			auto& mt = GetEngine();
+			return dist(mt);
+		}
+
+	private:
+		static std::mt19937& Random::GetEngine()
+		{
+			static std::random_device seed_gen;
+			static std::mt19937 engine(seed_gen());
+			return engine;
+		}
+	};
+
+}
--- a/Hazel/src/Hazel.h
+++ b/Hazel/src/Hazel.h
@@ -17,6 +17,8 @@
 
 #include "Hazel/ImGui/ImGuiLayer.h"
 
+#include "Hazel/Utility/Random.h"
+
 // ---Renderer------------------------
 #include "Hazel/Renderer/Renderer.h"
 #include "Hazel/Renderer/Renderer2D.h"


Hazel/src/Hazel/Utility/Random.h
```diff
@@ -0,0 +1,53 @@
...
+namespace Hazel {
+
+	class Random
+	{
+	public:
+		/*
+		* Returns a Random Real between min and max
+		*/
+		template<typename T>
+		static inline T Real(T min, T max)
```
8infy, line 13: `inline` here is redundant

Hazel/src/Hazel/Utility/Random.h
```diff
@@ -0,0 +1,53 @@
...
+			std::uniform_real_distribution<T> dist(min, max);
+			auto& mt = GetEngine();
+			return dist(mt);
+		}
+
+		/*
+		* Returns a Random Integer between min and max
+		*/
+		template<typename T>
+		static inline T Int(T min, T max)
```
8infy, line 25: `inline` here is redundant

Hazel/src/Hazel/Utility/Random.h
```diff
@@ -0,0 +1,53 @@
...
+			HZ_CORE_ASSERT(min < max, "min is > max");
+			std::uniform_int_distribution<T> dist(min, max);
+			auto& mt = GetEngine();
+			return dist(mt);
+		}
+
+		/*
+		* Returns a Random bool value, either 'true' or 'false'
+		*/
+		static inline bool Bool()
```
8infy, line 36: `inline` here is redundant

Hazel/src/Hazel/Utility/Random.h
```diff
@@ -0,0 +1,53 @@
...
+		{
+			//fifty fifty chance
+			std::bernoulli_distribution dist(0.5);
+			auto& mt = GetEngine();
+			return dist(mt);
+		}
+
+	private:
+		static inline std::mt19937& Random::GetEngine()
```
8infy, line 45: `inline` here is redundant

Hazel/src/Hazel/Utility/Random.h
```diff
@@ -0,0 +1,53 @@
...
+		static bool Bool()
+		{
+			//fifty fifty chance
+			std::bernoulli_distribution dist(0.5);
+			auto& mt = GetEngine();
+			return dist(mt);
+		}
+
+	private:
+		static std::mt19937& Random::GetEngine()
```
BaderEddineOuaich, line 45: Unnecessary Class:: and ::function() call

Hazel/src/Hazel/Utility/Random.h
```diff
@@ -0,0 +1,53 @@
...
+	class Random
+	{
+	public:
+		/*
+		* Returns a Random Real between min and max
+		*/
+		template<typename T>
+		static T Real(T min, T max)
+		{
+			HZ_CORE_ASSERT(min < max, "min is > max");
```
WhoseTheNerd, line 15: if `min < max` then `min == max` is also a possibility. Should be more like `"min is >= max"`

BaderEddineOuaich: Oh yep! i miss that one

BaderEddineOuaich:
i was wondering about the Random::Bool() function if we can provide the chance to the client as an optional parameter. will be like Random::Bool(double chance = 0.5); 50% 50% as default.

it can be useful for such an examples like if you want to render 10000 circles, and u want 5% of them to have a specific radius. you can then specify the chance to 0.95 (95% will be true) and work it out.
```c++
for(int i = 0; i < 10000; ++i)
{
   if(Random::Bool(0.95))
      radius = 20;
   else
      radius = 5;
}
```

what do you guys think?

Hazel/src/Hazel/Utility/Random.h
```diff
@@ -0,0 +1,53 @@
...
+	class Random
+	{
+	public:
+		/*
+		* Returns a Random Real between min and max
+		*/
+		template<typename T>
+		static T Real(T min, T max)
+		{
+			HZ_CORE_ASSERT(min < max, "min is >= max");
```
Lecrapouille, line 15: min max can conflict with std::min and std::max. Why not renaming to `low` and `high` ? Like this instead of the assert you can do std::uniform_real_distribution<T> dist(std::min(low, high), std::max(low, high));

Hazel/src/Hazel/Utility/Random.h
```diff
@@ -0,0 +1,53 @@
...
+			std::uniform_real_distribution<T> dist(min, max);
+			auto& mt = GetEngine();
+			return dist(mt);
+		}
+
+		/*
+		* Returns a Random Integer between min and max
+		*/
+		template<typename T>
+		static T Int(T min, T max)
```
Lecrapouille, line 25: Maybe an alternative + integer type only and using the previous method (but again I'm not specialist on random number and template type). I made it to include `high` in random values (instead of excluded).
```
template<typename T>
typename std::enable_if<std::is_integral<T>::value && !std::is_floating_point<T>::value, T>::type
static T Int(T const low = zero<T>(), T const high = one<T>())
{
      return low + static_cast<T>(std::floor(Random::Real<float>() * static_cast<float>(high - low + 1)));
}
```

Hazel/src/Hazel/Utility/Random.h
```diff
@@ -0,0 +1,53 @@
...
+
+		/*
+		* Returns a Random bool value, either 'true' or 'false'
+		*/
+		static bool Bool()
+		{
+			//fifty fifty chance
+			std::bernoulli_distribution dist(0.5);
+			auto& mt = GetEngine();
+			return dist(mt);
```
Lecrapouille, line 41: Maybe return !!dist(mt);

Hazel/src/Hazel/Utility/Random.h
WhoseTheNerd: they will not confict since they are parameters and there is no functions called min() and max() as I'm aware. sd::min and std::max are namespaced functions and are different from just min and max unless you are doing `using namespace std;`, please don't.

Hazel/src/Hazel/Utility/Random.h
salimp2009: if you are going to use it , 
i believe the second check for  **!std::is_floating_point** is unneccessary **std::integral** already covers it [](https://en.cppreference.com/w/cpp/types/is_integral)  

also you can shorten it by by using **std::is_integral_v** ; no need for the extra::type then  and same for enable_if_t . And i believe you do not need the second T if you are already providing the return type and if you use you dont have to specifiy the return type;
one more suggestion maybe move the whole check to template parameters;
```
template<typename T, typename =std::enable_if_t<std::is_integral_v<T>, T>> // the last T is for the return type 
static Int(....
```
or
```
template<typename T, typename =std::enable_if_t<std::is_integral_v<T>>> 
static  T Int(....
```
But might need to check as . Just a humble suggestion if used; those enables can be shortened by using an Alias we can define in the Core like;
**in Core.h define Alias;**
```
template<typename  T>
using EnableIfReal = std::enable_if_t<std::is_integral_v<T>>
```

**and actual code becomes shorter easy on the eyes;**
```
template<typename T, typename =EnableIfReal<T>> 
static  T Int(.....
```

PR #226, opened 2020-04-25
Fixed default value for indexCount in OpenGLRendererAPI::DrawIndexed

muthubro:
Currently, if you pass a value of zero (which is also the default value) to the `indexCount` parameter of `OpenGLRendererAPI::DrawIndexed`, it calls `glDrawElements` with the count parameter set to the total size of the index buffer.
This is an issue with the batch rendering setup. If between BeginScene and EndScene we do not call DrawQuad (for whatever reason), we end up calling DrawIndexed with `indexCount` 0, which instead of setting the count to 0, sets it to the maximum size. This causes artifacts (which I first observed when implementing TheCherno's particle system using batch rendering).

#### PR impact

#### Proposed fix
I just set the count value of `glDrawElements` to be the same as the passed parameter. This does mean that we will not be able to use the default parameter of 0, and so will always have to pass the count parameter.

#### Additional context
Tested successfully in Debug and Release modes on Windows x64.

### DIFF
--- a/Hazel/src/Hazel/Renderer/RenderCommand.h
+++ b/Hazel/src/Hazel/Renderer/RenderCommand.h
@@ -27,7 +27,7 @@ namespace Hazel {
 			s_RendererAPI->Clear();
 		}
 
-		static void DrawIndexed(const Ref<VertexArray>& vertexArray, uint32_t count = 0)
+		static void DrawIndexed(const Ref<VertexArray>& vertexArray, uint32_t count)
 		{
 			s_RendererAPI->DrawIndexed(vertexArray, count);
 		}
--- a/Hazel/src/Hazel/Renderer/RenderCommand.h
+++ b/Hazel/src/Hazel/Renderer/RenderCommand.h
@@ -27,7 +27,7 @@ namespace Hazel {
 			s_RendererAPI->Clear();
 		}
 
-		static void DrawIndexed(const Ref<VertexArray>& vertexArray, uint32_t count)
+		static void DrawIndexed(const Ref<VertexArray>& vertexArray, uint32_t count = 0)
 		{
 			s_RendererAPI->DrawIndexed(vertexArray, count);
 		}
--- a/Hazel/src/Hazel/Renderer/Renderer.cpp
+++ b/Hazel/src/Hazel/Renderer/Renderer.cpp
@@ -40,7 +40,7 @@ namespace Hazel {
 		shader->SetMat4("u_Transform", transform);
 
 		vertexArray->Bind();
-		RenderCommand::DrawIndexed(vertexArray);
+		RenderCommand::DrawIndexed(vertexArray, vertexArray->GetIndexBuffer()->GetCount());
 	}
 
 }
--- a/Hazel/src/Hazel/Renderer/Renderer.cpp
+++ b/Hazel/src/Hazel/Renderer/Renderer.cpp
@@ -40,7 +40,7 @@ namespace Hazel {
 		shader->SetMat4("u_Transform", transform);
 
 		vertexArray->Bind();
-		RenderCommand::DrawIndexed(vertexArray, vertexArray->GetIndexBuffer()->GetCount());
+		RenderCommand::DrawIndexed(vertexArray);
 	}
 
 }
--- a/Hazel/src/Hazel/Renderer/Renderer2D.cpp
+++ b/Hazel/src/Hazel/Renderer/Renderer2D.cpp
@@ -135,12 +135,15 @@ namespace Hazel {
 
 	void Renderer2D::Flush()
 	{
-		// Bind textures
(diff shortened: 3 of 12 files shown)


Puddlestomper:
> I would suggest something like:
> ```c++
> void Renderer2D::Flush() 
> {
> 	if (s_Data.QuadIndexCount > 0)
>  	{
> 		// Bind textures
> 		for (uint32_t i = 0; i < s_Data.TextureSlotIndex; i++)
> 			s_Data.TextureSlots[i]->Bind(i);
> 	
> 		RenderCommand::DrawIndexed(s_Data.QuadVertexArray, s_Data.QuadIndexCount);
> 	}
> 	s_Data.Stats.DrawCalls++;
> }
> ```
> 
> Any thoughts?

I think we should probably increment draw calls inside the if, so that it only increments when we actually do a draw call. Otherwise that looks good to me.

muthubro:
Looks good to me too

LovelySanta:
Seems like something is wrong with the submodules... this PR is not supposed to change them, but they do however show up as changed... Please fix that.

> I think we should probably increment draw calls inside the if, so that it only increments when we actually do a draw call. Otherwise that looks good to me.

My thought was to count at least 1 draw call/scene as we do however not have any metrics on the amount of scenes that we begin/end, which is not performance heavy, so I think it is fine. But I can see how counting the actual draw calls is wanted. Good catch.

Hazel/src/Hazel/Renderer/RenderCommand.h
```diff
@@ -27,7 +27,7 @@ namespace Hazel {
 			s_RendererAPI->Clear();
 		}
 
-		static void DrawIndexed(const Ref<VertexArray>& vertexArray, uint32_t count = 0)
+		static void DrawIndexed(const Ref<VertexArray>& vertexArray, uint32_t count)
```
LovelySanta, line 30:
```suggestion
		static void DrawIndexed(const Ref<VertexArray>& vertexArray, uint32_t count = 0)
```

Hazel/src/Hazel/Renderer/Renderer.cpp
```diff
@@ -40,7 +40,7 @@ namespace Hazel {
 		shader->SetMat4("u_Transform", transform);
 
 		vertexArray->Bind();
-		RenderCommand::DrawIndexed(vertexArray);
+		RenderCommand::DrawIndexed(vertexArray, vertexArray->GetIndexBuffer()->GetCount());
```
LovelySanta, line 43:
```suggestion
		RenderCommand::DrawIndexed(vertexArray);
```

Hazel/src/Hazel/Renderer/RendererAPI.h
```diff
@@ -19,7 +19,7 @@ namespace Hazel {
 		virtual void SetClearColor(const glm::vec4& color) = 0;
 		virtual void Clear() = 0;
 
-		virtual void DrawIndexed(const Ref<VertexArray>& vertexArray, uint32_t indexCount = 0) = 0;
+		virtual void DrawIndexed(const Ref<VertexArray>& vertexArray, uint32_t indexCount) = 0;
```
LovelySanta, line 22:
```suggestion
		virtual void DrawIndexed(const Ref<VertexArray>& vertexArray, uint32_t indexCount = 0) = 0;
```

Hazel/src/Hazel/Renderer/Renderer2D.cpp
```diff
@@ -135,12 +135,15 @@ namespace Hazel {
...
+		{
+			// Bind textures
+			for (uint32_t i = 0; i < s_Data.TextureSlotIndex; i++)
+				s_Data.TextureSlots[i]->Bind(i);
 		
-		RenderCommand::DrawIndexed(s_Data.QuadVertexArray, s_Data.QuadIndexCount);
-		s_Data.Stats.DrawCalls++;
+			RenderCommand::DrawIndexed(s_Data.QuadVertexArray, s_Data.QuadIndexCount);
+			s_Data.Stats.DrawCalls++;
+		}
```
LovelySanta, line 146: Since we decided to put the drawcall increment inside the if statement, it might be better to prematurely return instead? (github is messing up the suggestion, so i'm adding it as a regular code block instead)

```cpp
		if (s_Data.QuadIndexCount == 0)
			return; // Nothing to draw
		
		// Bind textures
		for (uint32_t i = 0; i < s_Data.TextureSlotIndex; i++)
			s_Data.TextureSlots[i]->Bind(i);
		
		RenderCommand::DrawIndexed(s_Data.QuadVertexArray, s_Data.QuadIndexCount);
		s_Data.Stats.DrawCalls++;
```

muthubro:
> Seems like something is wrong with the submodules... this PR is not supposed to change them, but they do however show up as changed... Please fix that.

Sorry about the submodules. I think I overlooked that change when I committed.

PR #230, opened 2020-04-30
delete QuadVertexBufferBase on shutdown

ilikenoodles2:
s_Data.QuadVertexBufferBase is allocated in Init, but isn't deleted on Shutdown.

#### PR impact

#### Proposed fix
Delete s_Data.QuadVertexBufferBase in the Shutdown method.

### DIFF
--- a/Hazel/src/Hazel/Renderer/Renderer2D.cpp
+++ b/Hazel/src/Hazel/Renderer/Renderer2D.cpp
@@ -106,6 +106,8 @@ namespace Hazel {
 	void Renderer2D::Shutdown()
 	{
 		HZ_PROFILE_FUNCTION();
+
+		delete s_Data.QuadVertexBufferBase;
 	}
 
 	void Renderer2D::BeginScene(const OrthographicCamera& camera)
--- a/Hazel/src/Hazel/Renderer/Renderer2D.cpp
+++ b/Hazel/src/Hazel/Renderer/Renderer2D.cpp
@@ -107,7 +107,7 @@ namespace Hazel {
 	{
 		HZ_PROFILE_FUNCTION();
 
-		delete s_Data.QuadVertexBufferBase;
+		delete[] s_Data.QuadVertexBufferBase;
 	}
 
 	void Renderer2D::BeginScene(const OrthographicCamera& camera)


Hazel/src/Hazel/Renderer/Renderer2D.cpp
```diff
@@ -106,6 +106,8 @@ namespace Hazel {
 	void Renderer2D::Shutdown()
 	{
 		HZ_PROFILE_FUNCTION();
+
+		delete[] s_Data.QuadVertexBufferBase;
```
8infy, line 110: Shouldn't `Renderer2DData` delete its own memory instead? 🤔

Hazel/src/Hazel/Renderer/Renderer2D.cpp
```diff
@@ -341,4 +343,4 @@ namespace Hazel {
 		return s_Data.Stats;
 	}
 
-}
```
8infy, line 344: Why did this move?

Hazel/src/Hazel/Renderer/Renderer2D.cpp
IshoAntar: I agree, `Renderer2DData` should be responsible for cleaning-up it's own memory.

ilikenoodles2: Then shouldn't it create it's own resources?

IshoAntar: It should, In that case you can declare a constructor for `Renderer2DData` which initializes everything (technically the whole `Renderer2D::Init` function) and a destructor that frees it, convert the 
static instance (`s_Data`) to a pointer, and `new` it in the `Renderer2D::Init` function and `delete` it in `Renderer2D::Shutdown`. not sure why cherno didn't do it this way, maybe he has a good reason?  not sure, but i think it's better (__RAII__ stuff, you know).

8infy: @IshoAntar I agree, I don't think there's a single reason why it shouldn't be that way.

ilikenoodles2: I tried, and it can't be done with a constructor, because it attempts to create OpenGL resources before OpenGL is initialized.  Perhaps Renderer2DData could have an `Init()` or `Shutdown` equivalent, maybe there is a better way?

ilikenoodles2: What about using an `Init()` and `Destroy()` pair instead of making `s_Data` a pointer, because it seems so tedious.

Puddlestomper: I think the intention behind the `Renderer2DData` struct is just for it to store a bunch of static state data that we need. It doesn't really need its own constructor and destructor as this will just clutter the file. As @IshoAntar mentioned, it's just a container and not a full object. I think it's perfectly fine for us to manually construct and delete this one variable in the Renderer2D methods.

LovelySanta: I completely agree with @Puddlestomper: `Renderer2DData` is just a data struct, no functionality, and it is linked to the `Renderer2D`. The struct itself is static, and it's the `Renderer2D` that is responsible for it, so this is correct, it should be deleted on shutdown, as it is now.

Hazel/src/Hazel/Renderer/Renderer2D.cpp
LovelySanta: These are just file endings... not really something I give much attention to.. Maybe we should see to get #52 integrated, because this is not the first time file endings change unneeded ...

8infy: Yeah, we definitely need #52 integrated.

PR #237, opened 2020-05-04
Linux support, CMake and C++17 implementation.

Neomer:
cmake support
linux support
moving to c++17 standard

Engine tested on Ubuntu 20.04.
Vendor: X.Org
Renderer: AMD Radeon (TM) RX 460 Graphics (POLARIS11, DRM 3.35.0, 5.4.0-28-generic, LLVM 9.0.1)
Version: 4.6 (Compatibility Profile) Mesa 20.2.0-devel (git-f1a40a2 2020-05-03 focal-oibaf-ppa)

Major refactoring.
**const std::string&** changed to **std::string_view**.
**virtual override** definition changed to **override** only definition.
most of constructors marked as explicit.
some methods marked as const.
remove const-reference arguments of std::shared_ptr type.

### DIFF
--- a/Hazel/CMakeLists.txt
+++ b/Hazel/CMakeLists.txt
@@ -1,0 +1,103 @@
+cmake_minimum_required(VERSION 3.16)
+project(Hazel)
+
+set(CMAKE_CXX_STANDARD 17)
+
+add_compile_definitions(HZ_DEBUG)
+
+include_directories(vendor/spdlog/include)
+include_directories(vendor/Glad/include)
+include_directories(vendor/GLFW/include)
+include_directories(vendor/imgui)
+include_directories(vendor/stb_image)
+include_directories(vendor/glm)
+
+add_subdirectory(vendor/glm)
+add_subdirectory(vendor/GLFW)
+
+set(IMGUI_SRC
+        vendor/imgui/imgui.h
+        vendor/imgui/imgui.cpp
+        vendor/imgui/imgui_widgets.cpp
+        vendor/imgui/imgui_demo.cpp
+        vendor/imgui/imgui_draw.cpp)
+
+set(STBIMG_SRC
+        vendor/stb_image/stb_image.cpp)
+
+set(GLAD_SRC
+        vendor/Glad/src/glad.c)
+
+set(SOURCES
+        src/Hazel/Core/Application.cpp
+        src/Hazel/Core/Application.h
+        src/Hazel/Core/Core.h
+        src/Hazel/Core/EntryPoint.h
+        src/Hazel/Core/Input.cpp
+        src/Hazel/Core/Input.h
+        src/Hazel/Core/KeyCodes.h
+        src/Hazel/Core/Layer.cpp
+        src/Hazel/Core/Layer.h
+        src/Hazel/Core/LayerStack.cpp
+        src/Hazel/Core/LayerStack.h
+        src/Hazel/Core/Log.cpp
+        src/Hazel/Core/Log.h
+        src/Hazel/Core/MouseCodes.h
+        src/Hazel/Core/Timestep.h
+        src/Hazel/Core/Window.cpp
+        src/Hazel/Core/Window.h
+        src/Hazel/Debug/Instrumentor.h
+        src/Hazel/Events/ApplicationEvent.h
+        src/Hazel/Events/Event.h
+        src/Hazel/Events/KeyEvent.h
+        src/Hazel/Events/MouseEvent.h
+        src/Hazel/ImGui/ImGuiBuild.cpp
+        src/Hazel/ImGui/ImGuiLayer.cpp
+        src/Hazel/ImGui/ImGuiLayer.h
+        src/Hazel/Renderer/Buffer.cpp
+        src/Hazel/Renderer/Buffer.h
+        src/Hazel/Renderer/GraphicsContext.cpp
+        src/Hazel/Renderer/GraphicsContext.h
+        src/Hazel/Renderer/OrthographicCamera.cpp
+        src/Hazel/Renderer/OrthographicCamera.h
+        src/Hazel/Renderer/OrthographicCameraController.cpp
+        src/Hazel/Renderer/OrthographicCameraController.h
+        src/Hazel/Renderer/RenderCommand.cpp
+        src/Hazel/Renderer/RenderCommand.h
+        src/Hazel/Renderer/Renderer.cpp
+        src/Hazel/Renderer/Renderer.h
+        src/Hazel/Renderer/Renderer2D.cpp
+        src/Hazel/Renderer/Renderer2D.h
+        src/Hazel/Renderer/RendererAPI.cpp
+        src/Hazel/Renderer/RendererAPI.h
+        src/Hazel/Renderer/Shader.cpp
+        src/Hazel/Renderer/Shader.h
+        src/Hazel/Renderer/Texture.cpp
+        src/Hazel/Renderer/Texture.h
+        src/Hazel/Renderer/VertexArray.cpp
+        src/Hazel/Renderer/VertexArray.h
+        src/Platform/OpenGL/OpenGLBuffer.cpp
+        src/Platform/OpenGL/OpenGLBuffer.h
+        src/Platform/OpenGL/OpenGLContext.cpp
+        src/Platform/OpenGL/OpenGLContext.h
+        src/Platform/OpenGL/OpenGLRendererAPI.cpp
+        src/Platform/OpenGL/OpenGLRendererAPI.h
+        src/Platform/OpenGL/OpenGLShader.cpp
+        src/Platform/OpenGL/OpenGLShader.h
+        src/Platform/OpenGL/OpenGLTexture.cpp
+        src/Platform/OpenGL/OpenGLTexture.h
+        src/Platform/OpenGL/OpenGLVertexArray.cpp
+        src/Platform/OpenGL/OpenGLVertexArray.h
+        src/Platform/Windows/WindowsInput.cpp
+        src/Platform/Windows/WindowsInput.h
+        src/Platform/Windows/WindowsWindow.cpp
+        src/Platform/Windows/WindowsWindow.h
+        src/Hazel.h
+        src/hzpch.cpp
+        src/hzpch.h)
+
+include_directories(src)
+
+add_library(${PROJECT_NAME} STATIC ${SOURCES} ${IMGUI_SRC} ${GLAD_SRC} ${STBIMG_SRC})
+
+target_link_libraries(${PROJECT_NAME} glfw)
--- a/Hazel/src/Hazel/Core/Log.h
+++ b/Hazel/src/Hazel/Core/Log.h
@@ -33,4 +33,4 @@ namespace Hazel {
 #define HZ_INFO(...)          ::Hazel::Log::GetClientLogger()->info(__VA_ARGS__)
 #define HZ_WARN(...)          ::Hazel::Log::GetClientLogger()->warn(__VA_ARGS__)
 #define HZ_ERROR(...)         ::Hazel::Log::GetClientLogger()->error(__VA_ARGS__)
-#define HZ_CRITICAL(...)      ::Hazel::Log::GetClientLogger()->critical(__VA_ARGS__)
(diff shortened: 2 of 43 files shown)


LovelySanta:
As shown in the most-up-to-date-but-outdated linux PR, you can see that premake can be used to build on linux, so I do not see the use of cmake since cherno wanted to use premake. More info about using premake for linux is mentioned [in this readme](https://github.com/TheCherno/Hazel/blob/9ce57a3d76ac4a23777106eab50373c6fe09a6ac/README.md).

I'll leave this PR open for now, as it does have some changes we might want to merge in, like marking functions as const and other improvements that are done here.

Hazel/CMakeLists.txt
```diff
@@ -0,0 +1,103 @@
+cmake_minimum_required(VERSION 3.16)
+project(Hazel)
+
+set(CMAKE_CXX_STANDARD 17)
+
+add_compile_definitions(HZ_DEBUG)
```
ReDucTor, line 6: If cmake was going to be used, I would suggest going with modern cmake not this legacy approach.

Hazel/src/Hazel/Core/Log.h
```diff
@@ -33,4 +33,4 @@ namespace Hazel {
 #define HZ_INFO(...)          ::Hazel::Log::GetClientLogger()->info(__VA_ARGS__)
 #define HZ_WARN(...)          ::Hazel::Log::GetClientLogger()->warn(__VA_ARGS__)
 #define HZ_ERROR(...)         ::Hazel::Log::GetClientLogger()->error(__VA_ARGS__)
-#define HZ_CRITICAL(...)      ::Hazel::Log::GetClientLogger()->critical(__VA_ARGS__)
\ No newline at end of file
+#define HZ_CRITICAL(...)      ::Hazel::Log::GetClientLogger()->critical(__VA_ARGS__); throw std::runtime_error("Fatal error.");
```
ReDucTor, line 36: You should avoid adding multiple expressions to a macro, this would break

if (something) HZ_CRITICAL(....)

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
```diff
@@ -11,7 +11,7 @@ namespace Hazel {
 	{
 	}
 
-	void OrthographicCameraController::OnUpdate(Timestep ts)
+	void OrthographicCameraController::OnUpdate(const Timestep& ts)
```
ReDucTor, line 14: Why make these const ref?

Hazel/src/Hazel/Renderer/Shader.cpp
```diff
@@ -30,41 +30,41 @@ namespace Hazel {
...
 		Add(name, shader);
 		return shader;
 	}
 
-	Ref<Shader> ShaderLibrary::Get(const std::string& name)
+	Ref<Shader> ShaderLibrary::Get(std::string_view name)
 	{
 		HZ_CORE_ASSERT(Exists(name), "Shader not found!");
-		return m_Shaders[name];
+		return m_Shaders[name.data()];
```
ReDucTor, line 62: You shouldn't be using `std::string_view::data` as an accessor to a null-terminated byte string, because it isn't.

This would now be broken with this change
```
std::string_view v = "hello world";
shaderLib.Get(v.substr(5));
```

Looking at this review, much of the `string_view` changes also have broken this, not to mention in some cases it is actually adding unnecessary conversions between `std::string` and `std::string_view` which will just create additional allocations and overhead such as here because `unordered_map::operator[]` take the template `Key` type not some other template type (`find` did get this in C++20 but not `operator[]`).

I feel it is reckless to make these changes in bulk without considering all of the use cases, moving straight to `string_view` does not mean you instantly get performance benefits, if performance is a concern with a `string` the solution of going straight to `string_view` is often not the best solution, strings are slow, error prone, etc.

All that moving it to a `string_view` has done is remove the edge case of a string literal constructing a string each time, while making it harder to work many API that expect null-terminated strings (e.g. OpenGL), even a `const char *` is better in many cases then a `string_view`

Hazel/src/Platform/OpenGL/OpenGLVertexArray.h
```diff
@@ -8,16 +8,16 @@ namespace Hazel {
 	{
 	public:
 		OpenGLVertexArray();
-		virtual ~OpenGLVertexArray();
+		~OpenGLVertexArray() override;
 
-		virtual void Bind() const override;
```
ReDucTor, line 13: Why get rid of `virtual`? It is much more visually distinct then just the `override`.

I'm in favor of keeping both `virtual` and `override`, especially in cases where there isn't existing conventions to identify what is part of an implemented interface or not.

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -110,6 +110,9 @@ namespace Hazel {
 					data.EventCallback(event);
 					break;
 				}
+
+                default:
```
ReDucTor, line 114: You should be careful adding `default`, while unfortunately these are macros and not enums adding a `default` makes it much harder to identify when a case statement is missing, some compilers and static analyzers will typically warning if a case statement is missing in a switch based on an enum without a default.

Additionally there is some formatting issues here where this is using spaces and not tabs, like the rest of the code (also a few other places in this PR where that occurs)

Hazel/src/Hazel/Core/Log.h
Neomer: You better should said that nobody should use IF expression without brackets, because it can bring a lot of unexpected bugs in your code.

I suggest rewrite this macro to:
```suggestion
#define HZ_CRITICAL(...)     { ::Hazel::Log::GetClientLogger()->critical(__VA_ARGS__); throw std::runtime_error("Fatal error."); }
```

Hazel/src/Hazel/Renderer/OrthographicCameraController.cpp
Neomer: because it's not necessary to be copied ...

ReDucTor: Why does that matter In this case? Timestep contains a single float that is created on the stack, using this now means the ABI is passing around a pointer equivalent that it must dereference on usage, instead of just a struct decomposed to a float.

While these optimizations are insignificant and unnecessary based on this struct I think it will be the opposite of an optimizations, as it now means the caller needs to store it in an addressable location (not registers staying in registers as could be done) and then has the overhead of loading from the address on it's usage

Hazel/src/Hazel/Core/Log.h
ReDucTor: 'nobody' might be easy when you control what is written but harder with more users, the alternative suggested solution now also means this is not mean to have a trailing semicolon unlike the other macros.

Also this introduces exceptions which the code base does not already use and is not designed to be exception safe.

Hazel/src/Hazel/Events/MouseEvent.h
```diff
@@ -52,11 +52,11 @@ namespace Hazel {
 	class MouseButtonEvent : public Event
 	{
 	public:
-		inline MouseCode GetMouseButton() const { return m_Button; }
+		[[nodiscard]] inline MouseCode GetMouseButton() const { return m_Button; }
```
8infy, line 55: I've said it some many times before, we don't need `inline` if we're inside a class...

Hazel/src/Platform/OpenGL/OpenGLShader.cpp
```diff
@@ -97,7 +97,7 @@ namespace Hazel {
 			size_t eol = source.find_first_of("\r\n", pos); //End of shader type declaration line
 			HZ_CORE_ASSERT(eol != std::string::npos, "Syntax error");
 			size_t begin = pos + typeTokenLength + 1; //Start of shader type name (after "#type " keyword)
-			std::string type = source.substr(begin, eol - begin);
+			std::string_view type = source.substr(begin, eol - begin);
```
8infy, line 100: Could be wrong, but isn't this "viewing" a deallocated substring?

Hazel/src/Hazel/Renderer/Shader.cpp
salimp2009: Using string_view in class member initialization and a couple of other places is not recommended in C++17 book by Nicolai M. Josuttis "C++17 - The Complete Guide" ; he gives summary of safe use of string_views ;

" Summary of Safe Use of String Views
To summarize, use std::string_view with care, meaning that you should adjust your programming style as follows:
• Do not use string views in APIs that pass the argument to a string.
– Do not initialize string members from string_view parameters.
– No string at the end of a string_ view chain.
• Do not return a string view."

class Person {
  std::string name;
public:
Person (std::string_view n)
: name{n} { }
// don’t do this

Author recommended using string as a copy and then std::move() is a better to initialize the
class Person {
     std::string name;
    public:
     Person (std::string n)
      : name{std::move(n)} {
     }
...
};

Hazel/src/Platform/OpenGL/OpenGLShader.cpp
ReDucTor: The result of `std::string_view::substr` is a `std::string_view`, so no deallocated string/temporary usage here

8infy: > The result of `std::string_view::substr` is a `std::string_view`, so no deallocated string/temporary usage here

right, it's a substring of a `string_view`... missed that one

Hazel/src/Hazel/Core/Log.h
bmeneguele: About the exception: agreed.
About the the if without {} being prohibited: that's not the only situation where macros like this one could break things... of course, this is a "final/error" macro, but if you have a macro that actually do something (calls another function, declares a var, or whatever) you could make use of it inside a while(<macro>). In order to make it fine, the solution would be:
```
({ ::Hazel::Log::GetClientLogger()->critical(__VA_ARGS__); throw std::runtime_error("Fatal error."); })
```
or
```
do { ::Hazel::Log::GetClientLogger()->critical(__VA_ARGS__); throw std::runtime_error("Fatal error."); } while(0)
```
now you must add a trailing semicolon and you have it guarded for both if() without brackets and inline conditions situations.

Hazel/CMakeLists.txt
```diff
@@ -0,0 +1,103 @@
...
+        vendor/imgui/imgui_draw.cpp)
+
+set(STBIMG_SRC
+        vendor/stb_image/stb_image.cpp)
+
+set(GLAD_SRC
+        vendor/Glad/src/glad.c)
+
+set(SOURCES
+        src/Hazel/Core/Application.cpp
```
LewyIsChewy, line 32: I think it's better if you just glob the files here instead of hardcoding all the source files to automate the process,

file(GLOB SRC_FILES "src/\*.c\* src/\*.c/\*.c\* src/\*.c/\*.c\*/\*.c\* src/\*.h\* src/\*.h/\*.h\* src/\*.h/\*.h\*/\*.h\*)

Then just use $SRC_FILES as necessary

Sandbox/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+include_directories(../Hazel/vendor/Glad/include)
+include_directories(../Hazel/vendor/GLFW/include)
+include_directories(../Hazel/vendor/imgui)
+include_directories(../Hazel/vendor/stb_image)
+include_directories(../Hazel/vendor/glm)
+
+include_directories(../Hazel/src)
+
+set(SOURCES
+        src/ExampleLayer.cpp
```
LewyIsChewy, line 16: Same globbing comment above

Hazel/CMakeLists.txt
LewyIsChewy, line 6: This adds this define for all build types right? Probably best to only include it in debug only with something like, add_compile_definiations($<$CONFIG:Debug:HZ_DEBUG>)

Hazel/CMakeLists.txt
NekkoDroid: you can also just `file(GLOB_RECURSE SRC_FILES src/*)`

LovelySanta:
I'll close this as #82 is back up to date, and as mentioned in #308, cmake is not required. We chose premake, and for now we're sticking with it...

PR #243, opened 2020-05-16
Add assertions

WhoseTheNerd:
If you don't set callback then you get std::bad_function or something like that. Since there is no error messages that tell you that callback is uninitialized or empty

Issues this solves        | None
Other PRs this solves  | None

Fix is to add assertions before calling EventCallback


HerrDingenz:
If we would implement this then it will always assert upon window creation, because we set the EventCallback for the Window after creating the window and setting up the glfwCallbacks

WhoseTheNerd:
@HerrDingenz No. Because after window creation we quickly set the callback and on first event from glfw the callback is not null. Basically the callback is set far quicker than the first glfw generated event.

LovelySanta:
This is just done by the engine, the question we were having, is there any use case where this would be the case. I feel like adding asserts everywhere, just because we forgot a line is a bit too much...

PR #247, opened 2020-05-19
Branching instead of indexing the texture sampler

### LINKED ISSUE #238: sampler2D should not use vertex attribute as index
#### Describe the bug
*Credit to @KronosB#2408 (discord)*
Some people have been complaining about some visual glitches (see picture) on textures. Recently another report [on trello](https://trello.com/c/dTDRRXUU) came in where it did only render black triangles.

image

From the OpenGL specs ([link](https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.60.html#opaque-types)):
> When aggregated into arrays within a shader, they can only be indexed with a dynamically uniform integral expression, otherwise results are undefined.

So the use of a vertex attribute to index a sampler2D array in the shader is illegal:
https://github.com/TheCherno/Hazel/blob/48269a5880dff948177b38364f56281be44d1321/Sandbox/assets/shaders/Texture.glsl#L42

#### To Reproduce
Render triangles using multiple textures. Does not occur on all systems. 

#### Expected behavior
To have no artifacts and render textures correctly.

#### Screenshots
If applicable, add extra screenshots to help explain your problem.

#### Additional context
Quote from cherno:
> Yes, technically we should be branching instead of array indexing, I'm just really not a fan of that so i wrote it that way. Same with a lot of mobile devices and OpenGLES 2.0. I just looked at Sparky and confirmed, it's in the D3D11 shaders as well: https://github.com/TheCherno/Sparky/blob/master/Sparky-core/src/sp/platform/directx/shaders/BatchRenderer.hlsl

LovelySanta:
As stated in #238, the texture sampler can only be indexed with a dynamically uniform integral expression. Not doing so results in quads rendering black instead of rendering textures on certain systems.

Issues this solves       | Closes #238

Branch the sampler instead of indexing it. Since currently Hazel supports 32 texture slots, I've branched up to 32 textures.

#### Additional context
In discord this is confirmed to resolve the black quad drawings. Since it wasn't an issue for me, I can only test that it still works. However, @9Y0 confirmed that this resolves the black quad drawing.

### DIFF
--- a/Sandbox/assets/shaders/Texture.glsl
+++ b/Sandbox/assets/shaders/Texture.glsl
@@ -39,5 +39,40 @@ uniform sampler2D u_Textures[32];
 
 void main()
 {
-	color = texture(u_Textures[int(v_TexIndex)], v_TexCoord * v_TilingFactor) * v_Color;
+	vec4 texColor = v_Color;
+	switch(int(v_TexIndex))
+	{
+		case 0: texColor *= texture(u_Textures[0], v_TexCoord * v_TilingFactor); break;
+		case 1: texColor *= texture(u_Textures[1], v_TexCoord * v_TilingFactor); break;
+		case 2: texColor *= texture(u_Textures[2], v_TexCoord * v_TilingFactor); break;
+		case 3: texColor *= texture(u_Textures[3], v_TexCoord * v_TilingFactor); break;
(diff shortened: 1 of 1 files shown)


frederikja163:
How is performance for this? Currently doing something similar in my own engine, and I've heard several times that branching for fragment shaders is terrible. I'm assuming switches use branches under the hood?

PR #248, opened 2020-05-20
Added error logs to Instrumentor in case no session is open.

HerrDingenz:
The instrumentor doesn't print error messages when writing profile results or ending a session if there is no session open. This might be unclear to the user.

Add logging of these errors in WriteProfile and InternalEndSession functions.

### DIFF
--- a/Hazel/src/Hazel/Debug/Instrumentor.h
+++ b/Hazel/src/Hazel/Debug/Instrumentor.h
@@ -95,6 +95,10 @@ namespace Hazel {
 				m_OutputStream << json.str();
 				m_OutputStream.flush();
 			}
+			else if (Log::GetCoreLogger()) // Edge case: WriteProfile() might be before Log::Init()
+			{
+				HZ_CORE_ERROR("Could not write ProfileResult, Instrumentor is not in session.");
+			}
 		}
 
 		static Instrumentor& Get()
@@ -127,6 +131,10 @@ namespace Hazel {
 				m_OutputStream.close();
 				delete m_CurrentSession;
 				m_CurrentSession = nullptr;
+			}
+			else if (Log::GetCoreLogger()) // Edge case: InternalEndSession() might be before Log::Init()
+			{
+				HZ_CORE_ERROR("Could not end session, Instrumentor is not in session.");
 			}
 		}
 


Hazel/src/Hazel/Debug/Instrumentor.h
```diff
@@ -95,6 +95,10 @@ namespace Hazel {
 				m_OutputStream << json.str();
 				m_OutputStream.flush();
 			}
+			else if (Log::GetCoreLogger()) // Edge case: WriteProfile() might be before Log::Init()
+			{
+				HZ_CORE_ERROR("Could not write ProfileResult, Instrumentor is not in session.");
```
ReDucTor, line 100: Why is this an error? Isn't it reasonable not to have an active session? 

tbh it should early return and not form the json when it doesn't, not necessarily throw an error

LovelySanta: If you don't have an active session, it means you are not profiling. When you're not profiling, you should not waste computing power by timing functions and generating profile results.

Returning early can only happen when you check m_CurrentSession. This would mean that you would have to require the lock a lot earlier (or twice if you release it). The only reason the stringstream is made beforehand is to reduce the time we lock the mutex?

ReDucTor: It probably depends on what is the future of the instrumentation, is there a desire for the ability to have this toggled on and off at runtime? Or is it solely triggered at compile time?

Ya the locking should only be necessary when it needs to write, tbh I'm very surprised that there is a flush within a lock, or even file io, that is going to be an awful performance hit, especially for something that is meant to be profiling.

LovelySanta: Keep in mind that cherno just copy pasted this as a temporary solution that 'works' for what it is intended for now. 

I would say that it is fine for now, but if it can be improved upon, i'm not against it. However, the toggling at runtime, I think that would need its own PR. I would expect in that case to still create the sessions when it is enabled, but toggled off, and prevent results from written to a session, in which case this error would never pop up at that point.

HerrDingenz: The error messages are just there to show the "user" that he might have forgotten to call HZ_PROFILE_BEGIN_SESSION(...) before calling HZ_PROFILE_SCOPE(name) or HZ_PROFILE_FUNCTION().
And then he wonders why there is no profiling file created.

Emilundpixeln: `HZ_CORE_ERROR` uses the logger from `Log::GetCoreLogger()` so if `Log::GetCoreLogger()` returns null calling `HZ_CORE_ERROR` will always fail. 
Logging the there is no logger doesn't make sense.

PR #249, opened 2020-05-20
Simplify log macros

3x3y3z3t:
This PR will improve Logging macros as described below.

This PR adds/improves the following:
- Add `HZ_CORE_LIB` preprocessor to Hazel engine Core.
- Add `HZ_LOG_DEBUG` which expands to spdlog's debug logger.
- All log macros `HZ_CORE_XXX` and `HZ_XXX` now become `HZ_LOG_XXX`. Depends on where the macro is called, it will expand to either Core or Client log code.
    - If the macro is called in Core (that have `HZ_CORE_LIB` defined) it will expand to Core log code.
    - If the macro is called in Client (that doesn't have `HZ_CORE_LIB` defined) it will expand to Client log code.
- Remove assert macro `HZ_CORE_ASSERT`. The other macro will expand to `HZ_LOG_ERROR` which expands to correct log code.
- Macros `HZ_LOG_DEBUG` and `HZ_LOG_TRACE` will expands to nothing in non-debug mode.

#### Additional context
Half of the log macros are almost the duplicate of the others. We can simplify them so users don't have to decide whether they should use Core log macro or Client log macro, instead they just need to call the same Log macro and let the compiler decide which code to expands to.


Hazel/src/Hazel/Core/Log.h
```diff
@@ -21,16 +21,23 @@ namespace Hazel {
...
-#define HZ_CORE_CRITICAL(...) ::Hazel::Log::GetCoreLogger()->critical(__VA_ARGS__)
-
-// Client log macros
-#define HZ_TRACE(...)         ::Hazel::Log::GetClientLogger()->trace(__VA_ARGS__)
-#define HZ_INFO(...)          ::Hazel::Log::GetClientLogger()->info(__VA_ARGS__)
-#define HZ_WARN(...)          ::Hazel::Log::GetClientLogger()->warn(__VA_ARGS__)
-#define HZ_ERROR(...)         ::Hazel::Log::GetClientLogger()->error(__VA_ARGS__)
-#define HZ_CRITICAL(...)      ::Hazel::Log::GetClientLogger()->critical(__VA_ARGS__)
\ No newline at end of file
+#ifdef HZ_CORE_LIB
```
ReDucTor, line 24: This won't work well for inline code

premake5.lua
```diff
@@ -57,7 +57,8 @@ project "Hazel"
 	defines
 	{
 		"_CRT_SECURE_NO_WARNINGS",
-		"GLFW_INCLUDE_NONE"
+		"GLFW_INCLUDE_NONE",
+		"HZ_CORE_LIB"
```
ReDucTor, line 61: This sort of macro is just waiting to become a one-definition-rule violation.

Hazel/src/Hazel/Debug/Instrumentor.h
```diff
@@ -48,7 +48,7 @@ namespace Hazel {
 				// profiling output.
 				if (Log::GetCoreLogger()) // Edge case: BeginSession() might be before Log::Init()
 				{
-					HZ_CORE_ERROR("Instrumentor::BeginSession('{0}') when session '{1}' already open.", name, m_CurrentSession->Name);
+					HZ_LOG_ERROR("Instrumentor::BeginSession('{0}') when session '{1}' already open.", name, m_CurrentSession->Name);
```
ReDucTor, line 51: This and all other logs in header files now violate the one-definition-rule, resulting in **undefined behavior**.

There should be one and only one implementation of a function, you should not have an inline function which has two different implementations, which for this it now will because it differs between two different libraries

LovelySanta:
This exact same approach (with different names for the macros) was already proposed and rejected in #116

PR #250, opened 2020-05-22
‎

BaderEddineOuaich:


LovelySanta:
Some thoughts:
- I don't know if we would want this specific in a separate file... Feels like cherno has that file there for that reason?  I might have a look later at how it looks if I put it all in one spot...
- Feels like it could be shorter if we define `HZ_DEBUGBREAK()` directly instead of splitting it in different implementation functions? That aside, it feels more readable to me, but that might be personal preferences speaking.

ReDucTor:
Maybe it's just me but this seems to be stolen from else where without acknowledgement,  did you test all these platforms? Do they all even support opengl 4.5? Pretty sure they don't.

BaderEddineOuaich:
@LovelySanta it was in the Base.h at first. it looked little confusing for much lines of code for a single implementation as to why i decided to separate it in a file.

BaderEddineOuaich:
@ReDucTor 
First i never said i own it. source has MIT LICENCE which gives everyone the right to use/modify.
secondly the implementation has nothing to do with OpenGL as you claim. it has to do with the platform. 
third i have tested the implementation myself and it seem to work. if you are so concerned about it you can test it yourself.
Thanks

ReDucTor:
If it's MIT you have just broken the license by not including the license with full attribution. 

Why I said did you test all platforms and that did they work with opengl 4.5 is because the list of platforms many won't support it, and that's a minimum requirement for hazel to run, so it would seem this is just extra code to maintain for unsupported platforms

BaderEddineOuaich:
@LovelySanta what do you think should be done at this step ?

LovelySanta:
I had to get the full convo out of my mailbox (with the deleted messages)...

Maybe we can come up with our own version of it, and only for the compilers we support? Since we only use the `raise(SIGTRAP)` for what will be linux... Maybe we should find a more solid solution?

PR #252, opened 2020-05-28
Add Linux Support

WhoseTheNerd:
Hazel doesn't support Linux yet. I know there other Linux PRs exist. This PR doesn't contain any other junk(cmake, etc) and is maintained to the current Hazel source tree.

Other PRs this solves    | closes #237, closes #82

Proposed fixes:

1. Remove all cross-platform code preprocessor macros
2. Update premake5.lua to make it compile under linux gcc.
3. Add linux libraries to premake5.lua

#### Additional context

- [x] Test Linux
- [x] Test Windows

### DIFF
--- a/Hazel/src/Hazel/Core/Input.cpp
+++ b/Hazel/src/Hazel/Core/Input.cpp
@@ -1,7 +1,9 @@
 #include "hzpch.h"
 #include "Hazel/Core/Input.h"
 
-#if defined(HZ_PLATFORM_WINDOWS) || defined(HZ_PLATFORM_LINUX)
+#if defined(HZ_PLATFORM_WINDOWS)
+	#include "Platform/Windows/WindowsInput.h"
+#elif defined(HZ_PLATFORM_LINUX)
 	#include "Platform/Windows/WindowsInput.h"
 #endif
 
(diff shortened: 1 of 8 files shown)


Hazel/src/Hazel/Core/Input.cpp
```diff
@@ -11,7 +13,11 @@ namespace Hazel {
 
 	Scope<Input> Input::Create()
 	{
-	#ifdef HZ_PLATFORM_WINDOWS
+	#if defined(HZ_PLATFORM_WINDOWS)
+		return CreateScope<WindowsInput>();
+	#elif defined(HZ_PLATFORM_LINUX)
```
DaveAxiom, line 18: These functions are delegated to a library that is cross-platform. Adding compiler directive branches is unneeded and only convolutes Hazel's source code.

WhoseTheNerd: This is suggestion by @LovelySanta. Linux version is using Windows implementation since that is cross-platform, but really it should use its own implementation, like LinuxInput and etc.

WhoseTheNerd: This is suggestion by @LovelySanta. Linux version is using Windows implementation since that is cross-platform, but really it should use its own implementation, like LinuxInput and etc.

LovelySanta: I think that would be something for Cherno to implement tho...

PR #256, opened 2020-06-05
Added Exit Docking Space Button

Clickyz:


LovelySanta:
As cherno mentioned in his last video, the sandbox is meant to be the runtime test environment. The docking space is only there for the editor, which we are currently working towards. Since this is currently only a test, it belongs in Sandbox. Once the editor is somewhat in place, this will move to its own project. So this feels like not needed for both cases, where the editor is only docking space, and sandbox only runtime testing.

Clickyz:
> As cherno mentioned in his last video, the sandbox is meant to be the runtime test environment. The docking space is only there for the editor, which we are currently working towards. Since this is currently only a test, it belongs in Sandbox. Once the editor is somewhat in place, this will move to its own project. So this feels like not needed for both cases, where the editor is only docking space, and sandbox only runtime testing.

Ok
:)

LovelySanta:
As it is clear now with the two seperate projects, this is not needed. Closing this as resolved.

PR #260, opened 2020-06-05
Removed Duplicate ImGui Window Creation

ultamatum:
Currently the ImGui Settings window creation is duplicated. This is unnecessary as to be included in the dockspace the window should be created after not within.

#### Proposed fix
Removed else statement and duplicate code.

### DIFF
--- a/Sandbox/src/Sandbox2D.cpp
+++ b/Sandbox/src/Sandbox2D.cpp
@@ -129,40 +129,23 @@ void Sandbox2D::OnImGuiRender()
 			ImGui::EndMenuBar();
 		}
 
-		ImGui::Begin("Settings");
-
-		auto stats = Hazel::Renderer2D::GetStats();
-		ImGui::Text("Renderer2D Stats:");
-		ImGui::Text("Draw Calls: %d", stats.DrawCalls);
-		ImGui::Text("Quads: %d", stats.QuadCount);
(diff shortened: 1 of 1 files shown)


LovelySanta:
It must be created before the last `ImGui::End()` in order to be dockable, as demonstrated in the [video](https://www.youtube.com/watch?v=UljMVrQ_zYY). The implementation you've provided will not work as intended.

Cherno also mentioned that the editor will have a seperate project, so this is just temporary as the sandbox is the runtime testing while the dockspace will be part of the editor. This will become more clear in future videos.

ultamatum:
It does work, it wasn't tested in the video. It just needs to be created after the line `ImGui::DockSpace(dockspace_id, ImVec2(0.0f, 0.0f), dockspace_flags);` to be included in the dockspace.

LovelySanta:
oh ok, it didn't seem like that in the video. Let's first see how cherno implements the docking before trying to clean up the code base then.

LovelySanta:
Now that Cherno moved Hazelnut to a seperate project (https://github.com/TheCherno/Hazel/commit/9ddc33488e50da9550023077227a4b973867c1c2), the duplicate creation in sandbox is also gone, and is not duplicated in Hazelnut itself. Closing this as resolved.

Issue #261, opened 2020-06-07
Foreign Keyboard Management


T4ouf:
#### Description
Currently Hazel only "works" for QWERTY keyboards. I mean that input polling is incorrect for other type of keyboards (like AZERTY or QWERTZ keyboards). (I am not sure if this issue should be in 'bug' or in 'feature' section :/ )

#### To Reproduce
0. Plug a foreign keyboard and make sure it works properly when you type (meaning that it prints the key that is on the keyboard. Make sure that your OS is not overwritting the default keyboard keys). I am currently using a french AZERTY keyboard.
1. Go to SandboxApp.cpp
2. Paste the following code in the OnEvent(....) function : 
```cpp
if (event.GetEventType() == Hazel::EventType::KeyPressed) {
		Hazel::KeyPressedEvent& e = (Hazel::KeyPressedEvent&)event;
		if(e.GetKeyCode() == HZ_KEY_A)
			HZ_TRACE("{0}", char(e.GetKeyCode()));
		}
}
```
3. Run SandboxApp.cpp.
4. If you have an AZERTY keyboard (which is my case) : when you click on "A" (where "Q" is located on a QWERTY keyboard), it will not work. You need to press Q (where "A" is located on a QWERTY keyboard) to enter the "if" statement.

#### Expected behavior
In a "normal" case, I would expect the engine (or the underneath library) to handle correctly all type of keyboards. Pressing "A" should not mean pressing any other key than "A" (regardless of the keyboard type)

### Why is it a problem ?
It can lead to bugs in the usage of Hazel by foreign developpers (+ can mislead the users).

#### Operating system: 
- OS: win-64
- Version: Windows 10 professional

BaderEddineOuaich:
This used to be an issue for non us keyboards since GLFW exists...
although in glfw 3.2, they provide glfwGetKeyName function. which can be used to solve this issue by mapping keys to be used by all keyboard types.
i agree with @T4ouf as this can be serious for non us keyboard users. and should be one of the priorities to implement.

T4ouf:
I'll check glfwGetKeyName and see how to implement it correctly ! If it works, I'll make a pull request. (Yet I don't know how much work I'll be able to involve in this :/ Also, maybe it is a big change and it will need to be supervised by @TheCherno in a video or something ?)

T4ouf:
I think I have found a solution to the problem ! I've made a pull request in order to fix this issue ! Thank you for your help @BaderEddineOuaich ! 😃

[...]

PR #262, opened 2020-06-08
Handling non-QWERTY keyboards with GLFW

T4ouf:
As described in the issue #261 , currently Hazel doesn't handle non-QWERTY keyboards. This can be a real issue for non-QWERTY users (like AZERTY users).

#### PR impact 
Issues this solves       |  #261 

#### Proposed fix
Now in the glfwSetKeyCallBack() function, we check if the current key is a letter. If so, we adapt to the keyboard layout the keycode that the key event will have.

### DIFF
--- a/Hazel/src/Platform/Windows/WindowsWindow.cpp
+++ b/Hazel/src/Platform/Windows/WindowsWindow.cpp
@@ -90,6 +90,12 @@ namespace Hazel {
 		{
 			WindowData& data = *(WindowData*)glfwGetWindowUserPointer(window);
 
+			//Adapting to non-QWERTY keyboards (like AZERTY or QWERTZ)
+			if (key >= GLFW_KEY_A && key <= GLFW_KEY_Z) {
+				const std::string keyName = glfwGetKeyName(key, scancode);
+				key = GLFW_KEY_A + (std::toupper(keyName[0]) - 'A'); //this way, we ensure that no matter the code that GLFW uses, the keycode will be correct
+			}
+
 			switch (action)
 			{
 				case GLFW_PRESS:


BaderEddineOuaich:
I tried it with non us keyboards, works just fine. 
i think we should make it little cleaner? the plan is to make the keycode <strong>cross-keyboard</strong> if necessary and immediately after we receive the key event from GLFW. so we don't have to do it for each key handling function. i made some changes based on your solution in this <strong>[fork](https://github.com/TheCherno/Hazel/compare/master...BaderEddineOuaich:master)</strong>
tell me what you think about it.
Thanks!

T4ouf:
I understand... it was maybe not super clean to have code copy-pasted twice... :/ 

I have checked your version. I think you have the right idea : we should try to abstract it ! 😃 
However your version seems has some problems :
- It works for events, not for polling (you forgot to put it in WindowsInput.cpp in IsKeyPressedImpl())
- I don't think we should have glfw calls (which is platform specific) in the platform-independant Input.h file...

Also I have some questions on some choices : 
- I am not a big fan of making a macro => Why not making a function and get type checking ?
- Why using a pointer ? 
- Why a do-while(false) loop ?

So I abstracted my code into a function GetCrossLayoutKey() that I put in Input.h that needs to be implemented by the plaform-specific implementation (WindowsInput in our case). I used the same organization as the other functions in Input.h to keep the consistency !

I pushed my code, it should be visible now ! 

=> Anyway, thank you for your review ! It helped me a lot ! 😃

BaderEddineOuaich:
the thing about macro is no big deal since it expands to the same code as an inline function, and the do{}while(false) its just a scope to wrap the macro body. 
i used the pointer there because the macro cannot return the key but modify it by passing the pointer to it.
not sure how it doesn't work for polling. as the keycode is being modified instantly (if necessary) after receiving from glfw key event. 
you're right about calling glfw functions in platform-independent code, i shouldn't do that!!!
however, the latest changes you made are much better!
lets wait for others to share their ideas with us.
Thank you!

PR #265, opened 2020-06-11
Fix flipped ImGui image

FryDay:
#### Describe the issue
ImGui uses a different coordinate system, so images are flipped.

#### Proposed fix
UV coords can be passed to Image. The image is flipped correctly by passing 0,1 and 1,0


FryDay:
Just saw this was done in the last episode.

PR #268, opened 2020-06-16
Fix black flicker on resize

xforce:
#### Describe the issue
Previously resize happened in the `OnImGuiRender` function
this runs after `OnUpdate` where the framebuffer gets filled.
This resulted in ImGui rendering an unfilled/new texture on every
resize.

#### Proposed fix
This changes the code to do the resize as the first step in `OnUpdate` before we will render to the framebuffer, so we always have a filled framebuffer to render in `OnImGuiRender`.

### DIFF
--- a/Hazelnut/src/EditorLayer.cpp
+++ b/Hazelnut/src/EditorLayer.cpp
@@ -32,6 +32,15 @@ namespace Hazel {
 	{
 		HZ_PROFILE_FUNCTION();
 
+		// Resize
+		if (Hazel::FramebufferSpecification spec = m_Framebuffer->GetSpecification();
+			m_ViewportSize.x > 0 && m_ViewportSize.y > 0 && // zero sized framebuffer is invalid
+			(spec.Width != m_ViewportSize.x || spec.Height != m_ViewportSize.y))
+		{
+			m_Framebuffer->Resize((uint32_t)m_ViewportSize.x, (uint32_t)m_ViewportSize.y);
+			m_CameraController.OnResize(m_ViewportSize.x, m_ViewportSize.y);
+		}
+
 		// Update
 		if (m_ViewportFocused)
 			m_CameraController.OnUpdate(ts);
@@ -157,13 +166,8 @@ namespace Hazel {
 		Application::Get().GetImGuiLayer()->BlockEvents(!m_ViewportFocused || !m_ViewportHovered);
 
 		ImVec2 viewportPanelSize = ImGui::GetContentRegionAvail();
-		if (m_ViewportSize != *((glm::vec2*)&viewportPanelSize))
-		{
-			m_Framebuffer->Resize((uint32_t)viewportPanelSize.x, (uint32_t)viewportPanelSize.y);
-			m_ViewportSize = { viewportPanelSize.x, viewportPanelSize.y };
+		m_ViewportSize = { viewportPanelSize.x, viewportPanelSize.y };
 
-			m_CameraController.OnResize(viewportPanelSize.x, viewportPanelSize.y);
-		}
 		uint32_t textureID = m_Framebuffer->GetColorAttachmentRendererID();
 		ImGui::Image((void*)textureID, ImVec2{ m_ViewportSize.x, m_ViewportSize.y }, ImVec2{ 0, 1 }, ImVec2{ 1, 0 });
 		ImGui::End();
--- a/Hazelnut/src/EditorLayer.cpp
+++ b/Hazelnut/src/EditorLayer.cpp
@@ -34,7 +34,7 @@ namespace Hazel {
 
 		// Resize
 		if (Hazel::FramebufferSpecification spec = m_Framebuffer->GetSpecification();
-			m_ViewportSize.x > 0 && m_ViewportSize.y > 0 && // zero sized framebuffer is invalid
+			m_ViewportSize.x > 0.0f && m_ViewportSize.y > 0.0f && // zero sized framebuffer is invalid
 			(spec.Width != m_ViewportSize.x || spec.Height != m_ViewportSize.y))
 		{
 			m_Framebuffer->Resize((uint32_t)m_ViewportSize.x, (uint32_t)m_ViewportSize.y);


LovelySanta:
Looks like a good solution to me 👍 

This does feel like part of the renderer preparations, so maybe this should be part of the renderer prep scope, right before binding the framebuffer?

https://github.com/TheCherno/Hazel/blob/24990184428c4a0e96f9a752df49f00459f79766/Hazelnut/src/EditorLayer.cpp#L49-L54

Another reason why I would change this, it does feel like the CameraController should be updated depending on the current visible frame instead of on the new one, meaning that the first thing would be the `m_CameraController.OnUpdate(ts);`

So my suggestion would look like:
```cpp
	void EditorLayer::OnUpdate(Hazel::Timestep ts)
	{
		HZ_PROFILE_FUNCTION();

		// Update
		m_CameraController.OnUpdate(ts);

		// Render
		Hazel::Renderer2D::ResetStats();
		{
			HZ_PROFILE_SCOPE("Renderer Prep");

			// Resize
			if (Hazel::FramebufferSpecification spec = m_Framebuffer->GetSpecification();
				m_ViewportSize.x > 0 && m_ViewportSize.y > 0 && // zero sized framebuffer is invalid
				(spec.Width != m_ViewportSize.x || spec.Height != m_ViewportSize.y))
			{
				m_Framebuffer->Resize((uint32_t)m_ViewportSize.x, (uint32_t)m_ViewportSize.y);
				m_CameraController.OnResize(m_ViewportSize.x, m_ViewportSize.y);
			}

			m_Framebuffer->Bind();
			Hazel::RenderCommand::SetClearColor({ 0.1f, 0.1f, 0.1f, 1 });
			Hazel::RenderCommand::Clear();
		}

		... // The rest of the existing code

Any thoughts?

**Comment by xforce** [comment]
Hmm, not entirely sure. Would have to look into how the camera does things. But I think right now it doesn't really make any difference.

The PR as is was to maintain original resize 'flow'.

Before:
```
OnUpdate
OnImGuiRender <- Resize Happened in here
OnUpdate
```

After:
```
OnUpdate
OnImGuiRender <- Resize change 'request' here
OnUpdate <- Change viewport and camera before actual render
```

I am happy to change it, but I am not sure what the intended 'flow' is supposed to be here now. Right

File: Hazelnut/src/EditorLayer.cpp [hunk 1]
```diff
@@ -150,14 +159,10 @@ namespace Hazel {
...
-		{
-			m_Framebuffer->Resize((uint32_t)viewportPanelSize.x, (uint32_t)viewportPanelSize.y);
-			m_ViewportSize = { viewportPanelSize.x, viewportPanelSize.y };
+		m_ViewportSize = { viewportPanelSize.x, viewportPanelSize.y };
 
-			m_CameraController.OnResize(viewportPanelSize.x, viewportPanelSize.y);
-		}
 		uint32_t textureID = m_Framebuffer->GetColorAttachmentRendererID();
 		ImGui::Image((void*)textureID, ImVec2{ m_ViewportSize.x, m_ViewportSize.y }, ImVec2{ 0, 1 }, ImVec2{ 1, 0 });
 		ImGui::End();
```
LovelySanta, line 168: > The PR as is was to maintain original resize 'flow'.
> 
> Before:
> 
> ```
> OnUpdate
> OnImGuiRender <- Resize Happened in here
> OnUpdate
> ```
> 
> After:
> 
> ```
> OnUpdate
> OnImGuiRender <- Resize change 'request' here
> OnUpdate <- Change viewport and camera before actual render
> ```

I see that you want to maintain the current flow, which is fine by me. After having another look, what you describe is more or less correct, but I think the current solution is not solving the flickering? Let my try and explain..

So first of all you got ImGui that is registering the resizing when you resize the window. This will propagate the new size inside ImGui, so to `ImGui::GetContentRegionAvail();`, which we can then use.

The issue that we're having is that the created framebuffer is not the same dimensions as the one that ImGui is rendering, which is why it's turning black in the first place? The solution for that is to render the 'old' sized framebuffer, like when we hardcoded the checkerboard size before:
- When you make it smaller than the checkerboard size, it will render the excess checkerboard outside of the window, so it won't be visible.
- When you make it bigger, the (black) ImGui background will be visible on the side/bottom.

However I just noticed that you did it in the wrong order:
https://github.com/TheCherno/Hazel/blob/24990184428c4a0e96f9a752df49f00459f79766/Hazelnut/src/EditorLayer.cpp#L160-L169

1. Line 163 you get the new content  size (of the empty window)
2. Line 164 you update the viewport size, which act as the 'request' to update the viewport
3. Line 167 you render the old size viewport, but you render it with the new viewport size.

Where I think you should:

1. Line 163, extract the new content size (of the empty window)
2. Line 167 render the old size viewport with the old viewport size
3. Line 164 update the viewport size, which act as the 'request' to update the viewport.

```suggestion
		ImGui::Begin("Viewport");
		ImVec2 viewportPanelSize = ImGui::GetContentRegionAvail();

		uint32_t textureID = m_Framebuffer->GetColorAttachmentRendererID();
		ImGui::Image((void*)textureID, ImVec2{ m_ViewportSize.x, m_ViewportSize.y }, ImVec2{ 0, 1 }, ImVec2{ 1, 0 });
		
		m_ViewportSize = { viewportPanelSize.x, viewportPanelSize.y };
		ImGui::End();
```

LovelySanta:
> Hmm, not entirely sure. Would have to look into how the camera does things. But I think right now it doesn't really make any difference.
> 
> ...
> 
> I am happy to change it, but I am not sure what the intended 'flow' is supposed to be here now. Right

You don't have to change it. I can see that you want to keep the layout, and it won't affect the camera at all as of now.

The main reason I asked is for the profiler, where it doesn't make a lot of sense (to me) that the resize would be counted as part of the on_update loop, and not part of the rendering, hence why I would've put it inside the `"Renderer prep"` scope. Another thought would be to put a profile scope `"Renderer resize"` or something. However this might indeed be outside the scope of this PR, so no need to change that.

Hazelnut/src/EditorLayer.cpp
```diff
@@ -32,6 +32,15 @@ namespace Hazel {
 	{
 		HZ_PROFILE_FUNCTION();
 
+		// Resize
+		if (Hazel::FramebufferSpecification spec = m_Framebuffer->GetSpecification();
+			m_ViewportSize.x > 0 && m_ViewportSize.y > 0 && // zero sized framebuffer is invalid
```
LovelySanta, line 37:
```suggestion
			m_ViewportSize.x > 0.0f && m_ViewportSize.y > 0.0f && // zero sized framebuffer is invalid
```

Hazelnut/src/EditorLayer.cpp
xforce: Size is unrelated to it turning black. The reason it is flickers black on resize is that on every size change it was creating a new framebuffer texture and directly rendering that in ImGui before actually rendering anything to the framebuffer, aka rendering empty new texture.

Using a different size than the texture in `ImGui::Image` 'only' causes a slight stretch, no flicker really, which I didn't really think about when I created the PR, so I guess something along the lines of your suggestion does make sense.
Though I think instead of using `m_ViewportSize` for the Image render it might be better to use the actual viewport size to avoid first frame incorrect size (0 in this case).

xforce: > Size is unrelated to it turning black. The reason it is flickers black on resize is that on every size change it was creating a new framebuffer texture and directly rendering that in ImGui before actually rendering anything to the framebuffer, aka rendering empty new texture.
> 
> Using a different size than the texture in `ImGui::Image` 'only' causes a slight stretch, no flicker really, which I didn't really think about when I created the PR, so I guess something along the lines of your suggestion does make sense.
> Though I think instead of using `m_ViewportSize` for the Image render it might be better to use the actual viewport size to avoid first frame incorrect size (0 in this case).

Hmm, I was just playing with this.
Using the old size results in small-ish black bars around the Viewport during resize as well as ImGui flickering the vertical scrollbar on shrinking.
The current PR just stretches the image to always be the same size as the Viewport window.

Personally I am not sure which one I find better.

More correct is the black bars, the flickering scrollbar really makes this not nice and I don't think there is a way in ImGui to disable that at the moment.

xforce:
> > Hmm, not entirely sure. Would have to look into how the camera does things. But I think right now it doesn't really make any difference.
> > ...
> > I am happy to change it, but I am not sure what the intended 'flow' is supposed to be here now. Right
> 
> You don't have to change it. I can see that you want to keep the layout, and it won't affect the camera at all as of now.
> 
> The main reason I asked is for the profiler, where it doesn't make a lot of sense (to me) that the resize would be counted as part of the on_update loop, and not part of the rendering, hence why I would've put it inside the `"Renderer prep"` scope. Another thought would be to put a profile scope `"Renderer resize"` or something. However this might indeed be outside the scope of this PR, so no need to change that.

I agree, though I think eventually this frame buffer resize should probably live in some other place, like a PreUpdate or PostRender phase maybe, not sure.

Hazelnut/src/EditorLayer.cpp
LovelySanta: After some searching: According to ocornut/imgui#1089, there should be a windowflag `ImGuiWindowFlags_NoScrollbar` to disable the scrollbar? Maybe that might help?

Not sure what behaviour we want. My initial thought process was, assume that we stretch the right side, while dragging the screen would jitter a bit. The scroll bar seems a big nono for sure. 

If that windowflag works, we still have to decide which to choose, otherwise the choice is easy to just stretch it a bit... My personal preference would go to the small black boarder when making it larger, if it works, as that will result in 'perfect' behaviour.

xforce: Ah yes, must have that flag. I does work as advertised, no scrollbars anymore.
So now just need a decision on what we want.

LovelySanta: Assuming 99.95% of the resizing will happen with the mouse, the keyboard inputs do not affect what's visual on the screen as feedback to application behavior. So we can assume that there are no mouse inputs at the time like clicking swiping, ...

Hence why I would go for the stretching, without setting the WindowFlags, and as a result not have the black borders during that frame, and have a small distortion in the rendering as it stretches/squashes the frame a bit.

LovelySanta: After testing it on my 4K monitor, I can almost not see the streching/squashing, however the black boarder is alot more obvious, so your approach seems to give the best compromise/result

Sam-1210:
My Purposed Solution :

the variable 'last' is declared global for temporary purpose, as it's not a good coding convention, it should be moved at any other place.


```
ImVec2 last;

void EditorLayer::OnUpdate(Hazel::Timestep ts)
{
	HZ_PROFILE_FUNCTION();

	// Update
	if (m_ViewportFocused)
		m_CameraController.OnUpdate(ts);

	// Render
	Hazel::Renderer2D::ResetStats();
	{
		HZ_PROFILE_SCOPE("Renderer Prep");

		if (m_ViewportSize != *((glm::vec2*) & last))
		{
			m_Framebuffer->Resize((uint32_t)last.x, (uint32_t)last.y);
			m_ViewportSize = { last.x, last.y };

			m_CameraController.OnResize(last.x, last.y);
		}

		m_Framebuffer->Bind();
		Hazel::RenderCommand::SetClearColor({ 0.1f, 0.1f, 0.1f, 1 });
		Hazel::RenderCommand::Clear();
	}

		m_Framebuffer->Bind();
		Hazel::RenderCommand::SetClearColor({ 0.1f, 0.1f, 0.1f, 1 });
		Hazel::RenderCommand::Clear();
	}

        //-----------exisiting code.................
```


and


```
ImGui::PushStyleVar(ImGuiStyleVar_WindowPadding, ImVec2{ 0, 0 });
ImGui::Begin("Viewport");

m_ViewportFocused = ImGui::IsWindowFocused();
m_ViewportHovered = ImGui::IsWindowHovered();
Application::Get().GetImGuiLayer()->BlockEvents(!m_ViewportFocused || !m_ViewportHovered);

ImVec2 viewportPanelSize = ImGui::GetContentRegionAvail();
last = viewportPanelSize;

uint32_t textureID = m_Framebuffer->GetColorAttachmentRendererID();
ImGui::Image((void*)textureID, ImVec2{ m_ViewportSize.x, m_ViewportSize.y }, ImVec2{ 0, 1 }, ImVec2{ 1, 0 });
ImGui::End();
ImGui::PopStyleVar();

```

Sam-1210:
> My Purposed Solution :
> 
> the variable 'last' is declared global for temporary purpose, as it's not a good coding convention, it should be moved at any other place.
> 
> ```
> ImVec2 last;
> 
> void EditorLayer::OnUpdate(Hazel::Timestep ts)
> {
> 	HZ_PROFILE_FUNCTION();
> 
> 	// Update
> 	if (m_ViewportFocused)
> 		m_CameraController.OnUpdate(ts);
> 
> 	// Render
> 	Hazel::Renderer2D::ResetStats();
> 	{
> 		HZ_PROFILE_SCOPE("Renderer Prep");
> 
> 		if (m_ViewportSize != *((glm::vec2*) & last))
> 		{
> 			m_Framebuffer->Resize((uint32_t)last.x, (uint32_t)last.y);
> 			m_ViewportSize = { last.x, last.y };
> 
> 			m_CameraController.OnResize(last.x, last.y);
> 		}
> 
> 		m_Framebuffer->Bind();
> 		Hazel::RenderCommand::SetClearColor({ 0.1f, 0.1f, 0.1f, 1 });
> 		Hazel::RenderCommand::Clear();
> 	}
> 
> 		m_Framebuffer->Bind();
> 		Hazel::RenderCommand::SetClearColor({ 0.1f, 0.1f, 0.1f, 1 });
> 		Hazel::RenderCommand::Clear();
> 	}
> 
>         //-----------exisiting code.................
> ```
> 
> and
> 
> ```
> ImGui::PushStyleVar(ImGuiStyleVar_WindowPadding, ImVec2{ 0, 0 });
> ImGui::Begin("Viewport");
> 
> m_ViewportFocused = ImGui::IsWindowFocused();
> m_ViewportHovered = ImGui::IsWindowHovered();
> Application::Get().GetImGuiLayer()->BlockEvents(!m_ViewportFocused || !m_ViewportHovered);
> 
> ImVec2 viewportPanelSize = ImGui::GetContentRegionAvail();
> last = viewportPanelSize;
> 
> uint32_t textureID = m_Framebuffer->GetColorAttachmentRendererID();
> ImGui::Image((void*)textureID, ImVec2{ m_ViewportSize.x, m_ViewportSize.y }, ImVec2{ 0, 1 }, ImVec2{ 1, 0 });
> ImGui::End();
> ImGui::PopStyleVar();
> ```


I watched the video for the first time because i was having some aspect ratio in my engine, and found the flash issue isn't fixed in current branch, I faced such issue before so i caught it, it is,
we have bind the framebuffer
we filled framebuffer and unbind
we resized framebuffer, which resulted in an empty texture
we passed empty texture to ImGui

PR #270, opened 2020-06-16
Adding virtual destructor on Framebuffer

VMormoris:
#### Describe the issue
No issue. I just added virtual destructor on the Framebuffer class (for consistency sake)

#### PR impact

#### What change
Added the below line of code on the Framebuffer class
```virtual ~Framebuffer() = default;```

#### Additional context
As discussed on the Discord channel this change isn't necessary since we are using smart pointers.

### DIFF
--- a/Hazel/src/Hazel/Core/Input.h
+++ b/Hazel/src/Hazel/Core/Input.h
@@ -11,10 +11,10 @@ namespace Hazel {
 	protected:
 		Input() = default;
 	public:
+		virtual ~Input() = default;
+
 		Input(const Input&) = delete;
 		Input& operator=(const Input&) = delete;
-
-		virtual ~Input() = default;
 
 		static bool IsKeyPressed(KeyCode key) { return s_Instance->IsKeyPressedImpl(key); }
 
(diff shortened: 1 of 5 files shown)


LovelySanta:
I feel like we should also add them to the other places where they are missing. I only found 3 classes that they were not defined.
- [`GraphicsContext`](https://github.com/TheCherno/Hazel/blob/master/Hazel/src/Hazel/Renderer/GraphicsContext.h)
- [`RendererAPI`](https://github.com/TheCherno/Hazel/blob/master/Hazel/src/Hazel/Renderer/RendererAPI.h)
- [`Input`](https://github.com/TheCherno/Hazel/blob/master/Hazel/src/Hazel/Core/Input.h)

matiasjpedro:
I'd probably try to avoid declaring a destructor where it's not needed, Why? Because it is not free, more detailed explanation: https://youtu.be/D8eCPl2zit4?t=154

salimp2009:
> I'd probably try to avoid declaring a destructor where it's not needed, Why? Because it is not free, more detailed explanation: https://youtu.be/D8eCPl2zit4?t=154

that is true about not to use destructor if not needed  but the video is mostly about a basic class ; it is not a pure virtual. I thought we better have virtual destructors so that when a derived class destructor is called, the compiler will use the right destructor. 

see Jason Turner's specific  video about virtual destructors 
https://www.youtube.com/watch?v=43qyUASBeUc

As far as I know, for a regular class it is better not to specify a destructor, copy constructor ...rule of 5 or zero; unless you are managing resources manually (raw pointers..etc) or need specific explicit copy constructor or such...

PR #272, opened 2020-06-21
adjust repeated key presses to bool instead of int

LewyIsChewy:
Repeat key presses don't increase the repeat count and hasn't been implemented since Cherno's key events video.

Change repeat count from int to bool


LovelySanta:
In the video where cherno implemented the events, he said that we will implement the repeat count in the future but that it's not needed for now. A version of the repeat count implementation is actually suggested in #142. I'll close this as #142 is closer to Cherno's vision as of now/back then.

PR #274, opened 2020-06-28
Create a type-safe method of dealing with angle measurements.

tcarrel:
Currently, when drawing rotated quads it is possible to pass degrees when radians are required.  This would likely continue to be an issue as additional features that use angles are added in the future.

Created types for Radians and Degrees.  Both use only explicit conversions to make it much more difficult to unintentionally pass
angles using incorrect units to the functions that need them.  Additionally both types are constexpr allowing computations to be made at compile time wherever possible.
I needed the constant pi for conversions so I also added PI as constant as M_PI is not defined in all implementations of the Standard Library (in Base.h).

#### Additional context
I've only tested compiling in Visual Studio 2019.  More testing/polish may be necessary but I believe everything to be correct.

Additionally, I had renamed tilingFactor to textureScale before I added these types.  I decided not to revert those as I think it is a more easily understandable name.


Hazel/src/Hazel/Core/Math/degrees.cpp
```diff
@@ -0,0 +1,72 @@
...
+		while( angle > MAX_ANGLE )
+		{
+			angle -= ONE_FULL_ROTATION;
+		}
+		while( angle < MIN_ANGLE )
+		{
+			angle += ONE_FULL_ROTATION;
+		}
+		return Degrees( angle );
+	}
```
Emilundpixeln, line 41: This will probably go wrong for extremly large angles.

Hazel/src/Hazel/Core/Math/degrees.h
```diff
@@ -0,0 +1,108 @@
...
+
+	class Degrees
+	{
+	public:
+		constexpr Degrees( void ) noexcept = default;
+		~Degrees( void ) noexcept = default;
+
+		constexpr Degrees( const Degrees& other ) noexcept : m_Angle( other.m_Angle ) {}
+		explicit Degrees( const Radians& radians ) noexcept;
+		Degrees( Degrees&& other ) noexcept { std::swap( this->m_Angle, other.m_Angle ); }
```
Emilundpixeln, line 28: Implementing a move constructer for a float seems a bit overkill.

Hazel/src/Hazel/Core/Math/degrees.cpp
tcarrel: Yeah, I forgot to take into account floating point precision for large values, it's easy just check for large values and do a straight division followed by multiplication by the fractional part (since you can't just use modulus for fp), but at that point you're losing too much precision so the actual angle just can no longer be stored anyway.  I considered just having a float for the angle that is always normalized between MIN and MAX angle and a separate integer for the number of full rotations and I also considered just reducing automatically after every math op. but bot approaches didn't seem ideal.  I can foresee many situations where a game might want to be able to determine that an object is rotated 1080 degrees rather than just storing 0 degrees for that.  I really wanted to just have integer degrees and not worry about a barely noticeable fractional angles.  Thinking about it now, the best thing would probably be to test for a threshold that uses the looped variation for smaller values where it will be faster than division and just divide whenever that would be faster.

Hazel/src/Hazel/Core/Math/degrees.h
tcarrel: "= default" works with it in Visual Studio 2019 but I wasn't sure about other compilers.   Rule of Five, if you implement one constructor, you should implement them all--obviously it's just a guideline and by itself it doesn't make a big performance difference especially in this case but they're trivial to implement if you do it as you're going but a pain in the ____ to go back and do later.  Having them gives the compiler more opportunities for optimization so I think it's wise to just throw them as you go.

Also, since the creation, itself, of Hazel is meant be educational; creating move ctors for simple things like this can be informative for newer programmers later on with larger classes.

tcarrel:
The recent updates to Hazel, i.e.: supplying transforms to the draw quad functions make these changes obsolete.   There's also some things that I probably should not have committed it as-is.   So, it's probably okay to close this request for now, but I'll leave that for someone else to decide.

However, I do think having type safe angles is a worthwhile addition to the engine so I'll work on it a bit more, just make it more polished and robust after which I'll make a new pull request.

LovelySanta:
I still feel it's a bit over the top... I feel like just creating a typdef for them should be sufficient since Hazel wants to stay consistent in using degrees.... At that point we should also create a typedef for RendererID and some others like cherno keeps saying... I'll close it for now, but we need to do something at some point...

PR #277, opened 2020-07-11
replaced std::bind with lambda in HZ_BIND_EVENT_FN macro

MomoDeve:
using std::bind in modern C++ is a bad practice, as lambdas can do the same and much more

#### PR impact

#### Proposed fix
```cpp
#define HZ_BIND_EVENT_FN(fn) std::bind(&fn, this, std::placeholders::_1)
```
is replaced with
```cpp
#define HZ_BIND_EVENT_FN(fn) [this](auto&&... args) -> decltype(auto) \ 
{ return this->fn(std::forward<decltype(args)>(args)...); }
```
~Behaviour of the code and api remain the same. args are not forwarded, but it will be weird to accept argument by value in event funtions and do not expect copying, so I do not see a problem here~
Behaviour of the code and api remain the same. Return type is correctly deduced and no additional copying of arguments performed

### Additional context
variadic args in lambdas were added in C++17. I suppose Hazel supports latest standard, so I can use this feature in engine code.

### DIFF
--- a/Hazel/src/Hazel/Core/Base.h
+++ b/Hazel/src/Hazel/Core/Base.h
@@ -67,7 +67,7 @@
 
 #define BIT(x) (1 << x)
 
-#define HZ_BIND_EVENT_FN(fn) [this](auto&&... args) { return this->fn(args...); }
+#define HZ_BIND_EVENT_FN(fn) [this](auto&&... args) -> decltype(auto) { return this->fn(std::forward<decltype(args)>(args)...); }
 
 namespace Hazel {
 
--- a/Hazel/src/Hazel/Core/Base.h
+++ b/Hazel/src/Hazel/Core/Base.h
@@ -67,7 +67,7 @@
 
 #define BIT(x) (1 << x)
 
-#define HZ_BIND_EVENT_FN(fn) std::bind(&fn, this, std::placeholders::_1)
+#define HZ_BIND_EVENT_FN(fn) [this](auto&&... args) { return this->fn(args...); }
 
 namespace Hazel {
 


MomoDeve:
We can also deduce arguments in bound function to remove redundant template specification in `Dispatch` function:
```cpp
Dispatch<EventType>(HZ_BIND_EVENT_FN(OnEvent)) ;
```
can be simplified to
```cpp
Dispatch(HZ_BIND_EVENT_FN(OnEvent)) ;
```
But this requires more than one line changes

salimp2009:
awesome. that is great and I was asking it in discord channel :)  

I used a simpler one in my code with no variadic and auto& and it works ok ... 
#define HG_BIND_EVENT_FN(fn)   [this](auto& x) {return fn(x);}

 if you are using auto&&... , as far as I know we better use fn(std::forward<decltype(args)>(args)...) because std::forward makes sure both R values and L values can be passed correctly.

if you do receive as && and you pass it as it is it might not work if you pass an L value so it is better to do this (as far as I know);
[this](auto&&... args) { return this->fn(std::forward<decltype(args)>(args)...); }
but it is better to use auto&&... with std::forward since it will be more flexible and more complete;

Even a better & complete one but a little complicated looking which will work better with STL containers and overloaded functions will be with noexcept and return type specified 
(from the book "C++ Lambda Story" by Bartomek Filipek") ;
(to prevent no throw and avoid extra copying for STL containers and determine which overloaded function to use) 
(we may not need. this in Hazel)

[](auto&&... x) \
noexcept(noexcept(foo(std::forward<decltype(x)>(x)...))) \ 
-> decltype(foo(std::forward<decltype(x)>(x)...)) \
{ return foo(std::forward<decltype(x)>(x)...); }

it is basically the same thing but explicitly specify noexcept and return type 

a sample code for this one is at ;
https://wandbox.org/permlink/r81jASiPPmYXTOmx

MomoDeve:
good point, I forgot that decltype can deduce variadic arguments. I will update my PR.
also, checked godbolt codegen, switching from bind to lambda also improves performance: https://godbolt.org/z/ndevWb

salimp2009:
yes. performance for lambda will better verse std::bind;  at least 20% on code size and also on runtime ..Jason Turner has a short video on that but I think we have enough info already :)

MomoDeve:
@salimp2009 
https://godbolt.org/z/jG6W3b
no comments.

salimp2009:
Just saw your commit Awesome..
Only; Did you test ->decltype(auto) in the return type ; cause it is different I know it is generally better to use decltype(auto) on return type generally since auto decays const and &  but if you checked and tested it should be ok

PR #283, opened 2020-07-22
Fixed potential memory leak and malformed JSON in Instrumentor. 

studiob56:
When using the Instrumentor for profiling, if a session is begun using `HZ_PROFILE_BEGIN_SESSION(name, filepath)` but `HZ_PROFILE_END_SESSION()` is not called before the program ends then the heap allocated `InstrumentationSession` is not deleted and the footer is not added to the file which creates malformed JSON. 

Also, the constructor for `Instrumentor` is public but `Instrumentor` is being used as a singleton, therefore more than one instance can be made.

Potential memory leak in `Instrumentor`, malformed JSON in profiling data, multiple instances of singleton       | None

- The memory leak and malformed JSON were fixed by writing a destructor for `Instrumentor` that called `InternalEndSession`. 
- The multiple instances of the singleton were prevented by making the constructor and destructor for `Instrumentor` private and explicitly deleting the copy and move constructors. 

#### Additional context
Tested on Windows only. I also made a couple of larger changes which can be seen in the gist here https://gist.github.com/studiob56/1a3b917b5b517045e8b5112d96fc20e3
Wherein I moved the writing to file etc to be managed by the `InstrumentationSession` itself rather than the `Instrumentor`, and also changed the instance of the session to be in an `std::optional` rather than as a heap allocation.

### DIFF
--- a/Hazel/src/Hazel/Debug/Instrumentor.h
+++ b/Hazel/src/Hazel/Debug/Instrumentor.h
@@ -107,7 +107,7 @@ namespace Hazel {
 
 		~Instrumentor()
 		{
-			InternalEndSession();
+			EndSession();
 		}		
 
 		void WriteHeader()
--- a/Hazel/src/Hazel/Debug/Instrumentor.h
+++ b/Hazel/src/Hazel/Debug/Instrumentor.h
@@ -27,15 +27,10 @@ namespace Hazel {
 
 	class Instrumentor
 	{
-	private:
-		std::mutex m_Mutex;
-		InstrumentationSession* m_CurrentSession;
-		std::ofstream m_OutputStream;
+	
 	public:
-		Instrumentor()
-			: m_CurrentSession(nullptr)
-		{
-		}
+		Instrumentor(const Instrumentor&) = delete;
+		Instrumentor(Instrumentor&&) = delete;
 
 		void BeginSession(const std::string& name, const std::string& filepath = "results.json")
 		{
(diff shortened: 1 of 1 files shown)


Hazel/src/Hazel/Debug/Instrumentor.h
```diff
@@ -105,6 +100,16 @@ namespace Hazel {
...
 	private:
 
+		Instrumentor()
+			: m_CurrentSession(nullptr)
+		{
+		}
+
+		~Instrumentor()
+		{
+			InternalEndSession();
```
LovelySanta, line 110: Shouldn't it call `EndSession()`? For now it's fine, but I'm thinking about multi threaded cases...

studiob56: Yeah good catch. I missed the comment above `InternalEndSession()` that notes you must already own the lock on mutex before calling it. Is there a reason why it doesn't own the lock itself? It would make sense for the lock to be created where it's needed rather than expecting it to be/hoping it is. If it was moved to inside `InternalEndSession()` then it would require the lock in `BeginSession(...)` to move to after the check to see if `m_CurrentSession` is active or not to ensure the rest of the method owns a lock.

studiob56: Fair enough, that makes sense. I've pushed a new commit that calls `EndSession()` instead of `InternalEndSession()` from the destructor just so the fix is complete.

PR #286, opened 2020-07-23
Fixed UVs in editor viewport

riccardopiola:
In the Hazelnut viewport using the "A" key makes the scene move RIGHT while it should move it LEFT like it has always been the case in Sandbox. The opposite behaviour is observed for the "D" key.

#### Proposed fix
Changing the UVs supplied during the creation of the viewport fixes the problem.

### DIFF
--- a/Hazelnut/src/EditorLayer.cpp
+++ b/Hazelnut/src/EditorLayer.cpp
@@ -167,7 +167,7 @@ namespace Hazel {
 		m_ViewportSize = { viewportPanelSize.x, viewportPanelSize.y };
 
 		uint32_t textureID = m_Framebuffer->GetColorAttachmentRendererID();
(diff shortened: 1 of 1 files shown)


LovelySanta:
> In the Hazelnut viewport using the "A" key makes the scene move RIGHT

This is the intended behaviour? Pressing A let's you move 'LEFT'. So you sit in front of your PC, move your **head** to the **left**, assume that your eyes have a fixed position, you see your **PC** move to the **right**. So the 'scene' moves to the opposite side. Cherno did mention this in [the camera controller video](https://www.youtube.com/watch?v=q7rwMoZQSmw) (around 35 min)

riccardopiola:
You're right, sorry i got confused.

PR #288, opened 2020-07-25
Avoid imgui.ini under version control - fixes #287

### LINKED ISSUE #287: imgui.ini under version control causes minor annoyance
#### Describe the bug
Not strictly a "bug", just a minor annoyance.
imgui.ini is under version control.
This file gets changed every time you run Hazelnut (imgui automatically saves window position and layout to this file).
This in turn means that git will say you've got changes in your workspace even when you haven't changed any code (the only thing changed is imgui.ini,  and usually in some non-important way (like the window is now five pixels further left).
Also, every time you do want to commit some code changes, you have to keep remembering to not commit imgui.ini

I do not think that imgui.ini should be under version control.

#### To Reproduce
(see description, above)

#### Expected behavior
imgui.ini not version controlled
Hazelnut should load imgui.ini if it exists (so that users window layout is still persistent), otherwise fallback to some sensible default window layout.

#### Screenshots
If applicable, add extra screenshots to help explain your problem.

- OS: win-64
- Version: Windows 10 2004 pro

freeman40:

Issues this solves       | #287

This fix removes imgui.ini from version control.
Hazelnut configuration is still loaded from imgui.ini if that file exists (so user changes to window position and layout will still be persisted from one session to the next.
If imgui.ini does not exist, then a default window position and layout is loaded from new version controlled file "EditorImGui.ini"

#### Additional context
I have tested this on the platform that is available to me (win10)
I cannot think of any obvious reason that this would not work on other platforms, as I have only used "standard" api's that are available with ImGUI


LovelySanta:
Closing this as explained in https://github.com/TheCherno/Hazel/issues/287#issuecomment-663891287

PR #292, opened 2020-07-30
Entt::View camera iteration

dashandslash:
This PR shows an alternative to use EnTT for fast iteration over entites/components.
By avoiding fetching components per each entity it's possible to have a better usage of L1/L2 caches, plus less code ;)

Thanks for your video series, very inspiring and intertainig 🙏

### DIFF
--- a/Hazel/src/Hazel/Scene/Scene.cpp
+++ b/Hazel/src/Hazel/Scene/Scene.cpp
@@ -79,6 +79,7 @@ namespace Hazel {
 						Renderer2D::DrawQuad(transformComp.Transform, spriteComp.Color);
 					});
 					Renderer2D::EndScene();
+					return;
 				}
 			});
 	}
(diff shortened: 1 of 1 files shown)


LovelySanta:
Cherno did cover this PR partially in his livestream that happened today (about 12 hours ago). We concluded that entt is more complicated that it was, with extra features that groups have. Thanks to this PR he discovered the group.each() function that you use twice here. After that we did a deep dive into the entt wiki about views and groups and did some quick and dirty testing in Hazel.

The (only) final conclusions we made are listed below:
1. We want to keep the main camera as a variable outside the lambda, as we'll need it later on as well for other things. So for this reason cherno wanted to assign the camera using the lambda. So the nested .each function is a no go.
2. In the video where he wrote this code, he had issues creating 2 groups, hence why the first one was converted to a view instead. We figured out how this was an issue and how to solve it
3. Due to the 2 conclusions above, cherno said to update the PR to the latest commit (from last video in particular), but that I would not merge the PR as he wants to include it in a future video (soon).

LovelySanta:
Closing this, cherno did review this a while ago on stream, we even took a deep dive into entt go get to the bottom of this.

PR #293, opened 2020-07-31
Improvements in Input system

SentientCoffee:
This PR solves spdlog warnings about C++17 deprecating `std::result_of`.
This PR will improve existing `Input` system by allowing users to compare directly to `uint16_t`s rather than to `enum class`es

Other PRs this solves    | #171 (which has been closed)

This PR improves the `Input` system by allowing the KeyCodes and the MouseCodes to be compared directly to integer types.
Removes a lot of static casting required to use enum classes, as well as the `std::ostream` functions required to log them.
Also changed `Input::GetMousePosition()` to return a `glm::vec2` rather than an `std::pair<float, float>`.

Additionally, the C4996 warning about `std::result_of` being deprecated in C++17 was fixed with a preprocessor define in the `Hazel` and `Sandbox` premake definitions.

#### Additional context
Tested on Windows 10 64-bit.

### DIFF
--- a/Hazelnut/imgui.ini
+++ b/Hazelnut/imgui.ini
@@ -21,7 +21,7 @@ Collapsed=0
 DockId=0x00000002,0
 
 [Docking][Data]
-DockSpace   ID=0x3BC79352 Window=0x4647B76E Pos=320,351 Size=1280,701 Split=X Selected=0x995B0CF8
+DockSpace   ID=0x3BC79352 Window=0x4647B76E Pos=86,128 Size=1280,701 Split=X Selected=0x995B0CF8
   DockNode  ID=0x00000001 Parent=0x3BC79352 SizeRef=426,701 Selected=0x1C33C293
   DockNode  ID=0x00000002 Parent=0x3BC79352 SizeRef=852,701 CentralNode=1 HiddenTabBar=1 Selected=0x995B0CF8
 
--- a/Hazelnut/src/EditorLayer.cpp
+++ b/Hazelnut/src/EditorLayer.cpp
@@ -180,7 +180,7 @@ namespace Hazel {
 		m_ViewportSize = { viewportPanelSize.x, viewportPanelSize.y };
 
 		uint32_t textureID = m_Framebuffer->GetColorAttachmentRendererID();
-		ImGui::Image((void*)textureID, ImVec2{ m_ViewportSize.x, m_ViewportSize.y }, ImVec2{ 0, 1 }, ImVec2{ 1, 0 });
+		ImGui::Image(reinterpret_cast<void*>(textureID), ImVec2{ m_ViewportSize.x, m_ViewportSize.y }, ImVec2{ 0, 1 }, ImVec2{ 1, 0 });
 		ImGui::End();
 		ImGui::PopStyleVar();
 
--- a/Hazel/src/Hazel/Core/Input.h
+++ b/Hazel/src/Hazel/Core/Input.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <glm/glm.hpp>
+
 #include "Hazel/Core/Base.h"
 #include "Hazel/Core/KeyCodes.h"
 #include "Hazel/Core/MouseCodes.h"
@@ -9,10 +11,10 @@ namespace Hazel {
 	class Input
 	{
 	public:
-		static bool IsKeyPressed(KeyCode key);
+		static bool IsKeyPressed(uint16_t key);
 
-		static bool IsMouseButtonPressed(MouseCode button);
-		static std::pair<float, float> GetMousePosition();
+		static bool IsMouseButtonPressed(uint16_t button);
+		static glm::vec2 GetMousePosition();
 		static float GetMouseX();
 		static float GetMouseY();
 	};
--- a/Hazel/src/Hazel/Core/KeyCodes.h
+++ b/Hazel/src/Hazel/Core/KeyCodes.h
@@ -2,146 +2,145 @@
 
 namespace Hazel
 {
-	typedef enum class KeyCode : uint16_t
+	namespace KeyCode
 	{
-		// From glfw3.h
-		Space               = 32,
-		Apostrophe          = 39, /* ' */
-		Comma               = 44, /* , */
-		Minus               = 45, /* - */
-		Period              = 46, /* . */
-		Slash               = 47, /* / */
+		enum : uint16_t
+		{
+			// From glfw3.h
+			Space               = 32,
+			Apostrophe          = 39, /* ' */
+			Comma               = 44, /* , */
+			Minus               = 45, /* - */
+			Period              = 46, /* . */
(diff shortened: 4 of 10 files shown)


LovelySanta:
- I do like the fact that the type is `KeyCode` and `MouseCode`. So a typedef for those could be handy, while maintaining the ability to compare them.
- The namespace can be named `Key` and `Mouse` like the typedefs had

So it would look something like this:
```cpp
namespace Hazel
{
	typedef uint16_t KeyCode;
	
	namespace Key
	{
		enum : KeyCode
		{
			// From glfw3.h
			Button0                = 0,
			...
		}
	}
}
```

This way we can leave the type `KeyCode` as type in the functions, and we can still address keys as `Key::Button0`. The same goes for the `MouseCode` with `Mouse::ButtonLeft`.

What do you think?

Another note, Cherno mentioned to delete all the macros. This could also be part of this PR?

SentientCoffee:
I like your idea, I'll fix that when I have a chance.
Also by the macros, I'm assuming you mean all the `HZ_KEY` macros at the bottom?

SentientCoffee:
I went ahead and implemented the changes. Removing the macros meant changing the `OrthographicCameraController.cpp` file since it used some of the `HZ_KEY` macros, although I don't know how much longer that file will survive lol.

Also added the `_SILENCE_CXX17_RESULT_OF_DEPRECATION_WARNING` to the Hazelnut project as well, since a warning cropped up there too about it.

Hazelnut/src/EditorLayer.cpp
```diff
@@ -179,8 +179,8 @@ namespace Hazel {
 		ImVec2 viewportPanelSize = ImGui::GetContentRegionAvail();
 		m_ViewportSize = { viewportPanelSize.x, viewportPanelSize.y };
 
-		uint32_t textureID = m_Framebuffer->GetColorAttachmentRendererID();
-		ImGui::Image((void*)textureID, ImVec2{ m_ViewportSize.x, m_ViewportSize.y }, ImVec2{ 0, 1 }, ImVec2{ 1, 0 });
+		uint64_t textureID = m_Framebuffer->GetColorAttachmentRendererID();
```
LovelySanta, line 182: `m_Framebuffer->GetColorAttachmentRendererID()` returns a `uint32_t`, shouldn't we at least explicitly cast to `uint64_t`?
https://github.com/TheCherno/Hazel/blob/13b0353ad66172e38767629e36eb1bdb72caf2d0/Hazel/src/Hazel/Renderer/Framebuffer.h#L26

It does look weird on this particular line, I feel we should leave this as a uint32_t and only cast to uint64_t while casting it to a void pointer...

Hazelnut/imgui.ini
```diff
@@ -21,7 +21,7 @@ Collapsed=0
 DockId=0x00000002,0
 
 [Docking][Data]
-DockSpace   ID=0x3BC79352 Window=0x4647B76E Pos=320,351 Size=1280,701 Split=X Selected=0x995B0CF8
+DockSpace   ID=0x3BC79352 Window=0x4647B76E Pos=190,232 Size=1280,701 Split=X Selected=0x995B0CF8
```
LovelySanta, line 24: Do not commit any `imgui.ini` changes

LovelySanta:
> Also by the macros, I'm assuming you mean all the `HZ_KEY` macros at the bottom?

That's indeed what I meant, you did the changes as I intended 👍 

> I went ahead and implemented the changes. Removing the macros meant changing the `OrthographicCameraController.cpp` file since it used some of the `HZ_KEY` macros, although I don't know how much longer that file will survive lol.

Yeah, by removing the macros, it had to be changed there as well. I assume this CameraController would still stay here as the editor camera (as it's not part of the scene itself). Cherno had some mixed feelings in the video where he discussed this. However, didn't hurt to update the file anyway.

> Also added the `_SILENCE_CXX17_RESULT_OF_DEPRECATION_WARNING` to the Hazelnut project as well, since a warning cropped up there too about it.

I guess that's indeed one way of doing it... Not sure if we want to do this for the whole engine versus only the vendors... I guess this is fine for now... It's more an annoyance at this point indeed.

Hazelnut/src/EditorLayer.cpp
SentientCoffee: I made it there because doing a `reinterpret_cast<void*>(static_cast<uint64_t>(textureID));` looked way like there was way too much going on for just casting to void*. I guess we could also put it on a separate line or `static_cast` the `GetColorAttachmentRendererID()` call, but I thought this was cleaner since there's no reason a 64-bit uint container can't hold a 32-bit uint value in any circumstance.

Hazelnut/imgui.ini
SentientCoffee: Should I fix the commit on my end or is there a way to discard specific file changes on your end? I've never had to manage pull requests myself so this is more out of curiosity for me.

Also, shouldn't `imgui.ini` be gitignored completely? Or does it need to be there for the Hazelnut UI to show up properly for other users?

SentientCoffee:
> I guess that's indeed one way of doing it... Not sure if we want to do this for the whole engine versus only the vendors... I guess this is fine for now... It's more an annoyance at this point indeed.
It was just to silence an irrelevant warning, so idk if it's strictly necessary. There's also a `_SILENCE_ALL_CXX17_DEPRECATION_WARNINGS` macro you could use instead if you wanted to take care of more warnings that may show up in the future, but I decided to just deal with the specific warning for now to avoid not getting possibly relevant warnings later

Hazelnut/imgui.ini
LovelySanta: > Should I fix the commit on my end or is there a way to discard specific file changes on your end? I've never had to manage pull requests myself so this is more out of curiosity for me.

There are several ways of resetting it. You can reset the file to a specific commit, or you could just commit the change to it manually such that the old state in the repo is equal to the new state in this PR. As I'm a collaborator I can also push commits to your branch, but I prefer not to (only minor details before committing). If I commit changes it will appear just the same as you did it, in a commit on your branch. That commit will then update this PR as it does for any other commit.

> Also, shouldn't `imgui.ini` be gitignored completely? Or does it need to be there for the Hazelnut UI to show up properly for other users?

There were numerous issues/PRs already to add it to gitignore. I refer to https://github.com/TheCherno/Hazel/pull/204#issuecomment-583718303 for the reason why it's not in gitignore. In the future we should add some system in place to load preferences, and at that point we should provide (manually) a default, if it doesn't exist.

Hazelnut/src/EditorLayer.cpp
LovelySanta: yeah, I guess we can live with it...

SentientCoffee:
I restored the imgui.ini file back to its original state before my pull request. Let me know if there's anything else I should change

LovelySanta:
Looks good to me 👍

PR #296, opened 2020-08-03
Closing ifstream both if its readable and unreadable

Logstor:
In OpenGLShader.cpp method ReadFile() ifstream isn't closed if we're unable to read from it.

#### PR impact

None

#### Proposed fix 

### DIFF
--- a/Hazel/src/Platform/OpenGL/OpenGLShader.cpp
+++ b/Hazel/src/Platform/OpenGL/OpenGLShader.cpp
@@ -68,12 +68,13 @@ namespace Hazel {
 				result.resize(size);
 				in.seekg(0, std::ios::beg);
 				in.read(&result[0], size);
-				in.close();
 			}
 			else
 			{
 				HZ_CORE_ERROR("Could not read from file '{0}'", filepath);
 			}
+
+			in.close();
 		}
 		else
 		{


ReDucTor:
Both instances of close seem unnecessary, RAII will handle this fine

Logstor:
> Both instances of close seem unnecessary, RAII will handle this fine

So RAII will clean up, when the stream goes out of scope?

Logstor:
ifstream.close() is now removed, and comment added for clarity.

ReDucTor:
I feel the comment is unnecessary,  RAII is one of the basics of c++ so doesn't really need to be said.

Logstor:
> I feel the comment is unnecessary, RAII is one of the basics of c++ so doesn't really need to be said.

Good point, but as this version of Hazel also is for the community to learn, so wouldn't it be nice to have?

LovelySanta:
Well, yes RAII will take care of it, but will it still work 'fine' when I load let's say a big amount of shaders all after each other? Taking all the memory into account that this would consume during them being open? I don't expect that they will be closed before the next one will open if you leave it up to resource manager?

Either way, it doesn't feel like an overhead to close them explicitly? Maybe I am missing something?

Logstor:
LovelySanta, I think RAII makes sure that the destructor is being run when the end of it's scope is reached. Only if you need to catch any errors when closing the stream, then you probably need to close it explicitly.

https://stackoverflow.com/questions/748014/do-i-need-to-manually-close-an-ifstream

PR #297, opened 2020-08-04
Remove redundant .get() in Renderer2D

saffronjam:
#### Issue
Minor redundant code impacts the sections readability

#### Proposed fix
Remove ".get()" and make use of smart pointers overloaded operators

### DIFF
--- a/Hazel/src/Hazel/Renderer/Renderer2D.cpp
+++ b/Hazel/src/Hazel/Renderer/Renderer2D.cpp
@@ -241,7 +241,7 @@ namespace Hazel {
 		float textureIndex = 0.0f;
 		for (uint32_t i = 1; i < s_Data.TextureSlotIndex; i++)
 		{
-			if (*s_Data.TextureSlots[i].get() == *texture.get())
+			if (*s_Data.TextureSlots[i] == *texture)
 			{
 				textureIndex = (float)i;
 				break;


Hazel/src/Hazel/Renderer/Renderer2D.cpp
```diff
@@ -241,7 +241,7 @@ namespace Hazel {
 		float textureIndex = 0.0f;
 		for (uint32_t i = 1; i < s_Data.TextureSlotIndex; i++)
```
LovelySanta, line 242: While we're here...
```suggestion
		for (size_t i = 1; i < s_Data.TextureSlotIndex; i++)
```

saffronjam: Made that commit quickly, without asking, Why are we doing this? It seems odd to me to add a size_t when the bound is 32-bit anyway.

LovelySanta: well, after having a closer look... the `s_Data.TextureSlotIndex` is indeed a `uint32_t` (which I didn't notice before)...
My thought was since indexing uses a `size_t` this could prevent an implicit cast.. Now I realise it's just moving the cast from the indexing operator to the comparison of `i < s_Data.TextureSlotIndex`...

Now I'm not sure what would be best... Maybe we should leave it like it was as that would give a better indication to the reader... Sorry, was my bad (was too late when I wrote it)...

saffronjam: "Now I realise it's just moving the cast from the indexing operator to the comparison of i < s_Data.TextureSlotIndex" is the reason I got confused! But I think it's best to undo the change this time to keep the PR clear. 

I guess it's good to keep an eye out for implicit conversions though, so +1 from that perspective.

PR #302, opened 2020-08-12
Partial Xlib platform support.

shepard47:
Hi,
I've been making a game engine and I thought that I could also write some code for Hazel. The part I am excited about is Linux/Xlib implementation (XCB is probably better but doesn't have OpenGL support). 
I know it's not a requested feature but I would be very glad if my code would be usable.

What it does:
XlibWindow::init creates an Xlib window and creates an OpenGL 3.3 context for it (omitting OpenGLContext::CreateContext and OpenGLContext::init).

However, it's cannot be built right know because there is an Xlib type Window that is the same as the Hazel name and I think the C++ compiler doesn't like it (I'm sure the C compiler doesn't).

If this code makes sense (probably not) it would be easy to extend it with event handling.

### DIFF
--- a/Hazel/src/Platform/Linux-Xlib/XlibInput.cpp
+++ b/Hazel/src/Platform/Linux-Xlib/XlibInput.cpp
@@ -1,0 +1,43 @@
+#include "hzpch.h"
+#include "Hazel/Core/Input.h"
+
+#include "Hazel/Core/Application.h"
+
+namespace Hazel {
+
+	bool Input::IsKeyPressed(KeyCode key)
+	{
+		auto window = static_cast<int*>(Application::Get().GetWindow().GetNativeWindow());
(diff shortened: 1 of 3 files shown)


ShazerHD:
To be honest, this doesn't make sense to impliment XLib support from scratch as TheCherno is already using GLFW which is platform-agnostic so unless you are going to rewrite the windowing system to not use GLFW, then this makes no sense to use.

shepard47:
@ShazerHD I supposed GLFW in Hazel is temporal.

ShazerHD:
Cherno has said on stream that he has no plans to replace/remove GLFW so it isn't temporary.

LovelySanta:
As @sfulham said, we're using GLFW, so this is not really of any use to Hazel.

PR #304, opened 2020-08-13
Fixed NativeScriptComponent pessimization and unintentional incorrect conversion

### LINKED ISSUE #303: Unnecessary overhead in native script function binding
I see that Cherno is using hand-crafted type erasure for native script (to get rid of one pointer de-referencing of the vtable), but the use of `std::function` is totally unnecessary and can actually be avoided by using regular function pointers.

`std::function` has more overhead (space-wise and time-wise) compared with function pointers. Since the five lambdas don't have any large members (three of them are capture-less, and two of them only hold a reference which is the same size as a pointer) they can be stored in the small buffers. However, common implementation of `std::function` come with a small buffer capable of hold 2 `std::string` objects, compared to the lambdas in question that's much space wasted. And for debug build performance, calling through `std::function`'s layers of indirection is a pure pessimization, although it usually doesn't affect release builds, but why pay for something we don't use?

For this issue I suggest replacing the `std::function`'s with function pointers. The modification is straightforward as well, since non-capturing lambdas can be implicitly converted into function pointers, and although the "instantiate" function and the "destroy" function do have captures, it is easy to just add another parameter in for the pointer and pass in the pointer later.

I can send a pull request later if needed.

Chlorie:
## Describe the issue
See #303. Also there're two places in "Scene.cpp" where unintentional incorrect rvalue to non-const lvalue conversions have been made. More on this issue later.

## PR impact

Issues this solves       | Resolves #303

## Proposed fix
NativeScriptComponent uses `std::function` which introduces unnecessary overhead. This PR fixes it by replacing the `std::function` objects with function pointers.

The second commit fixes a compile error under "/permissive-" in MSVC.

## Additional context
In Line 90 and 108 of Scene.cpp, I think Cherno's intention was to unpack the result using structured binding into two lvalue references, thus the `auto&`.
https://github.com/TheCherno/Hazel/blob/3177aca930a1cdd74bad082a80fb2fbfa27429a2/Hazel/src/Hazel/Scene/Scene.cpp#L90 https://github.com/TheCherno/Hazel/blob/3177aca930a1cdd74bad082a80fb2fbfa27429a2/Hazel/src/Hazel/Scene/Scene.cpp#L108

But actually that `auto&` corresponds to the return value, which is a `std::tuple<TransformComponent&, CameraComponent&>`, and since you cannot assign a return value into a non-const lvalue reference this should fail to compile (see [this cppreference link](https://en.cppreference.com/w/cpp/language/structured_binding#Notes) for detailed explanation).

MSVC compiles fine because it has an extension to make this not an error, but this is obviously non-portable (and not useful either). The original intention here would be "store the result tuple here and then unpack the two references", thus `auto` without the ampersand should be used here. (Also `const auto&` can be used here as well but why that works relates to lifetime extension rules which is another can of worms...).

I also would like to suggest to turn "/permissive-" on in MSVC builds to prevent this kind of problems, but I'm not a premake expert so I don't know the exact changes to make in the build script...

EDIT by @LovelySanta: Added a little help since @Chlorie didn't know how to get code permalinks working in a PR. (You have to link the code line at a specific commit and not just at the head of a branch.)

### DIFF
--- a/Hazel/src/Hazel/Scene/Components.h
+++ b/Hazel/src/Hazel/Scene/Components.h
@@ -54,22 +54,22 @@ namespace Hazel {
 	{
 		ScriptableEntity* Instance = nullptr;
 
-		std::function<void()> InstantiateFunction;
-		std::function<void()> DestroyInstanceFunction;
+		void(* InstantiateFunction)(ScriptableEntity*&);
+		void(* DestroyInstanceFunction)(ScriptableEntity*&);
 
-		std::function<void(ScriptableEntity*)> OnCreateFunction;
-		std::function<void(ScriptableEntity*)> OnDestroyFunction;
-		std::function<void(ScriptableEntity*, Timestep)> OnUpdateFunction;
+		void(* OnCreateFunction)(ScriptableEntity*);
+		void(* OnDestroyFunction)(ScriptableEntity*);
+		void(* OnUpdateFunction)(ScriptableEntity*, Timestep);
 
 		template<typename T>
 		void Bind()
 		{
-			InstantiateFunction = [&]() { Instance = new T(); };
-			DestroyInstanceFunction = [&]() { delete (T*)Instance; Instance = nullptr; };
+			InstantiateFunction = [](ScriptableEntity*& instance) { instance = new T(); };
+			DestroyInstanceFunction = [](ScriptableEntity*& instance) { delete static_cast<T*>(instance); instance = nullptr; };
 
-			OnCreateFunction = [](ScriptableEntity* instance) { ((T*)instance)->OnCreate(); };
-			OnDestroyFunction = [](ScriptableEntity* instance) { ((T*)instance)->OnDestroy(); };
-			OnUpdateFunction = [](ScriptableEntity* instance, Timestep ts) { ((T*)instance)->OnUpdate(ts); };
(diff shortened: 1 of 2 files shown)


LovelySanta:
In discord another issue came up when trying to add multiple script components to different entities (@Fiixel in Discord). The issue seems to be that `ncs->Instance` is still nullptr for the second component... According to @9YO (Discord) this PR does fix this. 

image
image

Chlorie:
@LovelySanta I've been thinking of why this PR fixed the issue you mentioned for a little while, and my suspicion is: The lambda captured `this` pointer as expected, and the lambda got saved into the `std::function` object. It's totally OK until another component got inserted into the registry and triggers a buffer expansion, then the old entities are moved into the new buffer, invalidating the `this` pointer stored in the lambda. After all these, `Scene::OnUpdate` finally get called, but `InstantiateFunction` write the pointer into garbage memory, leaving the current valid pointer null.

freeman40:
The following incantation somewhere near the top of the premake scripts will turn off MSVC permissive mode:
```
configuration "vs*"
   buildoptions { "/permissive-" }
```

LovelySanta:
Cherno did change the ScriptableEntity to use virtual functions in https://github.com/TheCherno/Hazel/commit/a54685681b7ac2a57442acd1e0d0ddf38589b17f. He also did change the std::functions to fuction pointers in the NativeScriptComponent. This means that this PR is now obsolete, so it can be closed.

PR #305, opened 2020-08-13
Added #undef min and #undef max to pch file

CybernetHacker14:
While compiling the project, especially on Windows, a bunch of compilation errors turn up. A quick check on the internet revealed conflicts with the Windows.h header library and std::max and std::min functions being used (in this case the entt library is using std::max)

#### PR impact

#### Proposed fix
Added #undef min and #undef max in the pre-compiled header. (It could be anywhere, for this fix I have added it in the pch file
just above #include <Windows.h> line)

### DIFF
--- a/Hazel/src/hzpch.h
+++ b/Hazel/src/hzpch.h
@@ -18,5 +18,7 @@
 #include "Hazel/Debug/Instrumentor.h"
 
 #ifdef HZ_PLATFORM_WINDOWS
+	#undef min
+	#undef max
 	#include <Windows.h>
 #endif


saffronjam:
What about `#define NOMINMAX` instead of two udefs?

CybernetHacker14:
> What about `#define NOMINMAX` instead of two undefs?

I first used #define NOMINMAX, but it was still giving me errors. Furthermore, I have kept both undefs separate, in case any-one wants to, at their disposal, remove one of them.

LovelySanta:
Imgui is actually [suggesting](https://github.com/skypjack/entt/wiki/Frequently-Asked-Questions#warning-c4003-the-min-the-max-and-the-macro) to use the `#define NOMINMAX` and I do know that solves the issue for them. After defining these, do you still get the same errors as in the [entt issue](https://github.com/skypjack/entt/issues/96) that describes this problem?

CybernetHacker14:
> Imgui is actually [suggesting](https://github.com/skypjack/entt/wiki/Frequently-Asked-Questions#warning-c4003-the-min-the-max-and-the-macro) to use the `#define NOMINMAX` and I do know that solves the issue for them. After defining these, do you still get the same errors as in the [entt issue](https://github.com/skypjack/entt/issues/96) that describes this problem?
> 
> However, the [skypjack suggested](https://github.com/skypjack/entt/issues/96#issuecomment-395867237) to put it all the way on top, before any other includes. So maybe we should define it on top as well? Can you check if this solves it? Also adding a reference to the [entt wiki section](https://github.com/skypjack/entt/wiki/Frequently-Asked-Questions#warning-c4003-the-min-the-max-and-the-macro) could be a good idea?

So `#define NOMINMAX` is working when I do that outside the `#ifdef HZ_PLATFORM_WINDOW` block before all includes. But should that define exist outside that block? I thought it was a `Windows.h` specific bug only and should be contained in that specific block only.

LovelySanta:
Yes it should only do that for windows only, no one said you could not include that... But if you look at the code below, `HZ_PLATFORM_WINDOWS` only gets defined inside `Base.h` which is included inside `Log.h`. A solution would be to include `Base.h` inside the pch all the way at the top as well? 
```cpp
#pragma once

// include Base.h for HZ_PLATFORM_WINDOWS definition?
#ifdef HZ_PLATFORM_WINDOWS
	#ifndef NOMINMAX
		#define NOMINMAX
	#endif
#endif

#include ...

#ifdef HZ_PLATFORM_WINDOWS
	#include <Windows.h>
#endif
```

I would not do that and rather split off the `PlatformDetection.h` into a seperate file (inside Hazel/Core folder), and include that one at the top of the pch (and also in `Base.h`)

Also I did notice that the `HZ_PLATFORM_WINDOWS` might not have been compiled...

CybernetHacker14:
On @LovelySanta 's suggestion, I have added a new file `PlatformDetection.h` , and his implemented his approach.

LovelySanta:
Does look good like you added it, but can you add indentations to the macro's? That's how we do it in Hazel atm

Hazel/src/Hazel/Core/Base.h
```diff
@@ -2,76 +2,36 @@
...
-	#define HZ_ENABLE_ASSERTS
+#if defined(HZ_PLATFORM_WINDOWS)
+#define HZ_DEBUGBREAK() __debugbreak()
+#elif defined(HZ_PLATFORM_LINUX)
+#include <signal.h>
+#define HZ_DEBUGBREAK() raise(SIGTRAP)
+#else
+#error "Platform doesn't support debugbreak yet!"
+#endif
+#define HZ_ENABLE_ASSERTS
```
LovelySanta, line 16:
```suggestion
	#if defined(HZ_PLATFORM_WINDOWS)
		#define HZ_DEBUGBREAK() __debugbreak()
	#elif defined(HZ_PLATFORM_LINUX)
		#include <signal.h>
		#define HZ_DEBUGBREAK() raise(SIGTRAP)
	#else
		#error "Platform doesn't support debugbreak yet!"
	#endif
	#define HZ_ENABLE_ASSERTS
```

Hazel/src/Hazel/Core/Base.h
```diff
@@ -2,76 +2,36 @@
...
 #else
-	#define HZ_DEBUGBREAK()
+#define HZ_DEBUGBREAK()
```
LovelySanta, line 18:
```suggestion
	#define HZ_DEBUGBREAK()
```

Hazel/src/Hazel/Core/Base.h
LovelySanta, line 16:
```suggestion
	#if defined(HZ_PLATFORM_WINDOWS)
		#define HZ_DEBUGBREAK() __debugbreak()
	#elif defined(HZ_PLATFORM_LINUX)
		#include <signal.h>
		#define HZ_DEBUGBREAK() raise(SIGTRAP)
	#else
		#error "Platform doesn't support debugbreak yet!"
	#endif
	#define HZ_ENABLE_ASSERTS
```

Hazel/src/Hazel/Core/Base.h
```diff
@@ -2,76 +2,36 @@
...
 #endif
 
 // TODO: Make this macro able to take in no arguments except condition
 #ifdef HZ_ENABLE_ASSERTS
-	#define HZ_ASSERT(x, ...) { if(!(x)) { HZ_ERROR("Assertion Failed: {0}", __VA_ARGS__); HZ_DEBUGBREAK(); } }
-	#define HZ_CORE_ASSERT(x, ...) { if(!(x)) { HZ_CORE_ERROR("Assertion Failed: {0}", __VA_ARGS__); HZ_DEBUGBREAK(); } }
+#define HZ_ASSERT(x, ...) { if(!(x)) { HZ_ERROR("Assertion Failed: {0}", __VA_ARGS__); HZ_DEBUGBREAK(); } }
+#define HZ_CORE_ASSERT(x, ...) { if(!(x)) { HZ_CORE_ERROR("Assertion Failed: {0}", __VA_ARGS__); HZ_DEBUGBREAK(); } }
```
LovelySanta, line 24:
```suggestion
	#define HZ_ASSERT(x, ...) { if(!(x)) { HZ_ERROR("Assertion Failed: {0}", __VA_ARGS__); HZ_DEBUGBREAK(); } }
	#define HZ_CORE_ASSERT(x, ...) { if(!(x)) { HZ_CORE_ERROR("Assertion Failed: {0}", __VA_ARGS__); HZ_DEBUGBREAK(); } }
```

Hazel/src/Hazel/Core/Base.h
```diff
@@ -2,76 +2,36 @@
...
 #else
-	#define HZ_ASSERT(x, ...)
-	#define HZ_CORE_ASSERT(x, ...)
+#define HZ_ASSERT(x, ...)
+#define HZ_CORE_ASSERT(x, ...)
```
LovelySanta, line 27:
```suggestion
	#define HZ_ASSERT(x, ...)
	#define HZ_CORE_ASSERT(x, ...)
```

Hazel/src/Hazel/Core/Base.h
```diff
@@ -2,76 +2,36 @@
...
 #endif
 
 #define BIT(x) (1 << x)
 
 #define HZ_BIND_EVENT_FN(fn) [this](auto&&... args) -> decltype(auto) { return this->fn(std::forward<decltype(args)>(args)...); }
 
 namespace Hazel {
-
 	template<typename T>
```
LovelySanta, line 35: We always leave an empty line after the Hazel namespace
```suggestion
namespace Hazel {

	template<typename T>
```

Hazel/src/Hazel/Core/Base.h
```diff
@@ -87,5 +47,4 @@ namespace Hazel {
 	{
 		return std::make_shared<T>(std::forward<Args>(args)...);
 	}
-
 }
```
LovelySanta, line 50: Similarly at the closing bracket of the namespace we have an empty line
```suggestion
	}

}
```

Hazel/src/Hazel/Core/PlatformDetection.h
```diff
@@ -0,0 +1,42 @@
...
+#elif defined(__ANDROID__)
+#define HZ_PLATFORM_ANDROID
+#error "Android is not supported!"
+#elif defined(__linux__)
+#define HZ_PLATFORM_LINUX
+#error "Linux is not supported!"
+#else
+	/* Unknown compiler/platform */
+#error "Unknown platform!"
+#endif // End of platform detection
```
LovelySanta, line 42:
```suggestion
// Platform detection using predefined macros
#ifdef _WIN32
	/* Windows x64/x86 */
	#ifdef _WIN64
		/* Windows x64  */
		#define HZ_PLATFORM_WINDOWS
	#else
		/* Windows x86 */
		#error "x86 Builds are not supported!"
	#endif
#elif defined(__APPLE__) || defined(__MACH__)
	#include <TargetConditionals.h>
	/* TARGET_OS_MAC exists on all the platforms
	 * so we must check all of them (in this order)
	 * to ensure that we're running on MAC
	 * and not some other Apple platform */
	#if TARGET_IPHONE_SIMULATOR == 1
		#error "IOS simulator is not supported!"
	#elif TARGET_OS_IPHONE == 1
		#define HZ_PLATFORM_IOS
		#error "IOS is not supported!"
	#elif TARGET_OS_MAC == 1
		#define HZ_PLATFORM_MACOS
		#error "MacOS is not supported!"
	#else
		#error "Unknown Apple platform!"
	#endif
/* We also have to check __ANDROID__ before __linux__
 * since android is based on the linux kernel
 * it has __linux__ defined */
#elif defined(__ANDROID__)
	#define HZ_PLATFORM_ANDROID
	#error "Android is not supported!"
#elif defined(__linux__)
	#define HZ_PLATFORM_LINUX
	#error "Linux is not supported!"
#else
	/* Unknown compiler/platform */
	#error "Unknown platform!"
#endif // End of platform detection
```

Hazel/src/hzpch.h
```diff
@@ -1,5 +1,13 @@
 #pragma once
 
+#include "Hazel/Core/PlatformDetection.h"
+
+#ifdef HZ_PLATFORM_WINDOWS
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+#endif
```
LovelySanta, line 9:
```suggestion
#ifdef HZ_PLATFORM_WINDOWS
	#ifndef NOMINMAX
		// See github.com/skypjack/entt/wiki/Frequently-Asked-Questions#warning-c4003-the-min-the-max-and-the-macro
		#define NOMINMAX
	#endif
#endif
```

Hazel/src/hzpch.h
```diff
@@ -18,5 +26,5 @@
 #include "Hazel/Debug/Instrumentor.h"
 
 #ifdef HZ_PLATFORM_WINDOWS
-	#include <Windows.h>
+#include <Windows.h>
```
LovelySanta, line 29:
```suggestion
	#include <Windows.h>
```

CybernetHacker14:
> Does look good like you added it, but can you add indentations back to the macro's? That's how we do it in Hazel atm, and it helps for readability. And with my comment of adding a reference to the entt wiki section I meant in the code, but also good that you added it to the PR.

Yeah will do that.

Issue #289, opened 2020-07-28
Advance HZ_ASSERT and HZ_CORE_ASSERT macros to take in no arguments except condition


alohaeee:
Current asserts macros could't provide a call without user message. 
![image](https://user-images.githubusercontent.com/45694435/88662081-20d15f00-d0e2-11ea-93ea-0476b98a722b.png)
It's noticed in TODO before macros definition: "// TODO: Make this macro able to take in no arguments except condition".

Current implementation also can't help with finding a assert call with no-printable argument. 
![image](https://user-images.githubusercontent.com/45694435/88663034-90941980-d0e3-11ea-8ecc-222b1bf197b9.png)
As in the example above the compiler will note only internal format library error but not a problem call.

### Suggestion
Replace the current assert definition
```cpp
#define HZ_ASSERT(x, ...) { if(!(x)) { HZ_ERROR("Assertion Failed: {0}", __VA_ARGS__); HZ_DEBUGBREAK(); } }
#define HZ_CORE_ASSERT(x, ...) { if(!(x)) { HZ_CORE_ERROR("Assertion Failed: {0}", __VA_ARGS__); HZ_DEBUGBREAK(); } }
```
to new one:
```cpp
#define HZ_ASSERT(x, ...) { if(!(x)) { Hazel::_AssertMessage(__FILE__, __LINE__, __VA_ARGS__); HZ_DEBUGBREAK(); } }
#define HZ_CORE_ASSERT(x, ...) { if(!(x)) { Hazel::_CoreAssertMessage(__FILE__, __LINE__, __VA_ARGS__); HZ_DEBUGBREAK(); } } 
```
Provide a separate file for assert log functions:
_Assert.h_
```cpp
#pragma once

namespace Hazel
{
	void _AssertMessage(const char* file, int line);
					   
	void _AssertMessage(const char* file, int line, const char* msg);

	void _CoreAssertMessage(const char *file, int line);

	void _CoreAssertMessage(const char* file, int line, const char* msg);

}
```
_Assert.cpp_
```cpp
#include "hzpch.h"
#include "Hazel/Core/Assert.h"

#include "Hazel/Core/Log.h"

namespace Hazel
{
	void AssertMessage(const char* file, int line)
	{
		HZ_ERROR("Assertion Failed at file: {0} Line: {1}", file, line);
	}

	void _AssertMessage([[maybe_unused]] const char* file,
		[[maybe_unused]] int line, const char* msg)
	{
[...]

PR #306, opened 2020-08-13
Allow assert macro to take no arguments

### LINKED ISSUE #326: HZ_ERROR macro not found in places it shouldn't be.
#### Describe the bug
In https://github.com/TheCherno/Hazel/commit/857d8d60e63b23c67edf1bb10e2715fb57089597, cherno had compiler issues that it could not find the HZ_ERROR macro. As a result he had to include `Log.h` where he shouldn't have.
https://github.com/TheCherno/Hazel/blob/857d8d60e63b23c67edf1bb10e2715fb57089597/Hazelnut/src/Panels/SceneHierarchyPanel.h#L4

#### Expected behavior
There should be no need to include `Log.h`. `Log.h` Should be included in `Base.h` where it is used:
https://github.com/TheCherno/Hazel/blob/857d8d60e63b23c67edf1bb10e2715fb57089597/Hazel/src/Hazel/Core/Base.h#L22-L28

such that this would not error in the future:
```c++
#include "Hazel/Core/Log.h"
#ifdef HZ_ENABLE_ASSERTS
#else
	#define HZ_ASSERT(x, ...)
	#define HZ_CORE_ASSERT(x, ...)
#endif
```

TigerCipher:
Changed the HZ_ASSERT and HZ_CORE_ASSERT to allow no additional arguments other than the condition (x)

Saw the TODO note in `Base.h` about enabling the assert macros to allow no arguments other than the condition. Simple fix was to prefix `##` to the `__VA_ARGS__`

#### Additional context
To the best of my knowledge the `##` prefix should work on all major C/C++ compilers, but I have tested it myself in my own projects using Visual Studio 2019, both a standard build and an nmake build via the CLion IDE (not 100% certain if nmake uses a different version of the visual C++ compiler or not, though I doubt it would) and with GCC 9.3 on Ubuntu 20. Also a quick apology if I've not followed your pull request template/guidelines properly, this is my first pull request I've made on someone else's project :P

This PR also resolves #326 and closes #327.


LovelySanta:
I've heard cherno talk about some bugs in MSVC, it's not that easy to do I think, otherwise cherno would've done it. Also have a look at #289 maybe?

TigerCipher:
> I've heard cherno talk about some bugs in MSVC, it's not that easy to do I think, otherwise cherno would've done it. Also have a look at #289 maybe?

Could you specify what bugs you are referring to? I've never had any kind of issue or bug with `##__VA_ARGS__` with msvc. The only thing I can think of would be the `HZ_ERROR` and `HZ_CORE_ERROR` macros not having the `##` (which I can't find where those are even defined). Or again like I said in the comment I added in code, if needing that `{0}` token replaced is required (not sure if whatever function does the replacing requires arguments to match the number of tokens) then you'd need to use the macro `HZ_ASSERT(x, n, ...)` instead

Other than those two things, which could easily be fixed as well, I personally can't think of any bugs. Hadn't noticed #289 before, but from what I can see, that issue suggests creating overloaded functions rather than a single macro. Both solve the same problem, but using `##__VA_ARGS__` (and adding that if needed to the error macros) will make things work exactly as before, with the only addition being that the additional arguments would not be required.

When I have a bit more time later on I'll clone hazel and make any needed changes and modify this PR (as at first I hadn't realized the macros called other macros, so I assumed they were functions that already supported optional arguments) (I assume a PR can be modified, or would I need to make a new one?)

LovelySanta:
> Could you specify what bugs you are referring to? I've never had any kind of issue or bug with `##__VA_ARGS__` with msvc. 

It has been a while since i've come across it, for one sometimes it's not expanding the actual macro (when passing in macro as argument), or not recognizing them.. I can't recall the details.. Cherno does mention some MSVC bug with the # and ## operators once in a while as well, so I know there is *something* wrong.

> The only thing I can think of would be the `HZ_ERROR` and `HZ_CORE_ERROR` macros not having the `##` (which I can't find where those are even defined).

They are defined in the `Log.h`  file:
https://github.com/TheCherno/Hazel/blob/3177aca930a1cdd74bad082a80fb2fbfa27429a2/Hazel/src/Hazel/Core/Log.h#L24-L36

> Or again like I said in the comment I added in code, if needing that `{0}` token replaced is required (not sure if whatever function does the replacing requires arguments to match the number of tokens) then you'd need to use the macro `HZ_ASSERT(x, n, ...)` instead

Yes I saw that, and as convention, I would say that `{0}` should be required. And it's actual spdlog that does the replacement.

> Other than those two things, which could easily be fixed as well, I personally can't think of any bugs. Hadn't noticed #289 before, but from what I can see, that issue suggests creating overloaded functions rather than a single macro. Both solve the same problem, but using `##__VA_ARGS__` (and adding that if needed to the error macros) will make things work exactly as before, with the only addition being that the additional arguments would not be required.

I think the most readable solution would be prefered, mainly since it's for coding purposes. I was also thinking, when not passing in any arguments, that it would print a default message like "Assertion failed at `__FILENAME__:__LINE__`, maybe?

> When I have a bit more time later on I'll clone hazel and make any needed changes and modify this PR (as at first I hadn't realized the macros called other macros, so I assumed they were functions that already supported optional arguments) (I assume a PR can be modified, or would I need to make a new one?)

To make a PR you most likely need a fork in order to commit changes. In this case you already created your fork at [TigerCipher/Hazel](https://github.com/TigerCipher/Hazel). In particular this PR is made from a branch called [patch-1](https://github.com/TigerCipher/Hazel/tree/patch-1), which is the default naming.

So you have to clone your fork locally, change to the `patch-1` branch, change the code, commit and push the code. By pushing it will update your branch online, and that will then reflect in this PR.

TigerCipher:
> It has been a while since i've come across it, for one sometimes it's not expanding the actual macro (when passing in macro as argument), or not recognizing them.. I can't recall the details.. Cherno does mention some MSVC bug with the # and ## operators once in a while as well, so I know there is something wrong.

Well without knowing exactly what you're referring to, not too sure what to say. I do know another option is that in new C++ versions there is an extension you can use, I believe it would be something along the lines of `#define MACRO(x, ...) replacement(x __VA_OPT__(,) __VA_ARGS__)` . Might also be good to see if you have the `/experimental:preprocessor ` compiler option set for msvc. That will make the compiler not suppress trailing commas unless the va args are prefixed by `##`. I **believe** the compiler option is set by default in newer versions of visual studio, which is why it should work no problem.  make things a big more clear, you can see some stuff about it on the msvc docs [comma elision](https://docs.microsoft.com/en-us/cpp/preprocessor/preprocessor-experimental-overview?view=vs-2019#comma-elision-in-variadic-macros) and of course if this still doesn't solve whatever `#` bugs cherno had mentioned before, then you can always ensure that compiler option is not set, then msvc will always remove trailing commas. In other words, assuming you have the compiler option on and are not using `##` then the macro would be replaced by `replacement(x, )` when no arguments are given, which would be an error because of that comma. But with the `##` the compiler option turns it into `replacement(x)` and the world is happy.

> I think the most readable solution would be prefered, mainly since it's for coding purposes. I was also thinking, when not passing in any arguments, that it would print a default message like "Assertion failed at __FILENAME__:__LINE__, maybe?

If you wanted it to be that with no arguments, but then just `"Assertion failed: {0}"` or whatever it currently is for when there are arguments, you would have to either use two separate macros or like that one issue states, use a function to determine how many arguments were passed in (and you'd have to use template variadic arguments in that case). IMO the "most readable" solution would be to either have no filename/line at all, or make it all have filename/line (though I'd assume the log outputs file/lines already?).

I would like to quickly point out, that the more I've thought about it, I've realized the fact that formatting is being used with an initial default message ("Assertion failed {0}") will actually be an issue. I began doing some testing on one of my projects so I'll just use that code as a proof of concept for an alternative option one could use.

```
#define LINE_NUMBER(l) #l
#define FILE_LINE(f, l) f ":" LINE_NUMBER(l)
#define FMT(fmt_, ...) fmt_, ##__VA_ARGS__
#define dbg(x, ...) if(!x) printf("Error in " FILE_LINE(__FILE__, __LINE__) FMT(__VA_ARGS__))

void TestListener::onEvent(const std::shared_ptr<Event>& evt)
{
	if (evt->getType() == TEST)
	{
		dbg(false); // expands to if(!false) printf("Error in testlistener.cpp:35")  (or whatever file/line you use it in
		dbg(false, "%s %i", "Something", 32); // expands to printf("Error in testlistener.cpp:36 %s %i", "something", 32); (might have spaces wrong)
// of course, for the log library you'd use {} instead
		std::shared_ptr<TestEvent> e = (const std::shared_ptr<TestEvent>&) evt;
		printf("Event data: %s\n", e->getData().c_str());
	}
}
```


If that right there is acceptable, I can modify the PR later tonight

LovelySanta:
Does the trailing comma also gets removed on other compilers, for example when we want to compile for linux or maxOS, which doesn't use the MSVC?

Also the default message should be replaced by the custom message instead of just concatenating?

LovelySanta:
I just did some quick test. This is what I came up with at this point:

```cpp
#define HZ_EXPAND_MACRO(x) x

#ifdef HZ_ENABLE_ASSERTS
	#define HZ_INTERNAL_ASSERT_WITH_MSG(assert_type, x, ...) { if(!(x)) { ##assert_type##_ERROR("Assertion Failed: {0}", __VA_ARGS__); HZ_DEBUGBREAK(); } }
	#define HZ_INTERNAL_ASSERT_NO_MSG(assert_type, x) HZ_INTERNAL_ASSERT_WITH_MSG(assert_type, x, "No message.")

	#define HZ_INTERNAL_ASSERT_GET_MACRO_NAME(arg1, arg2, macro, ...) macro
	#define HZ_INTERNAL_ASSERT_GET_MACRO(...) HZ_EXPAND_MACRO( HZ_INTERNAL_ASSERT_GET_MACRO_NAME(__VA_ARGS__, HZ_INTERNAL_ASSERT_WITH_MSG, HZ_INTERNAL_ASSERT_NO_MSG) )

	#define HZ_ASSERT(...) HZ_EXPAND_MACRO( HZ_INTERNAL_ASSERT_GET_MACRO(__VA_ARGS__)(HZ, __VA_ARGS__) )
	#define HZ_CORE_ASSERT(...) HZ_EXPAND_MACRO( HZ_INTERNAL_ASSERT_GET_MACRO(__VA_ARGS__)(HZ_CORE, __VA_ARGS__) )
#else
	#define HZ_ASSERT(...)
	#define HZ_CORE_ASSERT(...)
#endif
```

How it works:
- So `HZ_INTERNAL_ASSERT_GET_MACRO` will extract the correct macro depending if a message was provided or not. With the help of `HZ_INTERNAL_ASSERT_GET_MACRO_NAME`.
- Assume a message was provided, `HZ_INTERNAL_ASSERT_WITH_MSG` takes in the `assert_type` such that it resolves `##assert_type##_ERROR` to `HZ_ERROR` or `HZ_CORE_ERROR` depending what assert macro has been called.
- When no message was provided, `HZ_INTERNAL_ASSERT_NO_MSG` will provide a default message. I did not add in any file name or line numbers yet, so that could be added in the future, or by you? Depending what we wanna go for... I could've opted to copy the if statement and assert twice, like it was before, but it's unified now, which I think looks good as well?
- Quickly tested all cases at runtime, and also looked at what it expanded to with VS:
  - `HZ_CORE_ASSERT` with message expands correctly
image
  - `HZ_CORE_ASSERT` without message expands correctly
image
  - `HZ_ASSERT` with message expands correctly
image
  - `HZ_ASSERT` without message expands correctly
image

Any thoughts and/or improvements?

TigerCipher:
> Does the trailing comma also gets removed on other compilers, for example when we want to compile for linux or maxOS, which doesn't use the MSVC?

It definitely works on GCC/g++ as they were (I believe) the first to introduce that feature. I don't have any experience coding on a mac so I'm not entirely sure what compilers are even available there, but if I'm not mistaken mac is indeed a unix OS, so I'd assume you can still use gcc? And besides, I'm of the personal opinion games should never be made available on mac, I am very anti-apple product lol

> Also the default message should be replaced by the custom message instead of just concatenating?

If I understand what you mean, then at least to my knowledge that would require having overloaded functions, i.e `logMessage(file, line)` for the default and `logMessage(msg)` for the non default. Though as I said, not entirely sure what you meant as I thought I had already covered that in my last comment

> Any thoughts and/or improvements?

I'd have to play around with that some to really say. I've never made any macros that complex so I'm not 100% confident I understand what's going on with yours, perhaps my solution indeed doesn't work on some macos compiler and yours does, in which case by all means that should be the way to go, though as far as readability goes, I believe my solution wins in that category - though of course the end user wouldn't need to know how the macro works, just that it does.

Also regarding the default message, as I understand from your macros, the no message would result in "Assertion Failed: No message" which I personally feel is a bit off-putting for lack of a better word, I think it should simply be "Assertion Failed" which is another reason I prefer my macros, because then you could have that or for custom message rather than only being able to supply a message, I think it would be best to support custom formatting, like if you wanted to include an error code or something, so you could do `HZ_ASSERT(false, "(Error {}): {}", -243, "Failed with message")` so (with file/line excluded) you'd end up getting `...GetClientLogger()->error("Assertion Failed (Error {}): {}", -243, "Failed with message")`

LovelySanta:
> It definitely works on GCC/g++ as they were (I believe) the first to introduce that feature. I don't have any experience coding on a mac so I'm not entirely sure what compilers are even available there, but if I'm not mistaken mac is indeed a unix OS, so I'd assume you can still use gcc? And besides, I'm of the personal opinion games should never be made available on mac, I am very anti-apple product lol

I'm just trying to code the course that cherno has envisioned, so that means including mac support...

> > Also the default message should be replaced by the custom message instead of just concatenating?
> 
> If I understand what you mean, then at least to my knowledge that would require having overloaded functions, i.e `logMessage(file, line)` for the default and `logMessage(msg)` for the non default. Though as I said, not entirely sure what you meant as I thought I had already covered that in my last comment

I think it should be `"Assertion failed: {0}"` when there is a message present, and `"Assertion failed at __FILE__:__LINE__"`. With my comment I meant that `"Assertion failed: __FILE__:__LINE__"` looked a bit weird as default message, the one I just suggested looks more natural of a message.

I did notice that `__FILE__` shows the full path, where that might be too long of a "short" message, so maybe just the filename + line could work. I found [this macro](https://stackoverflow.com/questions/8487986/file-macro-shows-full-path) to extract the filename from the `__FILE__`. I would still have to test that out tho.

> I'd have to play around with that some to really say. I've never made any macros that complex so I'm not 100% confident I understand what's going on with yours, perhaps my solution indeed doesn't work on some macos compiler and yours does, in which case by all means that should be the way to go, though as far as readability goes, I believe my solution wins in that category - though of course the end user wouldn't need to know how the macro works, just that it does.

In a normal case I would say yes, they don't need to know how it works, but I since it's a teaching series I think understanding it also is somewhat a requirement.

> Also regarding the default message, as I understand from your macros, the no message would result in "Assertion Failed: No message" which I personally feel is a bit off-putting for lack of a better word, I think it should simply be "Assertion Failed" which is another reason I prefer my macros, because then you could have that or for custom message rather than only being able to supply a message, I think it would be best to support custom formatting, like if you wanted to include an error code or something, so you could do `HZ_ASSERT(false, "(Error {}): {}", -243, "Failed with message")` so (with file/line excluded) you'd end up getting `...GetClientLogger()->error("Assertion Failed (Error {}): {}", -243, "Failed with message")`

As I stated before, I didn't add in the default message thingy yet, it was just proof of concept. I'll add it in 'quickly' as I described above.

LovelySanta:
So this one has a custom message:
```cpp
#define HZ_EXPAND_MACRO(x) x

#ifdef HZ_ENABLE_ASSERTS
	#define HZ_INTERNAL_ASSERT_IMPL(assert_type, assert_msg, x, ...) { if(!(x)) { ##assert_type##_ERROR(assert_msg, __VA_ARGS__); HZ_DEBUGBREAK(); } }
	#define HZ_INTERNAL_ASSERT_WITH_MSG(assert_type, x, ...) HZ_INTERNAL_ASSERT_IMPL(assert_type, "Assertion failed: {0}", x, __VA_ARGS__)
	#define HZ_INTERNAL_ASSERT_NO_MSG(assert_type, x) HZ_INTERNAL_ASSERT_IMPL(assert_type, "Assertion failed at {0}:{1}", x, __FILE__, __LINE__)

	#define HZ_INTERNAL_ASSERT_GET_MACRO_NAME(arg1, arg2, macro, ...) macro
	#define HZ_INTERNAL_ASSERT_GET_MACRO(...) HZ_EXPAND_MACRO( HZ_INTERNAL_ASSERT_GET_MACRO_NAME(__VA_ARGS__, HZ_INTERNAL_ASSERT_WITH_MSG, HZ_INTERNAL_ASSERT_NO_MSG) )

	#define HZ_ASSERT(...) HZ_EXPAND_MACRO( HZ_INTERNAL_ASSERT_GET_MACRO(__VA_ARGS__)(HZ, __VA_ARGS__) )
	#define HZ_CORE_ASSERT(...) HZ_EXPAND_MACRO( HZ_INTERNAL_ASSERT_GET_MACRO(__VA_ARGS__)(HZ_CORE, __VA_ARGS__) )
#else
	#define HZ_ASSERT(...)
	#define HZ_CORE_ASSERT(...)
#endif
```

> I've never made any macros that complex so I'm not 100% confident I understand what's going on with yours

I'll try and explain it a bit more.

1. `HZ_INTERNAL_ASSERT_GET_MACRO` will get the correct `HZ_INTERNAL_ASSERT_XXX_MSG` macro name depending how many `__VA_ARGS__` are passed in.
    - If there is only 1 argument, the `HZ_INTERNAL_ASSERT_GET_MACRO_NAME` will see the argument list `(arg1, macro_name_with_msg, macro_name_no_msg)`, such that the 3rd argument is `macro_name_no_msg`, and it will return the 3rd argument, the macro name `HZ_INTERNAL_ASSERT_NO_MSG` as expected.
    - If there is 2 arguments provided, that means there is a message, then the macro will see the argument list `(arg1, arg2, macro_name_with_msg, macro_name_no_msg)` and it will return the 3rd argument, the macro name `HZ_INTERNAL_ASSERT_WITH_MSG` as expected.
2. Now that we have the macro name, we provide the argument list to that new macro: `(HZ, __VA_ARGS__)` or `(HZ_CORE, __VA_ARGS__)` depending which assert that is called
    - The argument `HZ` or `HZ_CORE` will be used to create the correct assert type log message (`HZ_ERROR` or `HZ_CORE_ERROR`). This means that the `HZ_INTERNAL_ASSERT` became agnostic to which assert is called, such that the amount of macro definitions is minimized to 1 set instead of 2 sets between client and core.
    - We do call the correct `HZ_INTERNAL_ASSERT` macro, so we do know inside those macros if there is a message or not, so this is also agnostic to if there is a message argument or not, which is exactly what this PR is about.
3. I've added an extra macro `HZ_INTERNAL_ASSERT_IMPL` which is the actual implementation of the macro, which now has a custom `assert_msg` argument. This way we can provide a custom assert message for both cases.
    - In case of an assert message, I left it untouched, so it stays `"Assertion failed: {0}"` with {0} being filled with the assert message provided.
    - In case of no message, I changed it to `"Assertion failed at {0}:{1}"` such that it looks more natural of a default message as you (and I) already discussed in previous posts. The arguments are the `__FILE__` and `__LINE__`. The end result of one of those messages for now is shown below. image This can still be refined to the filename only as I pointed out earlier. I still have to try that.

TigerCipher:
> I do like this idea. I'll have to see if we can do that with my macro as well... It should be.. That would require compile time concatenation of the actual "message" with the message provided by the assert macro... I do not see directly a way to customize the custom message to allow formatting inside there... Any ideas? The only way I see that makes it possible is if the actual message is a string literal, but should we assume that this is always the case?

Apologies for the delay, I had actually begun typing a reply before your latest post, essentially saying the same things as you and that the best solution would be to combine our ideas. Initially however, with your preference of having a separate macro for the "no msg" assert, it seemed at first that the idea of an infinite number of arguments for custom formatting was impossible, so I ended up playing around and doing a bunch of trial and error these last few hours. So before I explain what I've come up with, I'd like to point out something in your macro which happened to actually be an error for me. `##assert_type##` will tell my compiler that `HZ` and `HZ_CORE` are not valid processor tokens (Using visual studio 2019 toolchain via CLion/cmake) so I had to actually remove the first set of hashtags to make it `assert_type##` so let me know if that actually ends up causing issues on your end.

Anyways, here is what I've come up with, and mind you, it is but a proof of concept and in no way perfect. And because I use clion which only works with cmake, I tried this out in my own project replacing the error macros with printf and getting rid of the debugbreak macro, but there should be no issues replacing my version with the original content

```
#define HZ_ERROR(fmt_, ...) fprintf(stderr, fmt_"\n", __VA_ARGS__)
#define HZ_CORE_ERROR(fmt_, ...) fprintf(stderr, fmt_" (core) \n", __VA_ARGS__)
#define HZ_EXPAND_MACRO(x) x


#define ARGS_RANGE(\
      _1, _2, _3, _4, _5, _6, _7, _8, _9,_10, \
     _11,_12,_13,_14,_15,_16,_17,_18,_19,_20, \
     _21,_22,_23,_24,_25,_26,_27,_28,_29,_30, \
     _31,_32,_33,_34,_35,_36,_37,_38,_39,_40, \
     _41,_42,_43,_44,_45,_46,_47,_48,_49,_50, \
     _51,_52,_53,_54,_55,_56,_57,_58,_59,_60, \
     _61,_62,_63,N,...) N
#define REVERSE_ARGS_RANGE() \
     63,62,61,60,                   \
     59,58,57,56,55,54,53,52,51,50, \
     49,48,47,46,45,44,43,42,41,40, \
     39,38,37,36,35,34,33,32,31,30, \
     29,28,27,26,25,24,23,22,21,20, \
     19,18,17,16,15,14,13,12,11,10, \
     9,8,7,6,5,4,3,2,1,0
#define NUM_ARGS_I(...) ARGS_RANGE(__VA_ARGS__)
#define NUM_ARGS_(...) NUM_ARGS_I(__VA_ARGS__,REVERSE_ARGS_RANGE())


#define HZ_INTERNAL_ASSERT_IMPL(assert_type, x, assert_msg, ...) if(!(x)) { assert_type##_ERROR(assert_msg, __VA_ARGS__); }
#define HZ_INTERNAL_ASSERT_WITH_MSG(assert_type, x, ...) HZ_INTERNAL_ASSERT_IMPL(assert_type, x, "Assertion failed: " __VA_ARGS__)
#define HZ_INTERNAL_ASSERT_NO_MSG(assert_type, x) HZ_INTERNAL_ASSERT_IMPL(assert_type, x, "Assertion failed at %s:%i", __FILE__, __LINE__)



#define VFUNC2(func, x, ...) if(NUM_ARGS_(__VA_ARGS__) > 1){ HZ_INTERNAL_ASSERT_WITH_MSG(func, x, __VA_ARGS__)} \
else { HZ_INTERNAL_ASSERT_NO_MSG(func, x) }

#define HZ_ASSERT(...) VFUNC2(HZ, __VA_ARGS__)
#define HZ_CORE_ASSERT(...) VFUNC2(HZ_CORE, __VA_ARGS__)
```

Of course, the limitation currently is that it only supports 63 arguments, but you can expand if desired, though I see no reason why anyone would ever need that many so I think its safe to keep. 

The way the argument counter works is that, lets say you call `HZ_ASSERT(false, "msg", 3, 4)` which calls vfunc2, which calls the num_args_ and expands to this: `NUM_ARGS_(false, "msg", 3, 4, REVERSE_ARGS_RANGE())` To put it some what visually:

```
// HZ_ASSERT(false, "msg", 3, 4) expands the NUM_ARGS_ to be:
/*
* false, "msg", 3, 4, // This would represent _1, _2, _3, _4 in ARGS_RANGE
* // _5 to N gets replaced by the REVERSE_ARGS_RANGE
* 63, 62, 61, .... all the way until ARGS_RANGE hits the max of 63
* // because it's in reverse, N becomes the last number the reverse args was able to insert
* in this case being 4
*/
```

Let me know if I need to explain anything else in more detail. Did a few quick tests and it worked flawlessly:

```
void test()
{
	HZ_ASSERT(false)
	HZ_ASSERT(false, "msg")
	HZ_ASSERT(false, "msg %i", 1);
	HZ_CORE_ASSERT(false);
	HZ_ASSERT(false);
	HZ_CORE_ASSERT(false, "msg");
	HZ_CORE_ASSERT(false, "msg %i", 1);
	HZ_ASSERT(false, "msg %i %f", 1, 0.3f);
	HZ_CORE_ASSERT(false, "msg %i %f", 1, 0.3f);
}
```

Additionally if you're not happy with the if else inserted by the new assert macros, you can overload macros by changing it up to the following:

```
#define _VFUNC_(name, n) name##n
#define _VFUNC(name, n) _VFUNC_(name, n)
#define VFUNC2(func, ...) _VFUNC(func, NUM_ARGS_(__VA_ARGS__)) (__VA_ARGS__)


#define HZ1(...) HZ_INTERNAL_ASSERT_NO_MSG(HZ, __VA_ARGS__)
#define HZ2(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ, __VA_ARGS__)
#define HZ3(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ, __VA_ARGS__)
#define HZ4(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ, __VA_ARGS__)
#define HZ5(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ, __VA_ARGS__)
#define HZ6(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ, __VA_ARGS__)

#define HZ_CORE1(...) HZ_INTERNAL_ASSERT_NO_MSG(HZ_CORE, __VA_ARGS__)
#define HZ_CORE2(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ_CORE, __VA_ARGS__)
#define HZ_CORE3(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ_CORE, __VA_ARGS__)
#define HZ_CORE4(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ_CORE, __VA_ARGS__)
#define HZ_CORE5(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ_CORE, __VA_ARGS__)
#define HZ_CORE6(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ_CORE, __VA_ARGS__)
```

This is so far the best I've come up with. I personally think the first version with the if else is cleaner and easier to work with.

Let me know if you have any questions, or improvements you can make upon this

LovelySanta:
> So before I explain what I've come up with, I'd like to point out something in your macro which happened to actually be an error for me. `##assert_type##` will tell my compiler that `HZ` and `HZ_CORE` are not valid processor tokens (Using visual studio 2019 toolchain via CLion/cmake) so I had to actually remove the first set of hashtags to make it `assert_type##` so let me know if that actually ends up causing issues on your end.

Hmmm... The `##assert_type##` did not error on my end, might be because it starts with `##`? I think we could change it to passing in `_` and `_CORE_` such that it would be `HZ##assert_type##ERROR` and then it shouldn't error on your side either? I'll check bot cases tomorrow (what you suggested and what I just wrote).

> Anyways, here is what I've come up with, and mind you, it is but a proof of concept and in no way perfect. And because I use clion which only works with cmake, I tried this out in my own project replacing the error macros with printf and getting rid of the debugbreak macro, but there should be no issues replacing my version with the original content

The arg counting was also my idea, but I'm not a particular fan of that, hence why I didn't suggest it to begin with. One way to prevent that would be to limit the argument list to be maximum 2 long that we pass into the `HZ_INTERNAL_ASSERT_GET_MACRO`, which would be equivalent to the condition + formatting message. That's the only thing I would be a bit stuck on, and it would work for any number of parameters as far as I can tell...

> ```cpp
> #define VFUNC2(func, x, ...) if(NUM_ARGS_(__VA_ARGS__) > 1){ HZ_INTERNAL_ASSERT_WITH_MSG(func, x, __VA_ARGS__)} \
> else { HZ_INTERNAL_ASSERT_NO_MSG(func, x) }

Since the if statement is a macro, it can be an if constexpr, such that it wouldn't be compiled anyway, even less 'code'. This is more a minor detail. Also don't forget to encapsulate the function in a scope `{}` for safety reasons

> Of course, the limitation currently is that it only supports 63 arguments, but you can expand if desired, though I see no reason why anyone would ever need that many so I think its safe to keep.

As I said before, I'm not a 100% for this solution as it doesn't look nice either, and I'm sure cherno has the same thoughts about that, but if there is no other way, I might consider it.. I do wanna try and see if my method above would yield some results, I had compiling issues when I tried it (writing this late at night, so i'm not productive atm)...

> Additionally if you're not happy with the if else inserted by the new assert macros, you can overload macros by changing it up to the following:
> 
> ```
> #define _VFUNC_(name, n) name##n
> #define _VFUNC(name, n) _VFUNC_(name, n)
> #define VFUNC2(func, ...) _VFUNC(func, NUM_ARGS_(__VA_ARGS__)) (__VA_ARGS__)
> 
> 
> #define HZ1(...) HZ_INTERNAL_ASSERT_NO_MSG(HZ, __VA_ARGS__)
> #define HZ2(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ, __VA_ARGS__)
> #define HZ3(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ, __VA_ARGS__)
> #define HZ4(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ, __VA_ARGS__)
> #define HZ5(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ, __VA_ARGS__)
> #define HZ6(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ, __VA_ARGS__)
> 
> #define HZ_CORE1(...) HZ_INTERNAL_ASSERT_NO_MSG(HZ_CORE, __VA_ARGS__)
> #define HZ_CORE2(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ_CORE, __VA_ARGS__)
> #define HZ_CORE3(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ_CORE, __VA_ARGS__)
> #define HZ_CORE4(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ_CORE, __VA_ARGS__)
> #define HZ_CORE5(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ_CORE, __VA_ARGS__)
> #define HZ_CORE6(...) HZ_INTERNAL_ASSERT_WITH_MSG(HZ_CORE, __VA_ARGS__)
> ```

At that point you could just write dedicated macros for calling it without any message and generalizing the one with a message.

> EDIT: Oh and regarding the first "message" to not have to be a string literal, haven't quite worked that one out yet. For now you'd have to do `HZ_ASSERT(false, "{}", someVar)`

Indeed a minor issue tho.. I think it's fair to assume that it is a string literal. I wonder if it would just be good enough if it's a compiler error when it's not a string literal?

TigerCipher:
> The arg counting was also my idea, but I'm not a particular fan of that, hence why I didn't suggest it to begin with. One way to prevent that would be to limit the argument list to be maximum 2 long that we pass into the HZ_INTERNAL_ASSERT_GET_MACRO, which would be equivalent to the condition + formatting message. That's the only thing I would be a bit stuck on, and it would work for any number of parameters as far as I can tell...

Do you mean something different than the macros you had already suggested? Because maximum two arguments sounds like your original solution, which wouldn't accept any number of parameters.

The arg counting I agree is a bit messy, and if thats a no go then at least as far as I can tell, would have to stick with your solution and if someone wants to format the message they can just pass in a pre-formatted string (sprintf, fmt::format, etc etc). Either solution solves the initial "todo" note/issue of allowing the assert to have the message be optional, anything extra would just be, well, extra. I've just been coming at this from my own perspective, if I was using the engine or if I was making a similar assert macro for one of my own projects. Could also have 2 separate macros, i.e HZ_ASSERT and HZ_ASSERT_F where the _F version is with mine with the unlimited parameters formatting, though I think having both would be a bit redundant

> Since the if statement is a macro, it can be an if constexpr, such that it wouldn't be compiled anyway, even less 'code'. This is more a minor detail.

My bad. All my previous classes, high school and university both, have never been with C or C++ sadly, so I'm entirely self taught lol. Actually only just first came across "constexpr" the other day and haven't really had the time to see what it was

> At that point you could just write dedicated macros for calling it without any message and generalizing the one with a message.

Then in code you'd do HZ_ASSERT1, HZ_ASSERT2, etc etc or whatever you name them, so I personally still think a generalized HZ_ASSERT macro would be better

> Indeed a minor issue tho.. I think it's fair to assume that it is a string literal. I wonder if it would just be good enough if it's a compiler error when it's not a string literal?

Nothing wrong with that I think. I can't think of a single IDE that wouldn't mark it as an error while you're coding so you know somethings not right. There is also `#define STR(x) #x` but that would really only work for literal integers, floats, etc, not for variables themselves

image

LovelySanta:
> > The arg counting was also my idea, but I'm not a particular fan of that, hence why I didn't suggest it to begin with. One way to prevent that would be to limit the argument list to be maximum 2 long that we pass into the HZ_INTERNAL_ASSERT_GET_MACRO, which would be equivalent to the condition + formatting message. That's the only thing I would be a bit stuck on, and it would work for any number of parameters as far as I can tell...
> 
> Do you mean something different than the macros you had already suggested? Because maximum two arguments sounds like your original solution, which wouldn't accept any number of parameters.
> 
> The arg counting I agree is a bit messy, and if thats a no go then at least as far as I can tell, would have to stick with your solution and if someone wants to format the message they can just pass in a pre-formatted string (sprintf, fmt::format, etc etc). Either solution solves the initial "todo" note/issue of allowing the assert to have the message be optional, anything extra would just be, well, extra. I've just been coming at this from my own perspective, if I was using the engine or if I was making a similar assert macro for one of my own projects. Could also have 2 separate macros, i.e HZ_ASSERT and HZ_ASSERT_F where the _F version is with mine with the unlimited parameters formatting, though I think having both would be a bit redundant

Yes I was indeed refering to something I was trying. The reason why we have the arg counting is that we want to check if we have a message or not (aka 1 argument versus more than 1 argument). My way was to pass in the arguments and append the macro names after it and extract the correct macro depending if 1 or 2 arguments are provided.

If we want to enforce the ability to format a string, we will have even more than 2 arguments, so that `HZ_INTERNAL_ASSERT_GET_MACRO` would break as we would pass in more than 2 arguments. So I'm thinking of a way to make sure we pass in maximum 2 arguments to `HZ_INTERNAL_ASSERT_GET_MACRO`. I was editing the original post while you were replying, I'm planning to look at [this](https://stackoverflow.com/questions/32434217/limit-number-of-expanded-arguments-with-va-args) to see if we could use it maybe in this case?

> > Indeed a minor issue tho.. I think it's fair to assume that it is a string literal. I wonder if it would just be good enough if it's a compiler error when it's not a string literal?
> 
> Nothing wrong with that I think. I can't think of a single IDE that wouldn't mark it as an error while you're coding so you know somethings not right. There is also `#define STR(x) #x` but that would really only work for literal integers, floats, etc, not for variables themselves
I'm actually having issues with that concatenation in Hazel. Currently it looks like
```cpp
#define HZ_INTERNAL_ASSERT_WITH_MSG(type, check, ...) HZ_INTERNAL_ASSERT_IMPL(type, check, "Assertion failed: {0}", __VA_ARGS__)
```
when I change it to your example (below), It doesn't seem to be willingly to concatentate the assert (format) message with the string literal provided in the macro.
```cpp
#define HZ_INTERNAL_ASSERT_WITH_MSG(type, check, ...) HZ_INTERNAL_ASSERT_IMPL(type, check, "Assertion failed: " __VA_ARGS__)
```
This is how it looks expanded atm, and I'm not understanding why it's not concatenating them...
image

TigerCipher:
> This is how it looks expanded atm, and I'm not understanding why it's not concatenating them... Also there seems to be a comma too much at the end (after "Test").

So with the comma, that will be there regardless if you pass in args or no args, as long as there are no spaces then the compiler will see it as an empty argument and compile out the extra comma (hence my original solution of using `##` as it would ensure there were no spaces, because the compiler sometimes sees spaces as non empty arguments). As for why its not concatenating  for you, as I mentioned, my example was only proof of concept and I was using printf, i.e char strings. I've never used spdlog before, so based on your error I assume it uses std string instead? If so, then we could do
```cpp
#define HZ_INTERNAL_ASSERT_WITH_MSG(type, check, ...) HZ_INTERNAL_ASSERT_IMPL(type, check, "Assertion failed: " + 
 __VA_ARGS__)
```
instead? That would actually work for non string literals as well?

LovelySanta:
I searched the discord this morning, I found [this post](https://stackoverflow.com/questions/52891546/what-does-va-args-mean), which clearly states that `##__VA_ARGS__` is gcc only, hence why I didn't want to use it (as I previously stated without knowing why exactly).

TigerCipher:
> I searched the discord this morning, I found this post, which clearly states that ##__VA_ARGS__ is gcc only, hence why I didn't want to use it (as I previously stated without knowing why exactly).

Ah. Well if you recall earlier I posted a link on the Microsoft docs talking about the experimental preprocessor command option. By default visual studio automatically removes trailing commas, but with that option on it requires you to use `##__VA_ARGS__` to make it act like other compilers, such as gcc. (Or maybe when its off I don't recall off the top of my head). For my version of visusl studio (2019) I believe it defaults to that, as I haven't changed anything but I have to use ## to make it work. But regardless, thats part of the reason I used clion instead to test the stuff yesterday so that ## wouldn't be required. Though we might need to add checks to add the ## when using a compiler like gcc?

LovelySanta:
Played around a bit more with the macros. I even had issues with argument expansion in MSVC. After the help in discord, The macros got convoluted and eventualy I ended up with around 30-ish macros just to have the ability to make my current implementation work with formatting arguments (without actual formatting yet).

Reductor finally said, at that point, if you want to have the ability to expand your message, the easiest way would be to not have a default assert message to begin with. And for that, the easiest way would be to make another macro just for the custom formatting. And at that point, you could just log the custom message before the assert itself.

Asserts are not even error handling, it's just making sure some variable has some expected value. The comment next to the assert, or the assert itself should be more than enough explenation at that point. So this is what we're ended up with, and apperently it's close to how cherno did implement it in Hazel-dev...

```cpp
#define HZ_EXPAND_MACRO(x) x
#define HZ_STRINGIFY_MACRO(x) #x

#ifdef HZ_ENABLE_ASSERTS

	namespace Hazel::Assert {

		constexpr const char* CurrentFileName(const char* path) {
			const char* file = path;
			while (*path)
			{
				if (*path == '/' || *path == '\\')
					file = ++path;
				else
					path++;
			}
			return file;
		}

	}

    #define HZ_INTERNAL_ASSERT_IMPL(type, check, msg, ...) { if(!(check)) { HZ##type##ERROR(msg, __VA_ARGS__); HZ_DEBUGBREAK(); } }
	#define HZ_INTERNAL_ASSERT_WITH_MSG(type, check, ...) HZ_INTERNAL_ASSERT_IMPL(type, check, "Assertion '{0}' failed: {1}", HZ_STRINGIFY_MACRO(check), __VA_ARGS__)
	#define HZ_INTERNAL_ASSERT_NO_MSG(type, check) HZ_INTERNAL_ASSERT_IMPL(type, check, "Assertion '{0}' failed at {1}:{2}", HZ_STRINGIFY_MACRO(check), ::Hazel::Assert::CurrentFileName(__FILE__), __LINE__)

	#define HZ_INTERNAL_ASSERT_GET_MACRO_NAME(arg1, arg2, macro, ...) macro
	#define HZ_INTERNAL_ASSERT_GET_MACRO(...) HZ_EXPAND_MACRO( HZ_INTERNAL_ASSERT_GET_MACRO_NAME(__VA_ARGS__, HZ_INTERNAL_ASSERT_WITH_MSG, HZ_INTERNAL_ASSERT_NO_MSG) )

	#define HZ_ASSERT(...) HZ_EXPAND_MACRO( HZ_INTERNAL_ASSERT_GET_MACRO(__VA_ARGS__)(_, __VA_ARGS__) )
	#define HZ_CORE_ASSERT(...) HZ_EXPAND_MACRO( HZ_INTERNAL_ASSERT_GET_MACRO(__VA_ARGS__)(_CORE_, __VA_ARGS__) )
#else
	#define HZ_ASSERT(...)
	#define HZ_CORE_ASSERT(...)
#endif
```

So I would for now already suggest to put this in the PR, define the two macros for expanding and stringifying the macros in Base.h, followed by an include statement `#include "Assert.h` which contains the whole ifdef.

TigerCipher:
> Reductor finally said, at that point, if you want to have the ability to expand your message, the easiest way would be to not have a default assert message to begin with. And for that, the easiest way would be to make another macro just for the custom formatting. And at that point, you could just log the custom message before the assert itself.

Yeah that’s what I was talking about before when I mentioned `HZ_ASSERT_F` or just no default message (i.e have both say “Assertion Failed at file:line” and then someone could add a message if they wanted and it could be concatenated so they could use formatting if they wanted. Having the same default message for both is what I had believed to be the best option, I just continued working on the macros because you felt different default messages were the way to go.

> So I would for now already suggest to put this in the PR

Are you asking me to edit the pr/commit to include these changes, or if cherno had already implemented something, should it just be ignored until he merges that branch into here?

LovelySanta:
> > Reductor finally said, at that point, if you want to have the ability to expand your message, the easiest way would be to not have a default assert message to begin with. And for that, the easiest way would be to make another macro just for the custom formatting. And at that point, you could just log the custom message before the assert itself.
> 
> Yeah that’s what I was talking about before when I mentioned `HZ_ASSERT_F` or just no default message (i.e have both say “Assertion Failed at file:line” and then someone could add a message if they wanted and it could be concatenated so they could use formatting if they wanted. Having the same default message for both is what I had believed to be the best option, I just continued working on the macros because you felt different default messages were the way to go.

I do still believe that the 2 different default messages are the way... *some* way at least... I do want to implement your idea where we just concatenate `"Assertion failed: "` with the first argument, such that it does allow for custom formatting. This is not difficult, we can easily write a function `constexpr const char* concatenateStrings(const char* str1, const char* str2)`. It won't be a macro, but it would still be runtime, so it achieves the same result after compile time.

The issue I am/was trying to solve, when we do not provide any message, we also want to keep the default message. The issue here is needing a decent way of extracting if there are any parameters or not. The reason this is an issue, as my current implementation `HZ_INTERNAL_ASSERT_GET_MACRO(...)` assumes only 1 or 2 argument is passed in. When passing in a 3rd argument (for formatting) it will break.

> > So I would for now already suggest to put this in the PR
> 
> Are you asking me to edit the pr/commit to include these changes, or if cherno had already implemented something, should it just be ignored until he merges that branch into here?

Yes I am asking to edit the PR to include the changes I mentioned above. We can still edit it afterwards. The PR will be squashed into a single commit, so the amount of commits in this PR do not matter anyway....

TigerCipher:
> Yes I am asking to edit the PR to include the changes I mentioned above. It is a compiler agnositc solution to the initial intend of this PR, and better than what this PR suggests at this point. We can still edit it afterwards. The PR will be squashed into a single commit, so the amount of commits in this PR do not matter anyway....

Apologies, as I first mentioned, I've never been involved in a PR before, I've always been a lone wolf type of dev. I believe I've properly edited the PR, let me know if I misunderstood something about the changes you suggested I make

LovelySanta:
Don't be sorry, you learn by doing, since we're all here to learn there is no need to appologize.
I forgot to mention that we might also want to add that Assert file to the pch, similar that we did with the log file.

ilikenoodles2:
After playing around with macros, I arrived at this:
```
#define COMMA_IF_NO_PARENTHESIS(...)
#define COMMA_IF_NO_PARENTHESIS ,
#define EXPAND(...) __VA_ARGS__
#define HZ_GET_FMT(a, b, c, d) a " " b " " c " " d

template<typename... Args>
constexpr bool HZ_ASSERT_HAS_ARGS(Args... args) { return sizeof...(args); }

#define HZ_CORE_ASSERT(check, fmt, ...) {																														\
	if(!(check))																																				\
	{																																							\
		if constexpr(HZ_ASSERT_HAS_ARGS(__VA_ARGS__))																											\
			HZ_CORE_ERROR(HZ_GET_FMT("Assertion '", HZ_STRINGIFY_MACRO(check), "' failed:", fmt) COMMA_IF_NO_PARENTHESIS EXPAND(__VA_ARGS__));					\
		else																																					\
			HZ_CORE_ERROR(HZ_GET_FMT("Assertion '", HZ_STRINGIFY_MACRO(check), "' failed at {0}:{1}"), ::Hazel::Assert::CurrentFileName(__FILE__), __LINE__);	\
		HZ_DEBUGBREAK();																																		\
	}																																							\
}
```
This takes any amount of arguments, and allows for formatting.

LovelySanta:
Are you sure that it works? if I just pass in a check, then there will be no ftm, nor extra args, so that would already error on the macro args, not even inside the macro? msvc has a permissive setting, you have to disable that

ilikenoodles2:
Tested no arguments, and with multiple arguments, and it works.  Also the permissive setting says no for me, but your welcome to try it though.

I'll be honest, I don't exactly know how those string literals work, but it seems that with no fmt or args, it decomposes into
```" " " "```
which the compiler still sees as a string.

**Comment by LovelySanta** [comment]
Ok, that can work then indeed. A small feedback note, should we define `#define HZ_GET_FMT(a, b, c, d) a " " b " " c " " d` with just 2 arguments? It would be more usefull for the full Hazel API then as well. In case we need more, we can for example call `HZ_GET_FMT(HZ_GET_FMT(a, b), c)` or `HZ_GET_FMT(HZ_GET_FMT(a, b), HZ_GET_FMT(c, d))`.

Also I noticed the `HZ_CORE_ASSERT` is missing `\` after each line, which is a minor detail

**Comment by ilikenoodles2** [comment]
@LovelySanta, in that case you could call it ```CONCAT_CSTR``, or ```CAT_CSTR```.

Good catch, the fix for that would be ```if constexpr(HZ_ASSERT_HAS_ARGS(__VA_ARGS__) || HZ_ASSERT_HAS_ARGS(fmt))```

Furthermore, it's also possible to address the case if fmt is not a cstring with ```static_assert(HZ_IS_FMT_VALID(fmt), "Format must be a cstring");``` and 
```
template<typename... T>
constexpr bool HZ_IS_FMT_VALID(T... fmt)
{
	return std::is_same_v<T..., const char*>;
}
```

Or you could even allow fmt to be a std::string, and branch again, to handle each case.  There are many possibilities.

File: Hazel/src/Hazel/Core/Assert.h [hunk 1]
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
```
LovelySanta, line 2: - This PR should resolve #326 by including `Log.h`. Also mention this in the initial post then. This PR should also close #327.
- For similar reasons we should include `Base.h` for the `HZ_STRINGIFY_MACRO`

```suggestion

#include "Hazel/Core/Base.h"
#include "Hazel/Core/Log.h"

```

Hazel/src/Hazel/Core/Assert.h
```diff
@@ -0,0 +1,36 @@
...
+						path++;
+				}
+				return file;
+			}
+	}
+
+	// Alteratively we could use the same "default" message for both "WITH_MSG" and "NO_MSG" and
+	// provide support for custom formatting by concatenating the formatting string instead of having the format inside the default message
+	#define HZ_INTERNAL_ASSERT_IMPL(type, check, msg, ...) { if(!(check)) { HZ##type##ERROR(msg, __VA_ARGS__); HZ_DEBUGBREAK(); } }
+	#define HZ_INTERNAL_ASSERT_WITH_MSG(type, check, ...) HZ_INTERNAL_ASSERT_IMPL(type, check, "Assertion '{0}' failed: {1}", HZ_STRINGIFY_MACRO(check), __VA_ARGS__)
```
LovelySanta, line 24: Having a closer look, the message should explain the assert, so in this case the check should not be part of the message. It would become too long otherwise...
```suggestion
	#define HZ_INTERNAL_ASSERT_WITH_MSG(type, check, ...) HZ_INTERNAL_ASSERT_IMPL(type, check, "Assertion failed: {0}", __VA_ARGS__)
```

LovelySanta (commented):
Some more details after using it for a while

Hazel/src/Hazel/Core/Assert.h
LovelySanta: Resolved in https://github.com/TheCherno/Hazel/pull/306/commits/05c552bcb1d68cd6ffbdd6f63973fcfd8a6e408c

PR #309, opened 2020-08-17
Add uniform location cache for performance.

CleverSource:
Currently uniform locations are not cached, due to uniforms being constantly looked up, this can save lots of performance overhead. Cherno has talked about wanting to cache uniforms for a while now and this is a simple implementation of this.

My proposed fix for this is creating a function named ``GetUniformLocation`` which will return a ``GLint``, due to Glad not being included in the header file, I am currently just doing a typedef for ``GLint``, I have also categorized this under the TODO comment stating to remove the typedef's. 

The implementation is straight forward but basically all that is done is an unordered map is created, the key for it is the name, and the value is the GLint location; then the ``GetUniformLocation(const std::string& name)`` function will check if the name is in the cache, if so it returns that from the cache, if it's not in the cache it will retrieve the uniform location and add it to the cache.

I have marked the function as const because it will only retrieve the uniform location and it's not modifying the shader, due to it being const I make the unordered map "mutable" so that the unordered map can be modified by a const function, my reasoning for this being marked as const is because the actual operation of the function is to read the location of the uniform - basically the underlying implementation caches the value but obviously that isn't important to the caller since it's not modifying the shader at all.

### DIFF
--- a/Hazel/src/Platform/OpenGL/OpenGLShader.cpp
+++ b/Hazel/src/Platform/OpenGL/OpenGLShader.cpp
@@ -204,8 +204,9 @@ namespace Hazel {
 	{
 		HZ_PROFILE_FUNCTION();
 
-		if (m_UniformLocationCache.find(name) != m_UniformLocationCache.end())
-			return m_UniformLocationCache[name];
+		auto cachedLocation = m_UniformLocationCache.find(name);
+		if (cachedLocation != m_UniformLocationCache.end())
+			return cachedLocation->second;
 
 		GLint location = glGetUniformLocation(m_RendererID, name.c_str());
 		m_UniformLocationCache[name] = location;
--- a/Hazel/src/Platform/OpenGL/OpenGLShader.cpp
+++ b/Hazel/src/Platform/OpenGL/OpenGLShader.cpp
@@ -200,6 +200,18 @@ namespace Hazel {
 		glUseProgram(0);
 	}
 
+	GLint OpenGLShader::GetUniformLocation(const std::string& name) const
+	{
+		HZ_PROFILE_FUNCTION();
+
+		if (m_UniformLocationCache.find(name) != m_UniformLocationCache.end())
+			return m_UniformLocationCache[name];
+
+		GLint location = glGetUniformLocation(m_RendererID, name.c_str());
+		m_UniformLocationCache[name] = location;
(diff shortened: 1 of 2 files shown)


CleverSource:
> Is it possible for the shader to be recompiled? (in which case, the cache would need to be cleared)

Currently shaders are only compiled once in the constructor of OpenGLShader. The function used for compilation of a shader is also private, so I don't think this would happen, unless the engine was going to recompile it (which currently never happens).

Hazel/src/Platform/OpenGL/OpenGLShader.cpp
```diff
@@ -200,6 +200,18 @@ namespace Hazel {
 		glUseProgram(0);
 	}
 
+	GLint OpenGLShader::GetUniformLocation(const std::string& name) const
+	{
+		HZ_PROFILE_FUNCTION();
+
+		if (m_UniformLocationCache.find(name) != m_UniformLocationCache.end())
+			return m_UniformLocationCache[name];
```
freeman40, line 208: you're looking up the cache twice.  would it be slightly more efficient to store the result of find, and then if it isn't end() return ->second?

CleverSource: Yes, this would work, I'll update the code now.

mdaub:
It would be better to cache the Uniforms when compiling the Shader using glGetProgramiv with GL_ACTIVE_UNIFORMS which will return the number of uniforms. Then loop through and get name and type information for each uniform number with glGetActiveUniform. After that you can ask OpenGL for its location and cache it in the unordered map.

I just have the program assert in my GetUniformLocation function in debug mode if it can't find the uniform in the cache.

Here is my version if you want to take a look. My names are a little different.
[InitUniformCache.txt](https://github.com/TheCherno/Hazel/files/5096781/InitUniformCache.txt)

freeman40:
It may also be better to deal with the situation where the requested uniform does not exist in the shader.  At the moment this situation will be silently ignored, so if the developer makes a spelling mistake or whatever, they are left scratching their heads as to why nothing is being rendered (e.g. spelt "u_ViewProjection" wrong)

TheCherno:
Obsolete (switching to Vulkan and deprecating OpenGL).

PR #321, opened 2020-08-23
Removed redundant GetProjection()

advikkabra:
In `Scene.cpp`, while beginning the scene, the camera'a projection is sent, even though the function requires a `Camera` instance. This creates an unnecessary construction and copy.

Giving the `*mainCamera` instead of `mainCamera->GetProjection()` in the `Renderer2D::BeginScene()` function.

### DIFF
--- a/Hazel/src/Hazel/Scene/Scene.cpp
+++ b/Hazel/src/Hazel/Scene/Scene.cpp
@@ -100,7 +100,7 @@ namespace Hazel {
 
 		if (mainCamera)
 		{
-			Renderer2D::BeginScene(mainCamera->GetProjection(), *cameraTransform);
+			Renderer2D::BeginScene(*mainCamera, *cameraTransform);
 
 			auto group = m_Registry.group<TransformComponent>(entt::get<SpriteRendererComponent>);
 			for (auto entity : group)


LovelySanta:
The main camera is a SceneCamera, derived from Camera, however the renderer2D function requires an OrtographicCamera, also derived from Camera:

https://github.com/TheCherno/Hazel/blob/c4c66118b26d9a89547fd9230089f6869fc78238/Hazel/src/Hazel/Renderer/Renderer.h#L18

So I'm not sure if this is correct? By passing in the projection, it will call the constructor of the OrtographicCamera, which shouldn't be the same as converting SceneCamera into a Camera into a OrtographicCamera? Seems equally weird to me?

advikkabra:
The Scene.cpp file calls not the Renderer::BeginScene() function, but the Renderer2D::BeginScene() function, which takes in a Camera& and a transform matrix. Inside the function, the camera is only used to get the projection and multiply with the transform matrix. Hence, the construction is unnecessary as we can just get the transform from the mainCamera.

LovelySanta:
> Renderer2D::BeginScene() function, which takes in a Camera&

Yep, my mistake, you're right! We can just pass in the mainCamera.

PR #322, opened 2020-08-23
passing initializer_list by value, it is cheap to copy it, added const qualifiers, VertexArray now using default virtual destructor

andreimk:
- passing initializer_list by value, it is cheap to copy it, added const qualifiers
- added const qualifiers
- VertexArray now using default virtual destructor

### DIFF
--- a/Hazel/src/Hazel/Renderer/Buffer.h
+++ b/Hazel/src/Hazel/Renderer/Buffer.h
@@ -70,7 +70,7 @@ namespace Hazel {
 	public:
 		BufferLayout() {}
 
-		BufferLayout(const std::initializer_list<BufferElement>& elements)
+		BufferLayout(std::initializer_list<BufferElement> elements)
 			: m_Elements(elements)
 		{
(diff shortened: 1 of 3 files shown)


salimp2009:
you can also add std::move that will be even cheaper instead of 2 copies, it will be  one copy one move 
if the elements are not owned by anything else 

that is a more preferred modern C++ construct but it al depends if the elements are not owned by anything else.

BufferLayout(std::initializer_list<BufferElement> elements)
 			: m_Elements(std::move(elements))

ReDucTor:
@salimp2009 initalizer list is just a set of pointers it doesnt matter if its moved or not, infact the pass by value approach in this PR is the standard way to do it. I wouldnt be certain about what you say being faster, I would suspect its equal to or slower then the approach of pass by value in the PR and likely immeasurable.

If moving is important initializer list should not be used its just const pointers so does a bunch of copies.

salimp2009:
@ReDucTor ...ohhh...ok you are right if they are just pointers then copy is better :) thanks

PR #257, opened 2020-06-05
Add linux compatible GenProjects file

caelwithcats:
#### Description
I've added a Linux GenProjects file called `Lnx-GenProjects.sh` in the `scripts` folder and the premake executable ` premake5-linux` in `vendor/bin/premake/`.

PR #323, opened 2020-08-24
Add linux support v4 (up to date)

### LINKED ISSUE #308: Building solution
This repo his hard to implement on visual studio. To the developer do consider using build tools like Cmake or Scons for automatic analyzes source code file dependencies to aid in making solutions.

caelwithcats:
On linux some segmentation faults come up on the current state of Hazel.

Issues this solves       | Resolves #308
Other PRs this solves    | Resolves #257, Closes #237 (equivalent), Closes #302 (other approach)

This PR fixes a segmentation fault and add unofficial linux support. In the current state it should launch on Linux using OpenGL, if the error of "no Linux support" has been removed. No functionallity or code has been altered in order to make this compatibility.

#### Additional context
- [x] The viewport is blank in Hazelnut. Needs to be fixed
- [x] Stop segmentation fault when closing the Hazel window

Edited by @LovelySanta: Added a decent PR submission


Hazel/src/Hazel/Scene/Scene.cpp
```diff
@@ -124,11 +124,11 @@ namespace Hazel {
 		auto view = m_Registry.view<CameraComponent>();
 		for (auto entity : view)
 		{
-			auto& cameraComponent = view.get<CameraComponent>(entity);
+			auto cameraComponent = view.get<CameraComponent>(entity);
```
9Y0, line 127: Why did you remove the reference? If you hover over the `get` function, you'll see that it returns a CameraComponent&
```suggestion
			auto& cameraComponent = view.get<CameraComponent>(entity);
```

Hazelnut/premake5.lua
```diff
@@ -27,6 +27,25 @@ project "Hazelnut"
...
+			"ImGui",
+			"Xrandr",
+			"Xi",
+			"GLU",
+			"GL",
+			"X11",
+			"dl",
+			"pthread",
+			"stdc++fs"
+		}
```
9Y0, line 44: Do you really need to link against all of them? In my experience the following are enough but it may differ per distro. I'm running Linux Mint 20, which is Ubuntu based.
```suggestion
		links
		{
			"GLFW",
			"Glad",
			"ImGui",
			"X11",
			"dl",
			"pthread",
		}
```

Hazel/src/Hazel/Scene/Scene.cpp
OverShifted: you should leave that `&` there.
removing `&` is when getting multiple components
```c++
auto& cameraComponent = view.get<CameraComponent>(entity); // Correct
auto cameraComponent = view.get<CameraComponent>(entity); // Error

auto& [cam, trans] = (view / group).get<CameraComponent, TransFormComponent>(entity); // Error (for some reason it works for MSVC but I get compile errors when compiling on Linux with gcc / g++
auto [cam, trans]= (view / group).get<CameraComponent, TransFormComponent>(entity); // Correct
```
```suggestion
			auto& cameraComponent = view.get<CameraComponent>(entity);
```

Hazelnut/premake5.lua
OverShifted: Exacly I use this code
```lua
LinkLibs = {
	"GLFW",
	"Glad",
	"ImGui",
	"Lua",
	"Box2D"
}
```
```lua
links (LinkLibs)

filter "system:linux"
		pic "on"
		systemversion "latest"
		staticruntime "on"
		links { "dl", "pthread" }
```
I think even X11 is not needed but I'm not sure.
Linux Mint 19.3

Hazel/src/Hazel/Scene/Scene.cpp
caelwithcats: With GCC it gives you an error when using a reference

9Y0: That's pretty weird because for it works perfectly fine with gcc. Could you maybe share you error?

9Y0:
> Bundle the assets folder with the executable

Why does this need to be done? In my opinion you should just run the executable (i.e. Sandbox or Hazelnut) from their respective directories. That is also what VS does.

caelwithcats:
@9Y0 You have to have the assets folder inside the same folder as the executable. Having one executable would make it easier for  players to play your game.

9Y0:
> You have to have the assets folder inside the same folder as the executable

@caelwithcats Not entirely true. You need to start the application while being in the folder were the assets are stored (i.e. if you run it via the terminal). If I remember correctly, @LovelySanta was also working on a feature like this and I think Cherno said in a stream that he didn't like that and said people had to just start the application within the right directory. Can you maybe confirm this Santa?

LovelySanta:
Cherno said something more in the lines of this:

1. Hazel is not released yet, so don't expect it to run outside of VS, which also sets the working directory correctly, as defined through premake.
2. Understandably for those that do not have access to VS (for example they follow on linux), they should learn about how working directory works and make sure it's set correctly, for example as @9Y0 mentioned, by running it from the correct folder, which would be the correct working directory at that point.
3. When the engine would be 'released' then it would be expected that games could be made with it, and run without the need of VS or any working directory to be set correctly. That would be the part of an asset system to resolve this, which is not even implemented in Hazel-dev at this point in time (but it is [on the roadmap](https://hazelengine.com/dev/issues) of Hazel-dev).

caelwithcats:
can this be merged?

Hazel/src/Hazel/Renderer/Renderer2D.cpp
```diff
@@ -42,30 +42,34 @@ namespace Hazel {
...
 	void Renderer2D::Init()
 	{
 		HZ_PROFILE_FUNCTION();
 
-		s_Data.QuadVertexArray = VertexArray::Create();
+		HZ_CORE_ASSERT(!s_Data, "Renderer2D is already initialized!");
 
-		s_Data.QuadVertexBuffer = VertexBuffer::Create(s_Data.MaxVertices * sizeof(QuadVertex));
-		s_Data.QuadVertexBuffer->SetLayout({
+		s_Data = new Renderer2DData();
```
LovelySanta, line 53: This suggestion looks weird... I meant to suggest to delete the empty line on line 52, such that the assert & the new Render2DData allocation are grouped as a block...
```suggestion
		HZ_CORE_ASSERT(!s_Data, "Renderer2D is already initialized!");
		s_Data = new Renderer2DData();
```

LovelySanta (changes requested):
I currently don't have access to a linux PC since it has some issues. So I do recommend some reviews from like DaveAxiom or others with a linux machine. Other than leaving the actual #error for linux support in the platform detection I only have some (minor) formatting changes suggested.

I do feel like I could merge this in, I do however require you to edit your initial message, so it's conform with a PR layout.
- Mention the segmentation fault
- Mention compatibility with linux (only require users to delete the #error, maybe even link to that specific line of code even
- Mention that it still uses 100% of the windows code, that only syntax changes (and premake) in order to make it compatible to follow along on linux machines

Sandbox/premake5.lua
```diff
@@ -27,6 +27,25 @@ project "Sandbox"
...
+			"Xrandr",
+			"Xi",
+			"GLU",
+			"GL",
+			"X11",
+			"dl",
+			"pthread",
+			"stdc++fs"
+		}
+		defines
```
LovelySanta, line 45:
```suggestion
		}
		
		defines
```

Hazel/src/Hazel/Core/Window.cpp
```diff
@@ -1,20 +1,24 @@
 #include "hzpch.h"
 #include "Hazel/Core/Window.h"
 
-#ifdef HZ_PLATFORM_WINDOWS
+#if defined(HZ_PLATFORM_WINDOWS)
+	#include "Platform/Windows/WindowsWindow.h"
+#elif defined(HZ_PLATFORM_LINUX)
```
DaveAxiom, line 6: Why is there a check for precompiler directives when `WindowsWindow.h` uses GLFW which is cross platform? The filename should be renamed and does create confusion but checking for defines makes this out to be protecting the source code from an issue that doesn't exist.

Hazel/src/Hazel/Core/Window.cpp
```diff
@@ -1,20 +1,24 @@
...
 	#include "Platform/Windows/WindowsWindow.h"
 #endif
 
 namespace Hazel
 {
 	Scope<Window> Window::Create(const WindowProps& props)
 	{
-	#ifdef HZ_PLATFORM_WINDOWS
+	#if defined(HZ_PLATFORM_WINDOWS)
```
DaveAxiom, line 14: Preprocessor check is unnecessary as above and only causes confusion.

Hazel/premake5.lua
```diff
@@ -43,12 +43,32 @@ project "Hazel"
...
+
 	filter "system:windows"
 		systemversion "latest"
 
+		links 
+		{ 
+			"opengl32.lib"
+		}
+		
 		defines
```
DaveAxiom, line 72: Premake should handle the platform specific defines. `HZ_PLATFORM_WINDOWS` should be here otherwise there isn't much point in having Premake.

Hazel/src/Hazel/Core/PlatformDetection.h
```diff
@@ -33,7 +33,6 @@
 	#error "Android is not supported!"
 #elif defined(__linux__)
 	#define HZ_PLATFORM_LINUX
```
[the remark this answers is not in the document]
DaveAxiom: This is a Linux branch which should be in working order. The `#error` preprocessor directive raised by `HZ_PLATFORM_LINUX` shouldn't be in a branch that is effectively a prototype or recipe for Cherno to use in future videos. This branch should be a working state that is ready to clone.

Hazel/premake5.lua
```diff
@@ -43,12 +43,32 @@ project "Hazel"
...
+		links 
+		{ 
+			"Xrandr",
+			"Xi",
+			"GLU",
+			"GL",
+			"X11"
+		}
+
+		defines
```
DaveAxiom, line 60: `HZ_PLATFORM_LINUX` should be defined for the Hazel library since this is a Premake oriented project.

Sandbox/premake5.lua
```diff
@@ -27,6 +27,26 @@ project "Sandbox"
...
+			"Glad",
+			"ImGui",
+			"Xrandr",
+			"Xi",
+			"GLU",
+			"GL",
+			"X11",
+			"dl",
+			"pthread",
+			"stdc++fs"
```
DaveAxiom, line 43: `stdc++fs` was included in my Hazel Linux branch because `std::filesystem` was used. And the inclusion of this library is only for the transitional C++17 GCC (Clang?) compilers before the filesystem object was moved into the `stdc++` standard library -- which is linked by default.

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -52,6 +52,11 @@ namespace Hazel {
 			glfwSetErrorCallback(GLFWErrorCallback);
 		}
 
+
+		glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 4);
+		glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 5);
+		glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);
```
DaveAxiom, line 58: This file does handle GLFW and adding OpenGL requirements makes sense but maybe it'd be more appropriate in an OpenGL project file?  We know that the specification of Hazel is OpenGL version 4.5 yet Cherno didn't add OpenGL functions to enforce this requirement.

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -52,6 +52,11 @@ namespace Hazel {
...
+
 		{
 			HZ_PROFILE_SCOPE("glfwCreateWindow");
 		#if defined(HZ_DEBUG)
```
DaveAxiom, line 62: I want to point out that the `HZ_DEBUG` define is introduced by Premake depending on the build specification. Platform defines should be organized together.

Hazel/premake5.lua
DaveAxiom: Come to think of it. Premake handles every project submodule and it specifies the platform type for these libraries. A top down approach would make the entire project much neater.

Hazel/src/Hazel/Core/PlatformDetection.h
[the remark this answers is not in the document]
LovelySanta: I do want to merge this PR, since it has nothing 'new' for Linux compatibility other than some minor tweaks to the code without actually changing any behaviour... So yes, it 'should' be in working order, but when I merge this, Hazel will still not officially support it. Those that follow along on linux can easily delete this one line and it should work for them, rather than requiring to make many changes to the code base...

Hazel/src/Hazel/Core/Window.cpp
LovelySanta: I agree, but renaming it to `GLFWWindow` and `GLFWInput` feels out of the scope of this PR, so since this is specific to Windows (for now), I feel like we should at least treat it as such...

Hazel/src/Hazel/Core/Window.cpp
LovelySanta: Same responds as above

Hazel/premake5.lua
LovelySanta: Cherno did say that the current way of finding the platform (using predefined macros) in [`Hazel/Core/PlatformDetection.h`](https://github.com/TheCherno/Hazel/blob/857d8d60e63b23c67edf1bb10e2715fb57089597/Hazel/src/Hazel/Core/PlatformDetection.h) was indeed somewhat an issue. This also feels out of the scope of this PR.. In #320 this was changed to a seperate file and added to the pch...

Hazel/premake5.lua
LovelySanta: Same comment as above

Hazel/src/Platform/Windows/WindowsWindow.cpp
LovelySanta: I don't think this is bad, however I think it should add a check around it to check the rendering api to make sure openGL is being used?

I also think this needs to move down a but, such it is part of the `HZ_PROFILE_SCOPE("glfwCreateWindow");` ?

Sandbox/premake5.lua
caelwithcats: It doesn't matter if stdc++fs is included anyway?

Hazel/src/Hazel/Core/PlatformDetection.h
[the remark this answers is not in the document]
caelwithcats: @LovelySanta it works, what's wrong with it?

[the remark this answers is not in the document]
LovelySanta: That it officially isn't supported, so if someone downloads it from github, it's supposed to say it's not compatible, hence why the error must stay. However, you could add a comment that when you remove this error line, it should run unofficially

DaveAxiom (approved):

Sandbox/premake5.lua
DaveAxiom: I'd say it's fine to leave it since Hazel is a teaching project. `stdc++fs` is a provisional library that is needed for linking on outdated compilers. I have a comment on the Linux branch I've been maintaining to explain that.

DaveAxiom:
I'm really confused by the last two patches: "Define platform macros" and "Undefine platform macros". The first patch adds `#define` for the platforms which @LovelySanta indicated would be better than letting a header file figure it out. And the second patch undoes that patch.

LovelySanta:
I messaged caelwithcats on discord. I meant that the platform macros are currently resolved in a header file (which is part of the pch as of now). So declaring them also in premake is redundant. When it becomes an issue in the future, we can always go back to defining them in premake, but then we should also remove the header file. As I said before, this seems outside the scope of the PR, this PR should only add/change code to make the current state compatible.

Hazel/src/Hazel/Core/PlatformDetection.h
```diff
@@ -33,6 +33,9 @@
...
+	// Linux is not offically supported by Hazel.
+	// You can remove this line to enable Linux support
+	#error "Linux is not supported! Read message in Hazel/src/Hazel/Core/PlatformDetection.h"
```
DaveAxiom, line 38: I think this is misguided.

LovelySanta: I'm with DaveAxiom on this one...

caelwithcats: I'm confused on what you mean

DaveAxiom: Your branch should work without strings attached. That's my take.

LovelySanta (commented):
I had a closer look at this, In the other Linux PR, the premake uses a tar file, where this one doesn't, shouldn't we just keep the file as it was on the premake site?

caelwithcats:
I'm closing this pull request since the previous linux support PR has been updated

PR #327, opened 2020-08-26
Moves all Log.h inclusion to Base.h

3x3y3z3t:
The issue have been made: #326.

Issues this solves       | solves #326

The fix consists of two step:
- Adds `#define LOG_H` in `Log.h` to act as header guard.
- Includes `Log.h` in `Base.h` *after* the definition of `Hazel::Ref`.

#### Additional context
As @LovelySanta said, `Log.h` should be included in `Base.h`. However `Log.h` also includes `Base.h`, resulted in a circular inclusion. Since `Log.h` only needs the custom reference system `Hazel::Ref` defined in `Base.h`, I moved the inclusion of `Log.h` to *after* that definition so by the time `Log.h` is included, `Hazel::Ref` is already defined and ready to be used.

### DIFF
--- a/Hazel/src/Hazel/Core/Application.cpp
+++ b/Hazel/src/Hazel/Core/Application.cpp
@@ -1,8 +1,6 @@
 #include "hzpch.h"
 #include "Hazel/Core/Application.h"
 
-#include "Hazel/Core/Log.h"
-
 #include "Hazel/Renderer/Renderer.h"
 
 #include "Hazel/Core/Input.h"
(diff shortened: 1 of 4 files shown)


LovelySanta:
There is no need for that guard, ´#pragma once´ should take care of that.

3x3y3z3t:
> There is no need for that guard, `#pragma once` should take care of that inside `Log.h`.

The reason I put header guard in is for some reason the compilation failed all the time with some "symbol undefined" error. I thought that was because of the circular inclusion (most of the time circular inclusion is the cause of "symbol undefined"). Note that at that time, both `#include "Log.h"` and `#include "Base.h" is placed at the beginning of the other file.
It seems that the problem back then is because of where `#include "Log.h" is placed.

PR #328, opened 2020-08-28
Add EventCategoryMouseButton to MouseButtonEvent

feresr:
`EventCategoryMouseButton` is not being used by any Event.
I think `MouseButtonEvent` should be an `EventCategoryMouseButton`. We can alternatively get rid of it


LovelySanta:
Good catch!

PR #329, opened 2020-09-02
Removed decltype(auto) annotation in HZ_BIND_EVENT_FN lambda

Halbjobbit:
I removed the "-> decltype(auto)" to reduce code complexity. The return type can be deduced since C++14, which should not make any problems since Hazel is using C++17.

### DIFF
--- a/Hazel/src/Hazel/Core/Base.h
+++ b/Hazel/src/Hazel/Core/Base.h
@@ -29,7 +29,7 @@
 
 #define BIT(x) (1 << x)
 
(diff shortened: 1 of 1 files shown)


ReDucTor:
This is functionally different, think of `decltype(auto)` as a.perfect forwarding of return values. So if the wrapped function returned a reference  that would be maintained unlike this change.

ReDucTor:
Also the whole return type deduction thing you mention has existed for lambda since its additon in c++11

Halbjobbit:
Oh ok. It looked like it would work, but I looked it up and understand how I was wrong. Thanks for noticing @ReDucTor! I will close the PR

PR #330, opened 2020-09-05
Use std::regex for shader processing

feresr:
Just watched the episode on [Shader asset Files](https://www.youtube.com/watch?v=8wFEzIYRZXg&list=PLlrATfBNZ98dC-V-N3m0Go4deliWHPFwT&index=44) and found the shader parsing code a bit hard to follow.

I'm using `std::regex` in this PR.  I'm relatively new to CPP I'm not sure of the performance impact this might introduce (the original code is already doing a lot of relatively expensive string search operations anyways, but still something to keep in mind)
I think this PR makes the code a little bit more readable, wdty?


TheCherno:
Obsolete (SPIR-V).

PR #290, opened 2020-07-28
Updated Readme 

BEASTSM96:
#### Description
Updated readme. Moved the ``hazel-dev``  below the ``main features to come: `` part. And added Cherno's other links (aka discord, youtube, etc..).

PR #334, opened 2020-09-24
Fixed Properties panel

BEASTSM96:
A small change to the Properties panel.

#### Additional context
And also this comes with the commits from #290 but no changes to `README.md`.  *And the same with the `imgui.ini`*.

EDIT; I will include the changes from #290.


BEASTSM96:
^^^ Is the main fix. all the others is from #290.

PR #335, opened 2020-09-25
Made Editor work with high dpi monitors.

standtall007:
**Describe the issue (if no issue has been made)**
Issue: On monitors with high DPI such as 4K monitors the editor window and imgui ui is very small compared to monitors that use a dpi of 1.

**PR impact**
No PR to my knowledge.

Issues this solves       | small window / ui on monitors with a dpi greater than 1

**Proposed fix** 
The fix is included in this pull request... when the editor program is launched, it will scan the primary monitor for it's dpi and if it is above 1 then it changes the window and ui's resolution (e.g. on a 4K monitor it will have a dpi of 1.5 so it scales the window and imgui ui by 1.5)

**Additional context**
I have tested this and use it myself and have been doing so for a little while, there is no bugs with this request that i know of.


Hazel/src/Hazel/Core/Window.h
```diff
@@ -42,6 +42,8 @@ namespace Hazel {
 		virtual void* GetNativeWindow() const = 0;
 
 		static Scope<Window> Create(const WindowProps& props = WindowProps());
+	public:
+		static float highDPIScaleFactor;
```
CleverSource, line 46: To keep things in line with naming conventions; this should be ``s_HighDPIScaleFactor``.

Hazel/src/Platform/Windows/WindowsWindow.cpp
```diff
@@ -13,6 +13,8 @@
 
 namespace Hazel {
 	
+	float Window::highDPIScaleFactor = 1.0f;
```
CleverSource, line 16: Basically add a getter and setter function for this? Also would this be better suited to go in ``WindowProps``?

CleverSource (changes requested):

Hazel/src/Hazel/Core/Window.h
standtall007: Yeah that is true. i will change that.

Hazel/src/Platform/Windows/WindowsWindow.cpp
standtall007: What do you mean by WindowProps? Do you mean create a new script? or do you mean a class? I am honestly confused by this...

standtall007: I figured out what you meant, i am sleepy :P I won't do that unless Cherno suggests to do that because it doesn't really add anything to it, also the WindowProps struct is iirc only used in application's window creation, where as the Window class is more used.

standtall007: Ah, i see now what you meant by your questions. I don't like unneeded getters and/or setters, so i won't make them unless Cherno really wants them instead of a public static variable. Also there really is no difference in having the s_HighDPIScaleFactor to be inside WindowProps. But i still want to thank you for your suggestions. 😃

Dreaner:
Nice work! I also had the problem on my 4K screen.

standtall007:
> 
> 
> Nice work! I also had the problem on my 4K screen.

Thanks 😄 i appreciate your feedback.

standtall007:
Please merge #351 instead. Because this is a complete mess right now!

PR #339, opened 2020-09-30
Update LICENSE

caelwithcats:
#### Pull Request
This PR adds to the year and copyright to the Licence


LovelySanta:
That is not how it is done, that is just the appendix, and as the license state, instead of filling it in there, it should be added to the top of each source file, as I proposed in #109. I'll close this as changing it in the appendix has no meaning at all...

PR #340, opened 2020-10-01
Implemented Feature: OpenGL Error Macro HZ_GLCALL()

babaliaris:
This approach is based on the [Dealing with Errors in OpenGL](https://youtu.be/FBbPWSOQ0-w?list=PLlrATfBNZ98foTJPJ_Ev03o2oq3-GGOS2) video by **TheCherno**.

### New Feature / Maintenance & Debugging.
Getting **detailed information** about an OpenGL error can save you from hours of debugging! It's really common and easy to make a silly mistake in an OpenGL call that can waste your time.

**HZ_GLCALL()** will help a lot in spotting these kinds of mistakes. Below you can see an example output:

As you can see, **HZ_GLCALL()** will print the following information:
1. The **file** and the **line** where the macro is located.
2. The **Error Code Number** both in **decimal** and **hexadecimal** notation.
3. The **GLenum**.
4. A **description** of the error.

### Unscoped Macro
**HZ_GLCALL()** is unscoped, meaning that it does not add a scope allowing you to do things like:
```c++
HZ_GLCALL(GLuint program = glCreateProgram());
```
For this reason, make sure to include a body in if, while, switch, etc... statements:
```c++
if (something)
{
    HZ_GLCALL(GLuint program = glCreateProgram());
}
```

## How It Was Implemented
Two new files were created inside **Hazel/Platform/OpenGL**:
1. [OpenGLError.h](https://github.com/babaliaris/Hazel/blob/OpenGLErrorMacro/Hazel/src/Platform/OpenGL/OpenGLError.h)
2. [OpenGLError.cpp](https://github.com/babaliaris/Hazel/blob/OpenGLErrorMacro/Hazel/src/Platform/OpenGL/OpenGLError.cpp)

containing the code that implements the functionality of **HZ_GLCALL()**.

Then I just wrapped all the **gl** calls using the **HZ_GLCALL()** macro in all the files.

## Important note:

In the file [OpenGLError.cpp](https://github.com/babaliaris/Hazel/blob/OpenGLErrorMacro/Hazel/src/Platform/OpenGL/OpenGLError.cpp) at lines **141** and **159**, I had to check for the OpenGL Context using: `wglGetCurrentContext()` . The reason is that if another thread is running, like the profiler while the application is closed or the window has been destroyed, then `while ( glGetError() );` can be stuck in an infinity loop. This if statement makes sure that this will not happen.

Thank you for your time :)

### DIFF
--- a/Hazel/src/Platform/OpenGL/OpenGLBuffer.cpp
+++ b/Hazel/src/Platform/OpenGL/OpenGLBuffer.cpp
@@ -1,5 +1,6 @@
 #include "hzpch.h"
 #include "Platform/OpenGL/OpenGLBuffer.h"
+#include "OpenGLError.h"
 
 #include <glad/glad.h>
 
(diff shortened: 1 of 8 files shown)


LovelySanta:
I proposed your solution in #75 already. Cherno himself commented (https://github.com/TheCherno/Hazel/issues/75#issuecomment-510943404) that since OpenGL 4.3 there is an OpenGL callback function instead of having to poll after each call.

This solution is already implemented in the `OpenGLRendererAPI` with #114. The callback is registered inside `OpenGLRendererAPI::Init()`
https://github.com/TheCherno/Hazel/blob/c4c66118b26d9a89547fd9230089f6869fc78238/Hazel/src/Platform/OpenGL/OpenGLRendererAPI.cpp#L35
which calls the function that logs them to spdlog
https://github.com/TheCherno/Hazel/blob/c4c66118b26d9a89547fd9230089f6869fc78238/Hazel/src/Platform/OpenGL/OpenGLRendererAPI.cpp#L8-L26

Thanks for the PR, however I'll close this as a duplicate of #75 / #114 as this functionality is already present inside the codebase.

babaliaris:
Oh, I didn't know about it. Thanks for the review!

PR #341, opened 2020-10-02
Removed warnings from External Dependencies

sparkskapil:
### There are lot of compiler warnings from External Headers (spdlog and fmt)
Before Changes
---
image

After Changes
---
image

#### This Pull request will not have any impact on the functionality of Hazel Engine. 
It will just suppress all warnings arising from the external headers. 
1. spdlog/spdlog.h
2. spdlog/fmt/ostr.h

#### This solution to suppress warning can be found on stackoverflow.
https://stackoverflow.com/a/2541990/9271231

### DIFF
--- a/Hazel/src/Hazel/Core/Log.h
+++ b/Hazel/src/Hazel/Core/Log.h
@@ -2,8 +2,12 @@
 
 #include "Hazel/Core/Base.h"
 
+// This ignores all warnings raised inside External headers
+#pragma warning( push, 0)
 #include <spdlog/spdlog.h>
 #include <spdlog/fmt/ostr.h>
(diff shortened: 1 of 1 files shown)


Hazel/src/Hazel/Core/Log.h
```diff
@@ -2,8 +2,12 @@
 
 #include "Hazel/Core/Base.h"
 
+// This ignores all warnings raised inside External headers
+#pragma warning( push, 0)
```
LovelySanta, line 6:
```suggestion
#pragma warning(push, 0)
```

LovelySanta (approved):
👍

PR #344, opened 2020-10-04
Add Renderer2D::StartBatch() and make FlushAndReset() not call EndScene()

dr01d3k4:
Cleans up 2 issues:
- Renderer2D had duplicated code for starting a new batch
- FlushAndReset() called EndScene(), when it was actually just ending the batch. This will be incorrect if more logic is added to EndScene() that should only be performed once per BeginScene(), and also means that number of calls to BeginScene() and number of calls to EndScene() do not match

Fix was to move the begin batch logic to a new method called StartBatch(), and change FlushAndReset() to now directly call Flush() and StartBatch(). Also had to move the updating of the vertex buffer from EndScene() to Flush(), but as it's a prerequisite for Flush(), it should be handled there anyway.


Hazel/src/Hazel/Renderer/Renderer2D.cpp
```diff
@@ -163,12 +165,8 @@ namespace Hazel {
...
 	{
-		EndScene();
-
-		s_Data.QuadIndexCount = 0;
-		s_Data.QuadVertexBufferPtr = s_Data.QuadVertexBufferBase;
-
-		s_Data.TextureSlotIndex = 1;
+		Flush();
+		StartBatch();
 	}
```
LovelySanta, line 170: Since you added a function `StartBatch()`, wouldn't it make sense to rename this to something along the lines of `FlushAndRestartBatch()` or simply `NextBatch()`?

LovelySanta (approved):

Hazel/src/Hazel/Renderer/Renderer2D.cpp
dr01d3k4: I like the sound of NextBatch()! I renamed the method and updated the PR.

PR #346, opened 2020-10-07
Amended code to conform with C++ standard

DaveAxiom:
The C function `strcpy_s()` was used in a C++ program which is invalid in standard compliant compilers. `strcpy_s()` may even be proprietary to Microsoft. The equivalent C and C++ function `strncpy()` is used in this patch for better compatibility.

This patch was tested under Linux on a Linux branch compiled by GCC 9.3.


DaveAxiom:
This patch is not Linux specific. It's a conformity fix for all C++ compliant compilers!

LovelySanta (commented):
Just tested this in MSVC, I get a warning
```
\Hazelnut\src\Panels\SceneHierarchyPanel.cpp(74,9): warning C4996: 'strncpy': This function or variable may be unsafe. Consider using strncpy_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS
```

So shouldn't we change this to `strncpy_s` ?

DaveAxiom:
`strncpy_s` is not in the C++ standard.

caelwithcats:
@DaveAxiom nothing in windows is standard

csp256:
It's in since C++11, however:

> As with all bounds-checked functions, strncpy_s is only guaranteed to be available if __STDC_LIB_EXT1__ is defined by the implementation and if the user defines __STDC_WANT_LIB_EXT1__ to the integer constant 1 before including string.h.

I'd like to note that the memset is wasteful.

DaveAxiom:
@csp256 I'm finding that `STDC_LIB_EXT1` is a C11 optional Feature Test Macro. That makes sense because your reference says to include "string.h" which is a C header -- not a C++ header.

Where did you find that information? I couldn't find much in a search. I use [DevDocs](https://devdocs.io/) for programming language reference.

Are you on the TheCherno Discord server? I'm going to jump into the *C/C++* channel and get a consensus for a resolution.

csp256:
https://en.cppreference.com/w/c/string/byte/strncpy

That's a neat reference but I suggest you use cppreference for C++. I'm not active on discord but I can hop on.

DaveAxiom:
That's a C reference. C isn't C++. C enthusiasts point out that C has features that C++ doesn't. I'm going to get a few opinions on what's best. I'm leaning towards asking that this patch is merged because Microsoft's C++ compiler is out of order on the C++ standard -- that compiler only offers a warning.

PR #347, opened 2020-10-07
Removed and replaced code with equivalent standard library function

DaveAxiom:
A Hazel contributor created a function that has a Standard C++ Library implementation. This patch removes the custom work and swaps in `std::filesystem::path` for appropriate use.

I tested this patch outside of the context of Hazel. @LovelySanta Please test this fix in Visual Studio before accepting!

### DIFF
--- a/Hazel/src/Hazel/Core/Assert.h
+++ b/Hazel/src/Hazel/Core/Assert.h
@@ -2,30 +2,15 @@
 
 #include "Hazel/Core/Base.h"
 #include "Hazel/Core/Log.h"
+#include <filesystem>
 
 #ifdef HZ_ENABLE_ASSERTS
-
-	namespace Hazel::Assert
-	{
-		// Returns the simple file name rather than full path as suggested by LovelySanta
-		constexpr const char* CurrentFileName(const char* path) {
-				const char* file = path;
-				while (*path)
-				{
-					if (*path == '/' || *path == '\\')
-						file = ++path;
-					else
-						path++;
-				}
-				return file;
-			}
-	}
 
 	// Alteratively we could use the same "default" message for both "WITH_MSG" and "NO_MSG" and
 	// provide support for custom formatting by concatenating the formatting string instead of having the format inside the default message
 	#define HZ_INTERNAL_ASSERT_IMPL(type, check, msg, ...) { if(!(check)) { HZ##type##ERROR(msg, __VA_ARGS__); HZ_DEBUGBREAK(); } }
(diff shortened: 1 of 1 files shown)


Hazel/src/Hazel/Core/Assert.h
```diff
@@ -2,30 +2,15 @@
...
-				return file;
-			}
-	}
-
 	// Alteratively we could use the same "default" message for both "WITH_MSG" and "NO_MSG" and
 	// provide support for custom formatting by concatenating the formatting string instead of having the format inside the default message
 	#define HZ_INTERNAL_ASSERT_IMPL(type, check, msg, ...) { if(!(check)) { HZ##type##ERROR(msg, __VA_ARGS__); HZ_DEBUGBREAK(); } }
 	#define HZ_INTERNAL_ASSERT_WITH_MSG(type, check, ...) HZ_INTERNAL_ASSERT_IMPL(type, check, "Assertion failed: {0}", __VA_ARGS__)
-	#define HZ_INTERNAL_ASSERT_NO_MSG(type, check) HZ_INTERNAL_ASSERT_IMPL(type, check, "Assertion '{0}' failed at {1}:{2}", HZ_STRINGIFY_MACRO(check), ::Hazel::Assert::CurrentFileName(__FILE__), __LINE__)
+	#define HZ_INTERNAL_ASSERT_NO_MSG(type, check) HZ_INTERNAL_ASSERT_IMPL(type, check, "Assertion '{0}' failed at {1}:{2}", HZ_STRINGIFY_MACRO(check), std::filesystem::path(__FILE__).filename(), __LINE__)
```
LovelySanta, line 13: I'll suggest to explicitly ask for a string, as the [`std::operator<<(std::filesystem::path const &)`](https://en.cppreference.com/w/cpp/filesystem/path/operator_ltltgtgt) is specified as follows:
> Performs stream input or output on the path p. [`std::quoted`](http://en.cppreference.com/w/cpp/io/manip/quoted) is used so that spaces do not cause truncation when later read by stream input operator.

```suggestion
	#define HZ_INTERNAL_ASSERT_NO_MSG(type, check) HZ_INTERNAL_ASSERT_IMPL(type, check, "Assertion '{0}' failed at {1}:{2}", HZ_STRINGIFY_MACRO(check), std::filesystem::path(__FILE__).filename().string(), __LINE__)
```

which changes the result from
`[12:11:40] APP: Assertion 'false' failed at "Application.cpp":21`
to
`[12:07:07] APP: Assertion 'false' failed at Application.cpp:21`

PR #352, opened 2020-10-23
Refactor SerializeEntity

MURPHYENGINEERING:
- break out long method
- DRY

### DIFF
--- a/Hazel/src/Hazel/Scene/SceneSerializer.cpp
+++ b/Hazel/src/Hazel/Scene/SceneSerializer.cpp
@@ -82,70 +82,86 @@ namespace Hazel {
 	{
 	}
 
+	template<typename T>
+	static void SerializeComponent(YAML::Emitter& out, const T& component);
+
+  template<>
+	static void SerializeComponent(YAML::Emitter& out, const TagComponent& tagComponent)
+  {
+	  out << YAML::Key << "TagComponent";
+		out << YAML::BeginMap; // TagComponent
+
+		out << YAML::Key << "Tag" << YAML::Value << tagComponent.Tag;
+
+		out << YAML::EndMap; // TagComponent
+  }
+
+  template<>
+	static void SerializeComponent(YAML::Emitter& out, const TransformComponent& tc)
+  {
+  	out << YAML::Key << "TransformComponent";
+		out << YAML::BeginMap; // TransformComponent
+
+		out << YAML::Key << "Translation" << YAML::Value << tc.Translation;
+		out << YAML::Key << "Rotation" << YAML::Value << tc.Rotation;
+		out << YAML::Key << "Scale" << YAML::Value << tc.Scale;
+
+		out << YAML::EndMap; // TransformComponent
+  }
+
+  template<>
+	static void SerializeComponent(YAML::Emitter& out, const CameraComponent& cameraComponent)
+  {
+		out << YAML::Key << "CameraComponent";
+		out << YAML::BeginMap; // CameraComponent
+
+		auto& camera = cameraComponent.Camera;
+
+		out << YAML::Key << "Camera" << YAML::Value;
+		out << YAML::BeginMap; // Camera
+		out << YAML::Key << "ProjectionType" << YAML::Value << (int)camera.GetProjectionType();
+		out << YAML::Key << "PerspectiveFOV" << YAML::Value << camera.GetPerspectiveVerticalFOV();
+		out << YAML::Key << "PerspectiveNear" << YAML::Value << camera.GetPerspectiveNearClip();
+		out << YAML::Key << "PerspectiveFar" << YAML::Value << camera.GetPerspectiveFarClip();
+		out << YAML::Key << "OrthographicSize" << YAML::Value << camera.GetOrthographicSize();
+		out << YAML::Key << "OrthographicNear" << YAML::Value << camera.GetOrthographicNearClip();
+		out << YAML::Key << "OrthographicFar" << YAML::Value << camera.GetOrthographicFarClip();
+		out << YAML::EndMap; // Camera
+
+		out << YAML::Key << "Primary" << YAML::Value << cameraComponent.Primary;
+		out << YAML::Key << "FixedAspectRatio" << YAML::Value << cameraComponent.FixedAspectRatio;
+
+		out << YAML::EndMap; // CameraComponent
+  }
+
+  template<>
+	static void SerializeComponent(
+		YAML::Emitter& out, const SpriteRendererComponent& spriteRendererComponent)
+  {
+		out << YAML::Key << "SpriteRendererComponent";
+		out << YAML::BeginMap; // SpriteRendererComponent
+
+		out << YAML::Key << "Color" << YAML::Value << spriteRendererComponent.Color;
+
+		out << YAML::EndMap; // SpriteRendererComponent
+  }
+
+	template<typename T>
+	static void TrySerializeComponent(YAML::Emitter& out, const Entity& entity)
+  {
+	  if (entity.HasComponent<T>())
+      SerializeComponent(out, entity.GetComponent<T>());
+  }
+
 	static void SerializeEntity(YAML::Emitter& out, Entity entity)
 	{
 		out << YAML::BeginMap; // Entity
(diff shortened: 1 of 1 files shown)


Hazel/src/Hazel/Scene/SceneSerializer.cpp
```diff
@@ -82,70 +82,86 @@ namespace Hazel {
 	{
 	}
 
-	static void SerializeEntity(YAML::Emitter& out, Entity entity)
+	template<typename T>
+	static void SerializeComponent(YAML::Emitter& out, const T& component);
```
LovelySanta, line 86: I would suggest to do the same as with `Scene::OnComponentAdded`
https://github.com/TheCherno/Hazel/blob/398e41367885ef5697bb0701d0fe1bc798c783dc/Hazel/src/Hazel/Scene/Scene.cpp#L104-L108
Thus:
```suggestion
	template<typename T>
	static void SerializeComponent(YAML::Emitter& out, const T& component)
	{
		static_assert(false);
	}
```

Hazel/src/Hazel/Scene/SceneSerializer.cpp
```diff
@@ -82,70 +82,86 @@ namespace Hazel {
...
-			out << YAML::EndMap; // SpriteRendererComponent
-		}
+	template<typename T>
+	static void TrySerializeComponent(YAML::Emitter& out, const Entity& entity)
+	{
+		if (entity.HasComponent<T>())
+			SerializeComponent(out, entity.GetComponent<T>());
+	}
+
+	static void SerializeEntity(YAML::Emitter& out, Entity entity)
```
LovelySanta, line 156:
```suggestion
	static void SerializeEntity(YAML::Emitter& out, Entity& entity)
```

Hazel/src/Hazel/Scene/SceneSerializer.cpp
```diff
@@ -82,70 +82,86 @@ namespace Hazel {
...
+	}
 
-			out << YAML::EndMap; // SpriteRendererComponent
-		}
+	template<typename T>
+	static void TrySerializeComponent(YAML::Emitter& out, const Entity& entity)
+	{
+		if (entity.HasComponent<T>())
+			SerializeComponent(out, entity.GetComponent<T>());
+	}
```
LovelySanta, line 154: First of all, entity can't be const because of the `entity.HasComponent<T>()` and `entity.GetComponent<T>()` are not const either, the HasComponent could be const, but the GetComponent would require extra overloads, which are not worth it imo.. So at least the const modifier on Entity has to go. It creates errors otherwise.

```suggestion
	template<typename T>
	static void TrySerializeComponent(YAML::Emitter& out, Entity& entity)
	{
		if (entity.HasComponent<T>())
			SerializeComponent(out, entity.GetComponent<T>());
	}
```

Next, I am not so font of the "Try do do it" function...
Can I suggest a namechange? Why not call it `SerializeComponent` as well? Function overloading can handle this perfectly...

```suggestion
	template<typename T>
	static void SerializeComponent(YAML::Emitter& out, Entity& entity)
	{
		if (entity.HasComponent<T>())
			SerializeComponent(out, entity.GetComponent<T>());
	}
```

I would then also suggest moving this function to the top, before the other template specifications using the actual components (thus before line 84)

Hazel/src/Hazel/Scene/SceneSerializer.cpp
```diff
@@ -82,70 +82,86 @@ namespace Hazel {
...
+
+	static void SerializeEntity(YAML::Emitter& out, Entity entity)
+	{
+		out << YAML::BeginMap; // Entity
+		out << YAML::Key << "Entity" << YAML::Value << "12837192831273"; // TODO: Entity ID goes here
+
+		TrySerializeComponent<TagComponent>(out, entity);
+		TrySerializeComponent<TransformComponent>(out, entity);
+		TrySerializeComponent<CameraComponent>(out, entity);
+		TrySerializeComponent<SpriteRendererComponent>(out, entity);
```
LovelySanta, line 164: More a follow-up on a previous comment, if the previous suggestion is merged, then these require renaming
```suggestion
		SerializeComponent<TagComponent>(out, entity);
		SerializeComponent<TransformComponent>(out, entity);
		SerializeComponent<CameraComponent>(out, entity);
		SerializeComponent<SpriteRendererComponent>(out, entity);
```

Hazel/src/Hazel/Scene/SceneSerializer.cpp
```diff
@@ -82,70 +82,86 @@ namespace Hazel {
...
-			out << YAML::EndMap; // Camera
-
-			out << YAML::Key << "Primary" << YAML::Value << cameraComponent.Primary;
-			out << YAML::Key << "FixedAspectRatio" << YAML::Value << cameraComponent.FixedAspectRatio;
-
-			out << YAML::EndMap; // CameraComponent
-		}
+	template<>
+	static void SerializeComponent(
+		YAML::Emitter& out, const SpriteRendererComponent& spriteRendererComponent)
```
LovelySanta, line 139: I know the line looks long, cherno uses abriviations loke `src` for the spriteRenderComponent name, at least it should remain on a single line

```suggestion
	static void SerializeComponent(YAML::Emitter& out, const SpriteRendererComponent& spriteRendererComponent)
```

If you want you can rename them all to their abbriviation, i'm not gonna make suggestions for all of them, just for this one, but if you merge it, at least do it for all of them
```suggestion
	static void SerializeComponent(YAML::Emitter& out, const SpriteRendererComponent& src)
```
And don't forget to change it inside the function as well, obviously...

LovelySanta (changes requested):
Most of my comments are suggestions, but a few of them require changes.

Hazel/src/Hazel/Scene/SceneSerializer.cpp
MURPHYENGINEERING: What is the purpose of the static assert here? If we leave the default case undefined I get this message in MSVC, which seems useful enough:

`static function 'void SerializeComponent<T>(Emitter &,const T &)' declared but not defined`

Hazel/src/Hazel/Scene/SceneSerializer.cpp
MURPHYENGINEERING: Thanks, I might have given away that I didn't compile this

Hazel/src/Hazel/Scene/SceneSerializer.cpp
MURPHYENGINEERING: I feel the same about that name, especially since it suggests exception handling. How would you feel about `SerializeComponentIfExists`? I'm also okay with your suggestion; if that's what you prefer I'll support that.

Hazel/src/Hazel/Scene/SceneSerializer.cpp
MURPHYENGINEERING: Thanks for your feedback--these are all great points. I made this commit because I left some critical comments on the video and felt obligated to put in some work instead of just complaining. I didn't know you guys would be seriously reviewing it, so in light of that I'll retract for now and come back with a higher effort contribution. Cheers

Hazel/src/Hazel/Scene/SceneSerializer.cpp
LovelySanta: At least now it's giving a useful error saying "error here" rather than "hey, you might have forgotten to implement this one"

Hazel/src/Hazel/Scene/SceneSerializer.cpp
LovelySanta: I think `SerializeComponent` would be the best option, I first opted for `SerializeComponent_s` or something, but I tried especially using the same name if it would recognize the function overloading, it looks consistent in the implementation of SerializeEntity and SerializeScene

Hazel/src/Hazel/Scene/SceneSerializer.cpp
LovelySanta: Any community addition is always appreciated, and there is not one way of doing something, cherno does go out of his way when the community asks something specific even, on the other hand he praises when we do something good, and he gives feedbacks if he thinks it could be better...

PR #355, opened 2020-10-25
Added a default extension string

CybernetHacker14:
#### Added Default Extension
On saving files, if the extension is not specified, the file isn't saved with a `.hazel` extension, and hence doesn't turn up in the OpenFile Dialog.

#### Proposed fix
Set the default extension using `OPENFILENAMEA.lpstrDefExt` option.

### DIFF
--- a/Hazel/src/Platform/Windows/WindowsPlatformUtils.cpp
+++ b/Hazel/src/Platform/Windows/WindowsPlatformUtils.cpp
@@ -9,7 +9,6 @@
 #include "Hazel/Core/Application.h"
 
 namespace Hazel {
-
 	std::string FileDialogs::OpenFile(const char* filter)
 	{
 		OPENFILENAMEA ofn;
@@ -40,6 +39,7 @@ namespace Hazel {
 		ofn.nMaxFile = sizeof(szFile);
 		ofn.lpstrFilter = filter;
 		ofn.nFilterIndex = 1;
+		ofn.lpstrDefExt = ".hazel";
 		ofn.Flags = OFN_PATHMUSTEXIST | OFN_FILEMUSTEXIST | OFN_NOCHANGEDIR;
 		if (GetSaveFileNameA(&ofn) == TRUE)
 		{
@@ -47,5 +47,4 @@ namespace Hazel {
 		}
 		return std::string();
 	}
-
 }
--- a/Hazel/src/Platform/Windows/WindowsPlatformUtils.cpp
+++ b/Hazel/src/Platform/Windows/WindowsPlatformUtils.cpp
@@ -1,6 +1,7 @@
 #include "hzpch.h"
 #include "Hazel/Utils/PlatformUtils.h"
 
+#include <sstream>
 #include <commdlg.h>
 #include <GLFW/glfw3.h>
 #define GLFW_EXPOSE_NATIVE_WIN32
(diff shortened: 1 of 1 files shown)


LovelySanta (commented):
First of all, can you undo the lines you deleted between the function and the namespace? those are generally there in all files.

Now to the point, this function is part of a utils class, so it doesn't make sense to add the default .hazel? Would be appropriate as a 2nd parameter...

However, this does not feel like the root of the problem? When we are saving, it doesn't get the .hazel extension, shouldn't we fix the default extension during saving instead/as well? Such that if we save without an extension, that it gives a default extension (also specified as optional parameter?)

CybernetHacker14:
> First of all, can you undo the lines you deleted between the function and the namespace? those are generally there in all files.
> 
> Now to the point, this function is part of a utils class, so it doesn't make sense to add the default .hazel? Would be appropriate as a 2nd parameter...
> 
> However, this does not feel like the root of the problem? When we are saving, it doesn't get the .hazel extension, shouldn't we fix the default extension during saving instead/as well? Such that if we save without an extension, that it gives a default extension (also specified as optional parameter?)

Yeah, will undo the lines. Adding the extension as a second parameter looks like an overkill, cause inconsistencies can rise up, if a different string is passed, other than the expected one. We can use a `constexpr` though.

For the third concern, I believe, what I am doing is exactly what you told. I am using the `lpstrDefExt` option , while "saving". Both functions use the same `OPENFILENAMEA` thing.

LovelySanta:
> Yeah, will undo the lines. Adding the extension as a second parameter looks like an overkill, cause inconsistencies can rise up, if a different string is passed, other than the expected one. We can use a `constexpr` though.

Isn't that the whole point? What if we want to open/save an asset created with Hazel, it should be able to support other extensions, which is also why that function is taking in filters instead of just hardcoding Hazel extensions.

I can see how another parameter can be overkill. Another idea would be to extract the default extension from the filter? That would be a good alternative instead? I just don't think that hardcoding the default extension is the right way to go, for example, the filters could not even contain .hazel as an extension, but then it would still use it as a default?

> For the third concern, I believe, what I am doing is exactly what you told. I am using the `lpstrDefExt` option , while "saving". Both functions use the same `OPENFILENAMEA` thing.

Oh yes, my bad! I did quickly look at the diff, however I didn't unfold the collapsed code, it looked like you added it to the OpenFile method instead. You implemented it right.

CybernetHacker14:
Yeah, I got your point. Will be changing the implementation.

CybernetHacker14:
This extracts the extension from the filter, although I had to use a `stringstream`. Any more optimized suggestions are welcome.

Hazel/src/Platform/Windows/WindowsPlatformUtils.cpp
```diff
@@ -40,12 +41,25 @@ namespace Hazel {
...
+		std::stringstream s(filterString);
+		std::string defaultExtension;
+
+		while (s >> defaultExtension) {
+			if (defaultExtension.find('*') < defaultExtension.length()) {
+				ofn.lpstrDefExt = defaultExtension.c_str();
+			}
+		}
+		//
+
```
ilikenoodles2, line 56: Because each individual filter is null terminated, and assuming the first filter is the default filter, you could replace everything with 
```
ofn.lpstrDefExt = strchr(filter, '\0') + 1;
```

PR #356, opened 2020-10-25
Added .bat file for removing solution and build files

CybernetHacker14:
#### Automatically delete VS solution files and build directories
In order to make a thorough rebuild, we have to go in each submodule or project folder, to find each visual studio solution file and any build directories.

#### PR impact

#### Proposed fix
I have added a batch file in the scripts folder which detects `bin` and `bin-int` directories recursively and deletes it. Also it detects the generated visual studio solution files and deletes it as well.

#### Additional Context
A workaround is needed to exclude the bin folder in the premake directory

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -1,4 +1,5 @@
 include "./vendor/premake/premake_customization/solution_items.lua"
+include "./vendor/premake/premake_customization/clean_build.lua"
 include "Dependencies.lua"
 
 workspace "Hazel"
--- a/scripts/Win-CleanBuild.bat
+++ b/scripts/Win-CleanBuild.bat
@@ -1,8 +1,7 @@
 @echo off
 pushd %~dp0\..\
 RMDIR /S /Q .vs
-DEL *.sln 
-DEL /S /Q *.vcxproj *.vcxproj.filters *.vcxproj.user
+DEL /S /Q *.sln *.vcxproj *.vcxproj.filters *.vcxproj.user
 FOR /d /r . %%d IN ("bin") DO @IF EXIST "%%d" rd /s /q "%%d"
 FOR /d /r . %%d IN ("bin-int") DO @IF EXIST "%%d" rd /s /q "%%d"
 popd
--- a/scripts/Win-CleanBuild.bat
+++ b/scripts/Win-CleanBuild.bat
@@ -1,0 +1,9 @@
+@echo off
+pushd %~dp0\..\
+RMDIR /S /Q .vs
+DEL *.sln 
+DEL /S /Q *.vcxproj *.vcxproj.filters *.vcxproj.user
+FOR /d /r . %%d IN ("bin") DO @IF EXIST "%%d" rd /s /q "%%d"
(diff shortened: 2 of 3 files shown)


LovelySanta:
I don't really see the point in this? When we (re)build the script, it will generate the 'new' vs files, and when *anything* is different than what is already there, it will write the new file, if they are identical that would be obsolete...

Now for the build files, I don't see a reason for this, the .bat file is for windows, so we are developing using VS, and  if you right click your solution, you can press clean to do this exact step for the bin and bin-int directories before pressing build/rebuild.

Maybe you find a better use case that I am missing?

CybernetHacker14:
I had seen Cherno, maybe once or twice, manually delete the `bin` and `bin-int` folders to make a clean build. Yeah the `.bat` file will be Windows only, so on other OS, the specific script file needs to be re-written.

I am not sure whether - the options mentioned by you, do a thorough clean-up, incase we want to completely rebuild the project - be it the solution files, or the actual application build itself.
Also, if someone accidentally removes or edits the `.gitignore` file, this may cause unnecessary files to be committed. I wanted to create a script, which when run, cleans the entire source of unnecessary build directories and solution files, which may or may not be deleted via the `Win-GenProjects.bat` script or Visual Studio's options.

Right now, it does have a bug, cause it deletes the bin folder in premake, so maybe this pull request would be discarded.

CybernetHacker14:
Also, I had seen this being implemented in the Udacity C++ Nanodegree project, where there were two scripts - one to build the project using CMake, and other to clean the generated CMake build directory. So I thought it would be a good idea to add this feature here as well.

caelwithcats:
you could just do this in premake
```lua
if _ACTION == "clean" then
   os.rmdir("bin")
   os.rmdir("bin-int")
end
```
not tested (out of the docs)

LovelySanta:
I like the solution posted by @caelwithcats, but we should provide an absolute path, for example by using `"%{wks.location}/bin/"` and `"%{wks.location}/bin-int/"`

CybernetHacker14:
Alright, I didn't know premake had that facility of doing that stuff. I will give it a try.

CybernetHacker14:
> you could just do this in premake
> 
> ```lua
> if _ACTION == "clean" then
>    os.rmdir("bin")
>    os.rmdir("bin-int")
> end
> ```
> 
> not tested (out of the docs)

`os.rmdir` only deletes empty folders

LovelySanta:
Are you sure? According to the [premake documentation](https://github.com/premake/premake-core/wiki/os.rmdir), it does "Removes an existing directory as well as any files or subdirectories it contains."

CybernetHacker14:
Okay, so apparently, `os.rmdir` was able to delete the non-empty directory. The problem that now exists is, dynamic path, such as `%{wks.location}` isn't getting parsed properly. Also, I had to use the `newaction()` system to get the `clean` command to work properly.

TheCherno:
I agree with @LovelySanta , this isn't necessary. Closing.

PR #357, opened 2020-10-26
Added a custom event system

Sausty:
## Added support for custom events.
Code is super wacky but hm... it works?

Changes :
- Layer class now has a OnCustomEvent function
- Application class now has a OnCustomEvent method where it iterates all the layers and calls Layer::OnCustomEvent
- Added a new EventType as well as a new EventCategory
- CustomEvent class and CustomEventDispatcher class

How to use : 
I added a TestEvent example in Editor Layer, it works great. Right now it triggers the event on key press just for testing purposes, but it can works with any kind of condition as long as you call the CustomEvent properly.

### DIFF
--- a/Hazel/src/Hazel/Core/Application.cpp
+++ b/Hazel/src/Hazel/Core/Application.cpp
@@ -72,6 +72,18 @@ namespace Hazel {
 		}
 	}
 
+	void Application::OnCustomEvent(CustomEvent& e)
+	{
+		HZ_PROFILE_FUNCTION();
+
+		for (auto it = m_LayerStack.rbegin(); it != m_LayerStack.rend(); ++it)
+		{
+			if (e.Handled)
+				break;
+			(*it)->OnCustomEvent(e);
+		}
+	}
+
 	void Application::Run()
 	{
 		HZ_PROFILE_FUNCTION();
--- a/Hazel/src/Hazel/Core/Application.h
+++ b/Hazel/src/Hazel/Core/Application.h
@@ -22,6 +22,7 @@ namespace Hazel {
 		virtual ~Application();
 
 		void OnEvent(Event& e);
+		void OnCustomEvent(CustomEvent& e);
 
 		void PushLayer(Layer* layer);
 		void PushOverlay(Layer* layer);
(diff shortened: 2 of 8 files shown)


CleverSource:
This is actually really awesome! Though I feel like having 2 different dispatchers is weird, maybe instead of splitting them, have a function (or struct variable) to set that will define a normal dispatcher as a custom one; which will in turn alter it's functionality a small bit?

Doing the above would remove lots of redundant code and make more sense in my opinion.

Sausty:
About the IsCustomEvent inheritance stuff, because GetEventType needs to be override and the type conflicts, it's impossible. It's also why i can't have a OnEvent function that groups hazel events and custom events.

LovelySanta:
I share the same opinions like CleverSource, 2 dispatchers is weird, it should share the same dispatcher

About the CustomEvent, that looks good, but I wonder what the difference is between CustomEvent and the ApplicationEvent class? Isn't that why the ApplicationEvent class is there? At least CustomEvent should be an ApplicationEvent since it's raised by the application, in that case it should be named as AppCustomEvent and be put in the ApplicationEvent.h file?

I feel like that would be the primary 'separation' and could be used by the dispatcher. After that, there should be another mechanism to seperate different derived classes of CustomEvent, what about the CustomEvent class providing a unique ID for each type of derived class? Maybe CustomEvent itself could then have the unique ID 0?

CleverSource:
I also believe doing what LovelySanta is stating would allow for a function such as `IsCustomEvent` which by default returns `false` but for `AppCustomEvent` will return `true` by overriding. This would allow you to remove `OnCustomEvent` from the Application class and just use `OnEvent` and alter one like of code for custom events.

I'll test this theory out later, but it sounds to me like it would work, if someone else can test this before I can please comment on your results!

retrooper (changes requested):
EventCategoryNone should remain, "None" could be a constant for any enum.

TheCherno:
We'll be moving to C# soon for scripting, so a "gameplay" style event system isn't really needed in C++ (you would just build one of these in C#). The core engine event system is meant more for system/application events, such as the ones already present, and isn't meant to be extended.

PR #359, opened 2020-10-27
Warn if user tries to add a component if already present

advikkabra:
The Add Component menu retains the Camera and Sprite Renderer component even if they are already present. If one clicks on them, then the program crashes.

The menu item does not appear if the component is already there.

### DIFF
--- a/Hazelnut/src/Panels/SceneHierarchyPanel.cpp
+++ b/Hazelnut/src/Panels/SceneHierarchyPanel.cpp
@@ -221,23 +221,23 @@ namespace Hazel {
 
 		if (ImGui::BeginPopup("AddComponent"))
 		{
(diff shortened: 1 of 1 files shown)


CybernetHacker14:
What about component dependencies - like for a certain component, there should be certain other components present on an entity as well? similar to `RequireComponent` method of Unity. Also, wouldn't it be better to have some sort of boolean, where it would just check if the boolean is true, and the component is already present on the entity or not?

LovelySanta:
I would prefer that as well. Instead of altering the menu, just check in the function if it already exists, if it exists, put a warning in the console? I the order of "Cannot add component X to entity Y (ID). Component already present"

advikkabra:
Yeah that seems like a good idea. I'll commit that right away.

PR #362, opened 2020-10-29
Fixed Z-Fighting bug in Texture Shader and updated GLSL version

Sausty:
Without this change, going 3D with a perspective camera will affect how the textures are rendered. If you don't make v_TexIndex flat, the texture will glitch out when changing the yaw of the camera. Making it flat will solve this issue. I also updated the GLSL version so that it's compatible with flat.

Edit : 
Another way to fix this issue would be to make the TexIndex attrib an integer, and instead of having glVertexAttribPointer, we would have glVertexAttribIPointer.

### DIFF
--- a/Hazelnut/assets/shaders/Texture.glsl
+++ b/Hazelnut/assets/shaders/Texture.glsl
@@ -1,7 +1,7 @@
 // Basic Texture Shader
 
 #type vertex
-#version 330 core
+#version 450
 
 layout(location = 0) in vec3 a_Position;
 layout(location = 1) in vec4 a_Color;
@@ -13,7 +13,7 @@ uniform mat4 u_ViewProjection;
 
 out vec4 v_Color;
 out vec2 v_TexCoord;
-out float v_TexIndex;
+out flat float v_TexIndex;
 out float v_TilingFactor;
 
 void main()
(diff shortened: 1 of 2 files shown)


LovelySanta:
I am gonna wait to merge this, as we're still working on 2D only. Hazelnut does support a bit of 3D stuff just to save on re-iterating over the code once we actually go 3D. For now it's pure 2D...

CybernetHacker14:
@Sausty I, once, created a textured quad by creating a texture in code, and storing it inside the sprite renderer component. When Cherno implemented the camera component, with perspective projection, when I tried to change the z position of the camera, sometimes, the texture appeared proper, sometimes the texture disappeared and instead only a flat colored quad with no texture existed. Not sure, but did this occur to you as well? And does this pull request fix it?

Sausty:
In my engine i tried to implement a 3D first person camera, and changing the Y rotation of the camera would completely bug out the texture. Going to far away of the texture would also make it disappear, and i fixed this issue by not multiplying the transform to the view projection matrix.

SamuelHyman:
Honestly, This issue bit me hard, I have done a couple hours of debugging thinking I had done something wrong in my codebase.  Both adjusting yaw on the camera in addition to rotating objects around the Y axis caused the aforementioned texture issues.

After applying the flat modifier to v_TexIndex everything cleared right up.

Does adding the flat modifier cause any issues aside from requiring a later version of OpenGL support?  Is there an alternative that can be achieved with equally trivial changes while remaining compliant with project's current version?

LovelySanta:
On further thought, it shouldn't hurt to merge this, added it to the list.

PR #363, opened 2020-10-29
Change to FileDialogs' methods so that they return std::optional.

Edward-Billington:
Not a major change, just personally think we should be using `std::optional` in this instance instead of returning and checking for an empty string.

#### PR impact

Not sure any additional context is needed here!


DaveAxiom (approved):
This is a good modern solution for removing `null` which is where programming languages are heading.

PR #364, opened 2020-10-30
Added .hazel extension to files saved when no extension has been provided.

Edward-Billington:
When saving a scene, the dialogue menu says "Save as type" (_.hazel_)
However this is untrue and will save as a generic file-type and is therefore un-findable in the "Open" dialogue.

image

This change will make files that have no provided extension to be defaulted to **.hazel**.

#### PR impact

#### Additional context

(from master branch)
Above is the "Open" dialogue
Below is Windows Explorer
image


LovelySanta:
Thanks for the effort, however, #355 is more advanced, we don't want to hardcode the default extension. I'll close this as duplicate.

Edward-Billington:
@LovelySanta didn't see that PR, no worries!

PR #370, opened 2020-11-19
Caching Transformation Matrix in TransformComponent

3x3y3z3t:
This PR update `TransformationComponent` for (hopefully) better performance.

#### PR impact

#### Proposed fix
- `TransformComponent` now caches Transformation matrix and only reconstructs it when Translation, Rotation or Scale is changed.
- Transformation matrix reconstruction only take places when it is requested (when `GetTransform()` is called).

#### Additional context
- The cost for a better performance is `TransformerComponent` now takes up more memory for caching values (+88 bytes).

### DIFF
--- a/Hazel/src/Hazel/Scene/Components.h
+++ b/Hazel/src/Hazel/Scene/Components.h
@@ -29,16 +29,42 @@ namespace Hazel {
 		TransformComponent(const glm::vec3& translation)
 			: Translation(translation) {}
 
-		glm::mat4 GetTransform() const
+		glm::mat4 GetTransform()
 		{
-			glm::mat4 rotation = glm::rotate(glm::mat4(1.0f), Rotation.x, { 1, 0, 0 })
-				* glm::rotate(glm::mat4(1.0f), Rotation.y, { 0, 1, 0 })
-				* glm::rotate(glm::mat4(1.0f), Rotation.z, { 0, 0, 1 });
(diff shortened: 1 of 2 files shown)


Fredkiss3:
In term of performance, it is better only in certains circumstances, like static elements (ex: walls), since they need to be redrawn each Frame recalculating their transform is a wate of performance for sure. But for dynamic elements (ex: particles), i think it is better to recalculate the matrix each frame instead of caching it because, we are gonna do it anyway, so it occupies an unnecessary amount of memory.

3x3y3z3t:
> In term of performance, it is better only in certains circumstances, like static elements (ex: walls), since they need to be redrawn each Frame recalculating their transform is a wate of performance for sure. But for dynamic elements (ex: particles), i think it is better to recalculate the matrix each frame instead of caching it because, we are gonna do it anyway, so it occupies an unnecessary amount of memory.

I thought particles are managed separately and will have different calculation routine?
About other moving objects, you are right, that will cost unnecessary memory.

TheCherno:
Closing because this will probably not end up being any faster (and will probably be slower). The main problem is adding 88 bytes of memory to the TransformComponent will more than triple its size - this will make the component much less efficient to iterate through and much less CPU cache friendly.

In all likelyhood, I would expect it would actually be faster to just recalculate the matrix, however I haven't benchmarked this.

PR #371, opened 2020-11-21
fixed missing font in sandbox

falconmick:
Missing fonts cause sandbox project to get assertion failure
image

Just add the missing files


DaveAxiom:
I'm not a believer in that a source code revision management tool should be used to host binary files. It is possible that small binary files are integral to a project. Copying large files or making them redundant wastes space and far too much bandwidth.

I still think there is a better solution in code that can fix this architectural problem. I was expecting the community to get a comment from Cherno. Opening an issue for this project is overdue.

falconmick:
Yeah I don’t disagree, also if this is a required font then we should probably either copy it into the exe using the asset in the core game engine OR make a post build copy step as an easy option. But just fixing it for now probably is fine, sandbox isn’t terribly important

danger89:
Well its common to see to have at least 1 font included as a **fallback** scenario. So why not include such a font by default (could be opensans or anything for that matter). And if a particular font can't be found it should not fail, but give a warning.

TheCherno:
Obsolete.

PR #379, opened 2020-11-29
Fixed unexpected behaviour in OpenGLVertexArray when using int attributes

JonathanHiggs:
glVertexAttribPointer will auto cast any int types to float between setting the buffer data and data being passed into the shader program. Even when the shader layout states the attribute is an int, the value passed has already been cast to a float, and it is then reinterpreted as an int, ie. the shader program literally sees the float exponent and mantissa bits set. This fix assumes that any Hazel users would expect to receive the same bit-wise int value if they pushed an int into the buffer and set the layout attribute to have an int type in a shader. The issue is subtle for opengl newcommers and likely to cause a lot of confusion if not patched

Impact is minimal, no code in the repo makes use of DataType::Int. Can't comment on any projects that take Hazel as a dependency, but given it does not work in the current state the new code introduced is unlikely to be hit at the moment

No issue currently opened for this fix

Fix adds splits the floating point and integer cases in the switch statement where the vertex array attribute pointers are set

This fix was derrived from the following form post and opengl documentation:
[https://community.amd.com/t5/opengl-vulkan/integer-vertex-specification-is-cast-from-glint-to-glfloat/td-p/231010]
[https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glVertexAttribPointer.xhtml]

### DIFF
--- a/Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
+++ b/Hazel/src/Platform/OpenGL/OpenGLVertexArray.cpp
@@ -72,6 +72,17 @@ namespace Hazel {
 				case ShaderDataType::Float2:
 				case ShaderDataType::Float3:
 				case ShaderDataType::Float4:
+				{
+					glEnableVertexAttribArray(m_VertexBufferIndex);
+					glVertexAttribPointer(m_VertexBufferIndex,
+						element.GetComponentCount(),
+						ShaderDataTypeToOpenGLBaseType(element.Type),
+						element.Normalized ? GL_TRUE : GL_FALSE,
+						layout.GetStride(),
+						(const void*)element.Offset);
+					m_VertexBufferIndex++;
+					break;
+				}
 				case ShaderDataType::Int:
 				case ShaderDataType::Int2:
 				case ShaderDataType::Int3:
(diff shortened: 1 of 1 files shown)


LovelySanta:
Does it still happen if you define them as 'flat' as described in #397? otherwise we can still merge this... Do you have a repo to reproduce this?

JonathanHiggs:
AFAIK they are slightly orthogonal issues. The texId is passed in as a float it seems that even with the flat keyword the pipeline is liable to floating point precission and rounding errors. Using a float and casting to int felt like a bit of a smell to me so I was investigating staying int the whole way through

My issue is that the int value in the buffer ends up going through `reinterpret_cast<int>((float)value)` which does not cast back to the original value. I have a slightly larger writeup [here](https://www.reddit.com/r/opengl/comments/k3g239/int_vertex_buffer_values_mangled_in_2d_batch/) if it helps

`flat` is still needed when passing a int/uint from vertex to fragment because it isn't interpolatable... but my patch does not change the 2d renderer buffer or shader program since I wanted to limit the changes, though arguably it should be changed

ranoke:
I have a question guys and don't know where to ask becouse on Discord channel i've got ignored. So in OpenGLVertexArray there is a m_VertexBufferIndex that gets incremented during setting the layout. So after it all done it stays with the value for example 11 and when i do change buffer it all working not properly. So does it needs to be zeroed after this func call (void OpenGLVertexArray::AddVertexBuffer(const Ref<VertexBuffer>& vertexBuffer)).

PR #388, opened 2020-12-04
Fixed Math Function Declaration

CybernetHacker14:
#### Fixed function declaration of DecomposeTransform() function
When @TheCherno copied the DecomposeTransform() function diff, the declaration he created in Math.h class was different from the definition, which could lead to some inconsistencies

#### PR impact

#### Proposed fix
Changed the declaration variable names to the ones which were copied in diff

### DIFF
--- a/Hazel/src/Hazel/Math/Math.h
+++ b/Hazel/src/Hazel/Math/Math.h
@@ -3,5 +3,7 @@
 #include <glm/glm.hpp>
 
 namespace Hazel::Math {
+
 	bool DecomposeTransform(const glm::mat4& transform, glm::vec3& translation, glm::vec3& rotation, glm::vec3& scale);
+
 }
--- a/Hazel/src/Hazel/Math/Math.h
+++ b/Hazel/src/Hazel/Math/Math.h
@@ -3,7 +3,5 @@
 #include <glm/glm.hpp>
 
 namespace Hazel::Math {
-
-	bool DecomposeTransform(const glm::mat4& transform, glm::vec3& outTranslation, glm::vec3& outRotation, glm::vec3& outScale);
-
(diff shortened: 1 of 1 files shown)


Hazel/src/Hazel/Math/Math.h
```diff
@@ -3,7 +3,5 @@
 #include <glm/glm.hpp>
 
 namespace Hazel::Math {
-
-	bool DecomposeTransform(const glm::mat4& transform, glm::vec3& outTranslation, glm::vec3& outRotation, glm::vec3& outScale);
-
+	bool DecomposeTransform(const glm::mat4& transform, glm::vec3& translation, glm::vec3& rotation, glm::vec3& scale);
```
LovelySanta, line 6: I remember he wanted to prefix them with out to indicate they are the output... In any case, you should leave the new lines in there

```suggestion

	bool DecomposeTransform(const glm::mat4& transform, glm::vec3& translation, glm::vec3& rotation, glm::vec3& scale);

```

CybernetHacker14: Yeah, he wanted to do that. I just changed that to match the definition. In case, if he wants to change it, he can change it.

teonik:
I really think it should have been the other way around, i.e. keep the "out" prefix for the arguments in the declaration and change the definition's signature to match. That way it is clear that those arguments are the output of the Hazel::Math::DecomposeTransform() function.

LovelySanta:
That's indeed how he intended it to be when he prepared the diff during the livestream (to be more explicit), however in the youtube video he didn't do this.. I did think about this, in the end I went with how he implemented it in the youtube video, which is the usual approach I take when deciding these things... It's only a minor detail, and it should be obvious from the name of the function what it does anyway...

PR #390, opened 2020-12-04
Fix crash when selecting object without camera initialized

### LINKED ISSUE #387: Crash selecting Object without a camera
#### Description
Crash when selecting Object and pressing 'W', 'E' or 'R', while there is no camera in the scene and your cursor is on the viewport

#### To Reproduce
1. Open new scene
2. Create a new empty entity
3. Click on the entity
4. Press W, E or R inside the viewport
Note: no camera should be present

#### Expected behavior
(on visual studio) The program should crash and open "vector" at line 1484 at 
`_NODISCARD size_type size()`

3x3y3z3t:
#### Describe the issue
Issue #387: Currently Hazelnut will crash when trying manipulating entities without a camera entity added.

#### PR impact

Issues this solves       | resolves #387

#### Proposed fix 
- Hazelnut will now check if the primary camera is initialized. If not, manipulating entities will be disabled (no gizmos, no WASD...).

#### Additional context
- The helper function `IsValid()` is added to check if an entity is valid. Checking against `entt::null` is not enough since EnTT recycles entity handles.
- Another approach could be creating new scene every time Hazelnut is started (calling `NewScene()` on attaching EditorLayer), so a camera entity is always present.

### DIFF
--- a/Hazel/src/Hazel/Scene/Entity.h
+++ b/Hazel/src/Hazel/Scene/Entity.h
@@ -42,6 +42,14 @@ namespace Hazel {
 			m_Scene->m_Registry.remove<T>(m_EntityHandle);
 		}
 
+		bool IsValid() const
+		{
+			if (m_EntityHandle == entt::null) return false;
+			if (m_Scene == nullptr) return false;
+
+			return m_Scene->m_Registry.valid(m_EntityHandle);
+		}
+
 		operator bool() const { return m_EntityHandle != entt::null; }
 		operator entt::entity() const { return m_EntityHandle; }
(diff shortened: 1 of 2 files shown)


LovelySanta:
I am not a 100% sure if this actually needs to be fixed... there should always be at least an editor camera... so it should first check if a camera is primary, if none found, use the editor camera... I am unsure when cherno wants to implement this...

LovelySanta:
This is fixed in https://github.com/TheCherno/Hazel/commit/96c57dadf5c3616ba3613034dfb63d3918590359 by adding an editor camera. In runtime, there is no editor camera, but then there are no imguizmo's either

PR #391, opened 2020-12-04
Application will now use Nvidia card if available

CybernetHacker14:
#### Application uses integrated graphics by default
Even if an Nvidia card is available, the application uses integrated graphics solution like Intel UHD graphics. We need to change the settings in Nvidia Control Panel for that particular application.

#### PR impact

#### Proposed fix
In this [post](https://www.reddit.com/r/gamedev/comments/bk7xbe/psa_for_anyone_developing_a_gameengine_in_c/) mentioned on Cherno's Discord Server, a solution is stated regarding this problem. The highlighted code block in that post is added to `HazelnutApp.cpp`. This enables the application to launch with the high-performance GPU instead of the integrated one.

#### Additional context
As stated in the above mentioned post, this solution works only for Windows. Also, I have tested it on my fork, and it works, but it may need some additional testing on other machines. 

Reference link : https://www.reddit.com/r/gamedev/comments/bk7xbe/psa_for_anyone_developing_a_gameengine_in_c/

### DIFF
--- a/Hazel/src/Hazel/Core/EntryPoint.h
+++ b/Hazel/src/Hazel/Core/EntryPoint.h
@@ -11,8 +11,10 @@ extern "C" {
 	__declspec(dllexport) uint32_t NvOptimusEnablement = 0x00000001;
 }
 
-int main(int argc, char** argv) {
-=======
+extern "C" {
+	__declspec(dllexport) DWORD NvOptimusEnablement = 0x00000001;
+}
+
 int main(int argc, char** argv)
 {
 >>>>>>> 9ee977d... Updated as per @lovely_santa 's suggestion
--- a/Hazel/src/Hazel/Core/EntryPoint.h
+++ b/Hazel/src/Hazel/Core/EntryPoint.h
@@ -16,7 +16,7 @@ extern "C" {
 }
 
 extern "C" {
-	__declspec(dllexport) DWORD NvOptimusEnablement = 0x00000001;
+	__declspec(dllexport) uint32_t NvOptimusEnablement = 0x00000001;
 }
 
 int main(int argc, char** argv)
--- a/Hazel/src/Hazel/Core/EntryPoint.h
+++ b/Hazel/src/Hazel/Core/EntryPoint.h
@@ -15,6 +15,10 @@ extern "C" {
 	__declspec(dllexport) uint32_t NvOptimusEnablement = 0x00000001;
 }
 
+extern "C" {
+	__declspec(dllexport) DWORD NvOptimusEnablement = 0x00000001;
+}
+
 int main(int argc, char** argv)
 {
 >>>>>>> 9ee977d... Updated as per @lovely_santa 's suggestion
--- a/Hazel/src/Hazel/Core/EntryPoint.h
+++ b/Hazel/src/Hazel/Core/EntryPoint.h
@@ -6,18 +6,8 @@
 
 extern Hazel::Application* Hazel::CreateApplication(ApplicationCommandLineArgs args);
 
-<<<<<<< HEAD
-extern "C" {
-	__declspec(dllexport) uint32_t NvOptimusEnablement = 0x00000001;
-}
-
-extern "C" {
-	__declspec(dllexport) uint32_t NvOptimusEnablement = 0x00000001;
-}
-
 int main(int argc, char** argv)
 {
->>>>>>> 9ee977d... Updated as per @lovely_santa 's suggestion
 	Hazel::Log::Init();
 
 	HZ_PROFILE_BEGIN_SESSION("Startup", "HazelProfile-Startup.json");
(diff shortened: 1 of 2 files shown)


Hazel/src/Hazel/Core/EntryPoint.h
```diff
@@ -5,6 +5,10 @@
 
 extern Hazel::Application* Hazel::CreateApplication();
 
+extern "C" {
+	__declspec(dllexport) DWORD NvOptimusEnablement = 0x00000001;
```
caelwithcats, line 9: please use uint32_t instead of DWORD. `__declspec(dllexport)` is MSVC specific, you should use `export`

caelwithcats:
This is windows specific. Add this around it:
 ```c++
#ifdef HZ_PLATFORM_WINDOWS

#endif
```

CybernetHacker14:
> This is windows specific. Add this around it:
> 
> ```c++
> #ifdef HZ_PLATFORM_WINDOWS
> 
> #endif
> ```

Isn't that already in that block?

Hazel/src/Hazel/Core/EntryPoint.h
CybernetHacker14: Done

caelwithcats:
do a `git rebase` to clean up the commits

CybernetHacker14:
> do a `git rebase` to clean up the commits

Done

LovelySanta:
I do know some people that have a good CPU, but just an old GPU that can't use Hazel, maybe we should wrap it around an `#if defined(HZ_FORCE_NVIDIA)` for convenience?

CybernetHacker14:
> I do know some people that have a good CPU, but just an old GPU that can't use Hazel, maybe we should wrap it around an `#if defined(HZ_FORCE_NVIDIA)` for convenience? Any thoughts?

We can, but how will we detect if there's an nVidia GPU available automatically?
If there's a way, I can add it to `PlatformDetection.h` script. But then again, the person wanting to use the GPU will have to edit 
the premake script to add the define.

LovelySanta:
> If there's a way, I can add it to `PlatformDetection.h` script. But then again, the person wanting to use the GPU will have to edit
> the premake script to add the define.

There is no way of knowing at compile time...

CybernetHacker14:
> > If there's a way, I can add it to `PlatformDetection.h` script. But then again, the person wanting to use the GPU will have to edit
> > the premake script to add the define.
> 
> There is no way of knowing at compile time...

So, are there any recommendations? What should be changed? According to me, we can let Hazel launch through the GPU by default, and let users switch the hardware through their control panel. Plus, if the machine doesn't have any GPU and a good CPU, I guess the engine will still work on that.

LovelySanta:
After some digging (and I dug quite a deep hole), it seems like openGL doesn't allow you to select a GPU at all, the only thing is that you can let your GPU prefer to render it over the CPU, as you did. [This stackexchange post](https://gamedev.stackexchange.com/questions/72505/why-does-my-machine-render-opengl-using-my-onboard-chipset-instead-of-my-graphic) is the only source that I found, which indirectly lets GLFW set this for you, again, at compile time hardcoded. There is no such thing as by vulkan that you can select/preference a physical device (obviously).

On that note, depending on the application you're writing you might want to use the CPU or GPU. For high end simulations you want the CPU to render the application, such that the GPU can do the heavy lifting calculation wise. And since Hazel is an application engine (not only for making games), it feels unreasonable to let the engine force this setting. This should be something for the game to do.

So my suggestion would be, we could define these symbols in the 'Game' rather than in the engine. I mean by this define it in Sandbox and Hazelnut, as in both cases it is reasonable (obviously) to use the GPU to render the graphics, while leaving the option to the application/game to decide to enforce this or not.

CybernetHacker14:
> After some digging (and I dug quite a deep hole), it seems like openGL doesn't allow you to select a GPU at all, the only thing is that you can let your GPU prefer to render it over the CPU, as you did. [This stackexchange post](https://gamedev.stackexchange.com/questions/72505/why-does-my-machine-render-opengl-using-my-onboard-chipset-instead-of-my-graphic) is the only source that I found, which indirectly lets GLFW set this for you, again, at compile time hardcoded. There is no such thing as by vulkan that you can select/preference a physical device (obviously).
> 
> On that note, depending on the application you're writing you might want to use the CPU or GPU. For high end simulations you want the CPU to render the application, such that the GPU can do the heavy lifting calculation wise. And since Hazel is an application engine (not only for making games), it feels unreasonable to let the engine force this setting. This should be something for the game to do.
> 
> So my suggestion would be, we could define these symbols in the 'Game' rather than in the engine. I mean by this define it in Sandbox and Hazelnut, as in both cases it is reasonable (obviously) to use the GPU to render the graphics, while leaving the option to the application/game to decide to enforce this or not.

Sorry for the late reply, was caught up in something. Okay, will do this.

CybernetHacker14:
Will this pull request be reviewed again @caelwithcats @LovelySanta ?

CybernetHacker14:
Can anyone please help me fix this stuff? The various merge commits again have caused an issue.

TheCherno:
Obsolete (switch to Vulkan and deprecating OpenGL soon).

PR #392, opened 2020-12-07
fixed crashes when using the gizmo without a camera

### LINKED ISSUE #387: Crash selecting Object without a camera

mandar1jn:
Fixed that HazelNut crashes when trying to use the gizmos without a camera in the scene

Issues this solves       | Closes #387 

#### Proposed fix 
It checks if there is a valid camera in the scene and if the scene is not null. If these both pass the IsValid function returns true. If one of them fails the IsValid function returns false and the gizmo isn't executed

### DIFF
--- a/Hazel/src/Hazel/Scene/Entity.h
+++ b/Hazel/src/Hazel/Scene/Entity.h
@@ -42,6 +42,13 @@ namespace Hazel {
 			m_Scene->m_Registry.remove<T>(m_EntityHandle);
 		}
 
+		bool IsValid() {
+			if (m_EntityHandle == entt::null) return false;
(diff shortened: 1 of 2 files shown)


LovelySanta:
This implementation is 99% equivalent to #390. Closing this as duplicate

PR #395, opened 2020-12-09
Added Random number generator

FahimFuad:
### Description
Hazel is missing a random number generator. This Pull Request adds a random number generator class, which is a header only random number generator. This currently supports 4 data types, `int`, `float`, `double` and `bool`. It is a templated static function which is  super easy to use!

### DIFF
--- a/Hazel/src/Hazel.h
+++ b/Hazel/src/Hazel.h
@@ -22,6 +22,7 @@
 #include "Hazel/Scene/Entity.h"
 #include "Hazel/Scene/ScriptableEntity.h"
 #include "Hazel/Scene/Components.h"
+#include "Hazel/Utils/Random.h"
 
 // ---Renderer------------------------
 #include "Hazel/Renderer/Renderer.h"
--- a/Hazel/src/Hazel/Utils/Random.h
+++ b/Hazel/src/Hazel/Utils/Random.h
@@ -1,0 +1,73 @@
+#pragma once
+#include <random>
+
+namespace Hazel
+{
+	class Random
+	{
+	public:
+		template<typename T>
+		static T HZRandom(T min = 0, T max = 100)
+		{
+			if constexpr (std::is_same_v<T, int>)
+			{
+				if (min > max)
+				{
+					std::swap(min, max);
+					HZ_CORE_WARN("Minimum is greater than maximum in {0}", __FUNCTION__);
+				}
+
+				std::uniform_int_distribution<> distribution(min, max);
+				auto& mt = Engine();
+				return distribution(mt);
+			}
+
+			if constexpr (std::is_same_v<T, float>)
+			{
+				if (min > max)
+				{
+					std::swap(min, max);
+					HZ_CORE_WARN("Minimum is greater than maximum in {0}", __FUNCTION__);
+				}
+
+				std::uniform_real_distribution<T> distribution(min, max);
+				auto& mt = Engine();
+				return distribution(mt);
+			}
+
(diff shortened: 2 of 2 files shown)


CybernetHacker14:
Does this return duplicate values? Or does it work like a UUID generator?

FahimFuad:
@CybernetHacker14  Yes this return duplicate values, depends on your luck! 🙂

3x3y3z3t:
Possibly duplicate with #225, but it seems that the other author have abandoned his PR...

csp256:
I haven't tried it but it looks reasonable enough for a start. 

As a nitpick, I'd prefer to name the static function something other than `HZRandom()`, but I've been a bit out of the loop on Hazel and not sure if that follows the naming conventions. Why not something like `Random::Uniform()` or `Random::Sample()`?

I think providing default values to min and max is a little dangerous. I would expect `Random<int>(20)` to return something from 0..20, not 20..100. I also would not expect `Random()` to return something other than a random value from either the unit interval or the full range.

As an aside, I clicked on this PR originally wondering if Hazel had a vector3 class yet, because functions like "random point on sphere surface" and whatnot are very useful in games and should be added. I see that's not the case...

Lecrapouille:
Hi guys, I was watching PR from latest to newer, so I did not see this topic first and this is the reason why I added comments on #225 but that can be in relation with this PR. But beware of your comments: you should inform if your random function return a number within `[min max[` or `[min max]`

FahimFuad:
@Lecrapouille It's inclusive.

csp256:
Isn't the standard a half open interval? (excludes max) It is certainly the default behavior I've come to expect.

FahimFuad:
Okay I am making it as the half open interval 😊. @LovelySanta is the function `HZRandom<>()` name suits for this? I don't perfectly know what convention or naming style of static functions is preferred by you guys.

LovelySanta:
It's always tricky to fullfill everyones needs...

1. If it was up to me, I would first of all let the Random class be the thing you use, not the distribution you get from it which you then use to sample... So every instance of Random would result in independent random variables, the constructor of `Hazel::Random()` would allow for a seed, which is also usefull, think about deterministic behaviour in multiplayer, where random is not so random, as it must be the same for all players...

2. Have functions (+overloaders) such that for example I would have a function `double Hazel::Random::Uniform()` as a default, which is uniform in [0,1[, then `T Hazel::Random::Uniform(T max)` in [0, max[ and `T Hazel::Random::Uniform(T min, T max)` in [min, max[. A basic example without boilerplate code around it is shown below.
```cpp
double Hazel::Random::Uniform() {...}

template<typename T>
T Hazel::Random::Uniform(T max)
{
    return max * Uniform();
}

template<typename T>
T Hazel::Random::Uniform(T max)
{
    return min + Uniform(max - min);
}
```

3. Back to your code... I would then have different functions for bernoulli and maybe a normal distribution as well?
In any case, the last if constexpr at line 61 can be ommited, the other constexpr are returning early...

That's how I would do it as a start, after that, start optimizing with performance tests

FahimFuad:
@LovelySanta  Will I change the code? 🤔

LovelySanta:
I don't feel it's ready in its current state, keep in mind this is not the smallest project either as it can see some intense use when going for procedural generation for example, however I do feel it could be a useful addition, maybe not on it's own, but with some uses, yes, for sure.

FahimFuad:
I will try to implement these later if I get time, I have exams now and I don't want a junk PR in Hazel's Open Pull Requests

PR #396, opened 2020-12-11
add audio and process manager

rocketman123456:
this is the code use in my project, which has so many change compare with your origin code, so it might have many problems, but i hope it will help, and add sounds in your game engine. Actually i made a simple piano example in my project using these codes

### DIFF
--- a/Hazel/src/Hazel/Core/AudioManager.h
+++ b/Hazel/src/Hazel/Core/AudioManager.h
@@ -1,0 +1,36 @@
+#pragma once
+
+#include <AL/al.h>
+#include <AL/alc.h>
+#include <AL/alext.h>
+#include <sndfile.h>
+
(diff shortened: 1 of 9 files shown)


AzadKshitij:
@TheCherno has created his own library for audio for Hazel and he is going to use that instead of OpenAL.

LovelySanta:
> @TheCherno has created his own library for audio for Hazel and he is going to use that instead of OpenAL.

That's actually not entirely true.. Cherno wrote already a wrapper around it to use inside Hazel: [HazelAudio](https://github.com/TheCherno/HazelAudio), which uses [OpenAL Soft](https://openal-soft.org/)

LovelySanta:
I would put it on hold, but since it will be quite a while before cherno get to it, I will close this instead.

PR #398, opened 2020-12-11
Ref now get's created using CreateRef() in Log class

CybernetHacker14:
#### Describe the issue
In Log class, `s_CoreLogger` and `s_ClientLogger` are Refs, an alias for `std::shared_ptr`. However they are getting created using `std::make_shared` instead of `CreateRef()` in the Log class. This leads to code convention inconsistency.

#### PR impact

#### Proposed fix
Replaced `std::make_shared` with `CreateRef()`.

### DIFF
--- a/Hazel/src/Hazel/Core/Log.cpp
+++ b/Hazel/src/Hazel/Core/Log.cpp
@@ -5,7 +5,6 @@
 #include <spdlog/sinks/basic_file_sink.h>
 
 namespace Hazel {
-
 	Ref<spdlog::logger> Log::s_CoreLogger;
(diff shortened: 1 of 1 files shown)


LovelySanta:
This is logging code for debugging purposes, for this reason cherno did not want to tie this to anything .

peter1745:
@LovelySanta @CybernetHacker14 I believe he also kept it that way to allow us to implement or own version of shared_ptr, just like we did in Hazel-dev, which would be incompatible with spdlogs functions. Just wanted to mention that 🙂

PR #400, opened 2020-12-16
ImGuizmo Crash Fix

### LINKED ISSUE #383: Crash when changing gizmos while using them
#### Description
Found an issue, if you change gizmos while using it, it will crash

#### My fix
i fixed it by locking the changing when i use it. not the best but it works
Im opening this issue, because i know that my fix is not the best and can be improved

#### Reproduce
To Reproduce: Try moving an object and change the gizmos to rotation / scale while you're moving it

twarug:
The issue

When you change a gizmo while using it, it will crash entire engine.

- Crash when changing gizmos while using them #383

Issues this solves       | Resolves #383

Proposed fix:

Disable and Enabel ImGuizmos after gizmo change.

### DIFF
--- a/Hazelnut/src/EditorLayer.cpp
+++ b/Hazelnut/src/EditorLayer.cpp
@@ -303,8 +303,11 @@ namespace Hazel {
 		if (e.GetRepeatCount() > 0)
 			return false;
 
+		int gizmoChange = -2;
+
 		bool control = Input::IsKeyPressed(Key::LeftControl) || Input::IsKeyPressed(Key::RightControl);
 		bool shift = Input::IsKeyPressed(Key::LeftShift) || Input::IsKeyPressed(Key::RightShift);
+
 		switch (e.GetKeyCode())
 		{
 			case Key::N:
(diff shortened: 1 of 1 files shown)


LovelySanta:
I find this more of a 'workaround' than an actual fix... When you're 'using' (read as fine tuning) an ImGuizmo I assume that pressing the switch buttons is rather an accident compared to the oposite (switching intended and 'using' an accident).

For this reason I would prefer to implement it as follows, when pressing the keys, check for `ImGuizmo::IsUsing()` and only switch modes when it is not being used?

I feel like that's a cleaner solution rather than what is proposed here? Feel free to prove me wrong?

twarug:
I think that your solution is better. I decide to solve it in that way, becouse that is behaviour in Unity, wich i use a lot. I can iplement your solution? or cloase a pull request.

FahimFuad:
@twarug You can just make another commit, and the changes will be applied automatically 🙂

Hazelnut/src/EditorLayer.cpp
```diff
@@ -303,7 +303,7 @@ namespace Hazel {
 		if (e.GetRepeatCount() > 0)
 			return false;
 
-		int gizmoChange = -2;
+		bool canGizmoChange = !ImGuizmo::IsUsing();
```
LovelySanta, line 306: no need to store it as a variable, it is only used in the switch case anyway (once)

```suggestion

```

twarug: yes, but in the future can, be more conditions to change gizmo

LovelySanta: We can change it in the future then.. now it's just bloating the code...

PR #412, opened 2021-01-23
Assert single instance of ImGuiLayer

MHamiid:
Multiple instances of ImGuiLayer could be created and pushed into the Application's LayerStack, resulting in ImGui being initialized more than once in OnAttach function.

#### Proposed fix 
Assert single instance in ImGuiLayer's constructor.

### DIFF
--- a/Hazel/src/Hazel/ImGui/ImGuiLayer.cpp
+++ b/Hazel/src/Hazel/ImGui/ImGuiLayer.cpp
@@ -15,6 +15,8 @@
 
 namespace Hazel {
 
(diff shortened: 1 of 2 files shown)


LovelySanta:
Assume you would create 2 layerstacks for 2 different windows, you could add one imgui layer to each stack? I don't see a practical use, but it could be possible?

In that aspect, wouldn't it be a solution to only initialize it once? Such that the 2nd instance would still work, rather than not allowing it?

TheCherno:
Unnecessary.

PR #415, opened 2021-02-03
Fixes #380 and Removes Unused Variables in SandboxApp

### LINKED ISSUE #380: Font Paths in Hazel Breaks Sandbox Project
[Relative font paths](https://github.com/TheCherno/Hazel/commit/aab43e62129abc869d00ee508b4467bbe1c4f319#diff-b63db334a2644ff74fc7cb786b8c8fbea0b4e4219129b7a624171e8c47030499R37) that only work in the *Hazelnut* project were added to the *Hazel* library! Hazelnut compiles and runs fine but the template project *Sandbox* terminates with an error as it can't find the asset font files.

The adding of Hazelnut specific file paths into Hazel is an architectural problem. The problem code needs to be moved into Hazelnut as either a modification of the `ImGuiLayer` object, or move the font file paths into Hazelnut that would invoke a file asset manager in Hazel.

### LINKED ISSUE #414: Sandbox App crashes on boot.
#### Describe the bug
When loading the default Sandbox application, it crashes as ImGui cannot find the font file.

#### To Reproduce
1. Run the default Sandbox Application

#### Expected behavior
Run normally with the checkered background and quads.

#### Screenshots
N/A

- OS: win-64
- Version: Windows 10 home

lochnessdragon:
Since there hasn't been a major change to the Sandbox Application, I cleaned up temporary variables to orchestrate a concise and clean working environment. I also added the font files so that the Sanbox Application will run out of the box.

#### PR impact

Fixes Sandbox App Crashes on Boot       | #380 
Removes unused Sandbox App variables    | None

#### Proposed fix 
1. Copy the font file from the Hazelnut project into the Sandbox App.
2. Remove the unused variables: m_SquareVA and m_FlatColorShader

#### Additional context
Tested In:
- [x] Debug - x64
- [x] Release - x64 
- [x] Dist - x64


DaveAxiom:
This patch fixes the problem by duplicating Hazelnut's font files. I don't believe this is a proper fix because Git isn't a file repository.

LovelySanta:
> This patch fixes the problem by duplicating Hazelnut's font files. I don't believe this is a proper fix because Git isn't a file repository.

As long as we do not have a file manager, this is the only option, as the different applications do not 'share' a directory with the engine... the file mananger could work around this...

TheCherno:
This will be fixed in the future by setting the working directory programatically.

PR #416, opened 2021-02-06
Added build guide

HP1324:
There was no explicit guide about how to build the Hazel Engine , so I added it.


LovelySanta:
Since the build process has changed, this is outdated. Since I have closer conversations with the author of #442 I'll close this as duplicate rather than closing #442. Feel free to update this and create a new PR anyway. Thanks for the effort. 👍

PR #423, opened 2021-02-17
Move Assets Out of the Hazel Library!

### LINKED ISSUE #380: Font Paths in Hazel Breaks Sandbox Project

DaveAxiom:
The font files included in Hazelnut's sub-project directory are initialized inside of the Hazel library. When Sandbox is run, a different working directory is used and thus, the file paths for Hazelnut are invalid.

Issues this solves       | Resolves #380
Other PRs this solves    | Closes #415, this PR doesn't duplicate files to fix the problem.

The short of it is that this PR moves a few lines from the Hazel sub-project into Hazelnut. Sandbox is left alone from this invalid path.

#### Additional context
This patch worked on my Linux development branch. Please verify this patch is valid on Windows. :)

EDIT by @LovelySanta: Updated PR layout to use closing keywords


LovelySanta (commented):
My concern is that we will have ImGui code all over the code base. For the ImGui Rendering we tried to contain this inside the `OnImGuiRendering()` function.

However, I do like the idea of moving the font loading to the EditorLayer, since this is th place where we will use it. So I was thinking of how we can remove the ImGui code while 'adding' the fonts inside the EditorLayer still. So this is what I came up with:

We have already a function `Application::GetImGuiLayer()`. The ImGuiLayer already has a theme function, called `ImGuiLayer::SetDarkThemeColors();`. My idea was to add additional functionality to the ImGuiLayer by adding a function to load a TTF font:
```cpp
void ImGuiLayer::AddFontFromFileTTF(std::string filepath, float fontSize, bool setAsDefaultFont = false)
{
  if(setAsDefaultFont)
  {
    ImGuiIO& io = ImGui::GetIO();
    io.FontDefault = io.Fonts->AddFontFromFileTTF(filepath, fontSize);
  }
  else
    ImGui::GetIO().Fonts->AddFontFromFileTTF(filepath, fontSize);
}
```

On the EditorLayer we can add the font now indirectly, same place as you did before:
```cpp
auto imGuiLayer = Application::Get().GetImGuiLayer()
imGuiLayer.AddFontFromFileTTF("assets/fonts/opensans/OpenSans-Bold.ttf", 18.0f);
imGuiLayer.AddFontFromFileTTF("assets/fonts/opensans/OpenSans-Regular.ttf", 18.0f, true);
```

This suggestion allows the custom ImGui configuration outside of the ImGuiLayer while the ImGui code itself is still contained inside the ImGuiLayer.

TheCherno:
Will be fixed as part of issue #561

PR #424, opened 2021-02-18
Replaced premake5.exe with newer version

Halbjobbit:
I wanted to play around with some C++20 features and noticed that you cannot generate C++20 solutions with the current premake binary. This also possible with the most recent release on the premake site, so I compiled it manually form the most recent commit (1b7d93e).

It just is a convenience thing for now, but helps when maybe some C++20 features are used in the future.


LovelySanta:
I'll close this as #441 will download the (latest) binaries to the correct location during the install process.

PR #403, opened 2020-12-28
Texture and Tiling support in SpriteRendererComponent along with Seri…

MohitSethi99:
…alization and Deserialization of Scene.

- Also added ImGui::ImageButton in SpriteRendererComponent UI which displays currently applied texture just like Unreal Engine.

PR #425, opened 2021-02-23
Implemented Texture UVs Offset for SpriteComponent

lucho1:
It follows the same implementation than texture tilling of #403

It also includes an extension of the scene serializer to read glm::vec2 (required for doing the serialization of the UVOffset variable, but I can change it to 2 separated floats)

It's no issue actually, it's just a feature consisting on adding a vec2 acting as an X & Y offset for the UVs of a texture in the Texture shader, as it can be common in for sprites.

#### Additional context
I tested it on Windows 10 (my PC), I just loaded a square with a texture and works perfectly in my machine (including the save & load)

### DIFF
--- a/Hazel/src/Hazel/Renderer/Renderer2D.cpp
+++ b/Hazel/src/Hazel/Renderer/Renderer2D.cpp
@@ -16,6 +16,7 @@ namespace Hazel {
 		glm::vec2 TexCoord;
 		float TexIndex;
 		float TilingFactor;
+		glm::vec2 UVOffsetFactor;
 	};
 
 	struct Renderer2DData
@@ -56,7 +57,8 @@ namespace Hazel {
 			{ ShaderDataType::Float4, "a_Color" },
 			{ ShaderDataType::Float2, "a_TexCoord" },
 			{ ShaderDataType::Float, "a_TexIndex" },
-			{ ShaderDataType::Float, "a_TilingFactor" }
+			{ ShaderDataType::Float, "a_TilingFactor" },
+			{ ShaderDataType::Float2, "a_UVOffsetFactor" }
 		});
 		s_Data.QuadVertexArray->AddVertexBuffer(s_Data.QuadVertexBuffer);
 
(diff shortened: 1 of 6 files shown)


Hazel/src/Hazel/Renderer/Renderer2D.cpp
```diff
@@ -56,7 +57,8 @@ namespace Hazel {
 			{ ShaderDataType::Float4, "a_Color" },
 			{ ShaderDataType::Float2, "a_TexCoord" },
 			{ ShaderDataType::Float, "a_TexIndex" },
-			{ ShaderDataType::Float, "a_TilingFactor" }
+			{ ShaderDataType::Float, "a_TilingFactor" },
+			{ ShaderDataType::Float2, "a_UVOffsetFactor" }
```
lfilipozzi, line 61: I don't think you need to modify the shader and buffer layout, the texture coordinate could be offset in the renderer directly with `s_Data.QuadVertexBufferPtr->TexCoord = textureCoords[i] - uvOffsetFactor;`

lucho1: Oh yes, that's true, I don't know how I didn't see that

Hazel/src/Hazel/Renderer/Renderer2D.cpp
```diff
@@ -196,19 +196,19 @@ namespace Hazel {
 		DrawQuad(transform, color);
 	}
 
-	void Renderer2D::DrawQuad(const glm::vec2& position, const glm::vec2& size, const Ref<Texture2D>& texture, float tilingFactor, const glm::vec4& tintColor)
+	void Renderer2D::DrawQuad(const glm::vec2& position, const glm::vec2& size, const Ref<Texture2D>& texture, float tilingFactor, const glm::vec4& tintColor, const glm::vec2& texture_uvoffset)
```
LovelySanta, line 199: Naming convention (change everywhere).
```suggestion
	void Renderer2D::DrawQuad(const glm::vec2& position, const glm::vec2& size, const Ref<Texture2D>& texture, float tilingFactor, const glm::vec4& tintColor, const glm::vec2& textureUVOffset)
```

Hazel/src/Hazel/Renderer/Renderer2D.cpp
```diff
@@ -219,6 +219,7 @@ namespace Hazel {
 		const float textureIndex = 0.0f; // White Texture
 		constexpr glm::vec2 textureCoords[] = { { 0.0f, 0.0f }, { 1.0f, 0.0f }, { 1.0f, 1.0f }, { 0.0f, 1.0f } };
 		const float tilingFactor = 1.0f;
+		const glm::vec2 uvOffsetFactor = { 0.0f, 0.0f };
```
LovelySanta, line 222: In the other DrawQuad function you called it `texture_uvoffset` and here you call it `uvOffsetFactor`. Unify them to one and the same name. I suggest `textureUVOffset` as this is what you used in the YAML functions.

Hazelnut/src/Panels/SceneHierarchyPanel.cpp
```diff
@@ -323,6 +323,15 @@ namespace Hazel {
...
 			ImGui::ColorEdit4("Color", glm::value_ptr(component.Color));
+
+			// Texture UVs Offset
+			ImGui::Text("UV Offset");
+			ImGui::SameLine(ImGui::GetContentRegionAvail().x * 0.5f);
+			ImGui::SetNextItemWidth(100.0f);
+			ImGui::DragFloat("##UVOffsetX", &component.TextureUVOffset.x, 0.1f, 0.0f, 0.0f, "%.2f");
+			ImGui::SameLine();
+			ImGui::SetNextItemWidth(100.0f);
+			ImGui::DragFloat("##UVOffsetY", &component.TextureUVOffset.y, 0.1f, 0.0f, 0.0f, "%.2f");
```
LovelySanta, line 334: I suggest the same approach as cherno did, create a separate function, similar to [`DrawVec3Control`](https://github.com/TheCherno/Hazel/blob/d14d33fb9c4c49b049a8a467aab823a1a82ad33f/Hazelnut/src/Panels/SceneHierarchyPanel.cpp#L102-L166):
```cpp
static void DrawVec2Control(const std::string& label, glm::vec2& values, float resetValue = 0.0f, float columnWidth = 100.0f)
```

Hazel/src/Hazel/Scene/SceneSerializer.cpp
```diff
@@ -239,6 +270,10 @@ namespace Hazel {
 				{
 					auto& src = deserializedEntity.AddComponent<SpriteRendererComponent>();
 					src.Color = spriteRendererComponent["Color"].as<glm::vec4>();
+
+					auto uvoff_node = spriteRendererComponent["TextureUVOffset"];
+					if(uvoff_node)
+						src.TextureUVOffset = spriteRendererComponent["TextureUVOffset"].as<glm::vec2>();
```
LovelySanta, line 276: No need to check if its valid, it will always be serialized, it will only error on existing saved scenes, this PR should update those as well, rather than checking, for sure since we don't have a released version yet.
```suggestion
					src.TextureUVOffset = spriteRendererComponent["TextureUVOffset"].as<glm::vec2>();
```

LovelySanta (changes requested):

PR #428, opened 2021-03-14
Unnecessary need for a cpp file

ChristianPanov:


3x3y3z3t:
Please correct me if I'm wrong, but isn't that static member cannot be initialize in the header file like this?

ShenMian:
> Please correct me if I'm wrong, but isn't that static member cannot be initialized in the header file like this?

I think it's OK to write like this. Inline has the function of the weak symbol. So the linker only links one implementation.

LovelySanta:
At least the removal of the cpp file should be part of this PR.

TheCherno:
You remind me of parties 🥳 

But in all seriousness, you didn't even remove the file in your commit.

PR #435, opened 2021-04-01
Move code to default case in switch statement instead of leaving it outside of the statement

IMBeniamin:
The code which handles the case in which a valid rendering API has not been selected is placed outside of the switch block.

Code in function is not consistent       | None

Use the _default_ case of the switch statement to control the behaviour if the value is incorrect.

#### Additional context
This solution adds elegance, makes the code easier to read and perhaps allows the compiler to better optimize it.

### DIFF
--- a/Hazel/src/Hazel/Renderer/Buffer.cpp
+++ b/Hazel/src/Hazel/Renderer/Buffer.cpp
@@ -13,10 +13,8 @@ namespace Hazel {
 		{
 			case RendererAPI::API::None:    HZ_CORE_ASSERT(false, "RendererAPI::None is currently not supported!"); return nullptr;
 			case RendererAPI::API::OpenGL:  return CreateRef<OpenGLVertexBuffer>(size);
+			default:                        HZ_CORE_ASSERT(false, "Unknown RendererAPI!"); return nullptr;
 		}
-
-		HZ_CORE_ASSERT(false, "Unknown RendererAPI!");
-		return nullptr;
 	}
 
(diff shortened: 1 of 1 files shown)


salimp2009:
Althought it may not matter in this case,
 i believe leaving "invalid Renderer API case" outside the switch statement and not using "default" in switch statement is better if the logic is correct; because
- all other options of switch cases return from the function if it matches  if not then we already cover the default value and return nullptr
- and there are some notes with Jason Turners book (c++ best practices) about not use the default in switch statements that returns (if possible ); it is mostly about compiler not catching when all options in switch are not evaluate if default is used ; he gives this as an example 
"
Unhandled case; 
enum class Values {
val1,
val2,
val3 // added a new value
};
std::string_view get_name(Values value) {
switch (value) {
case Values::val1: return "val1";
case Values::val2: return "val2";
default: return "unknown";
}
// no compiler diagnostic that `val3` is unhandled
}
Instead, prefer code like this:
Avoid default In switch Statements 56
Prefered version
enum class Values {
val1,
val2,
val3 // added a new value
};
std::string_view get_name(Values value) {
switch (value) {
case Values::val1: return "val1";
case Values::val2: return "val2";
} // unhandled enum value warning now
return "unknown";
}
"

LovelySanta:
It will also warn you if there are cases not handled in the switch statement, where it will not if you use the default case. I'll mark this as a suggestion, but i'll close it as this has no (significant) impact at this point.

PR #437, opened 2021-05-02
Added post-build commands for copying Vulkan .dlls to output directory

CybernetHacker14:
#### Describe the issue
After pulling the repository, if we try running Hazelnut, the application crashes because the required `.dll` files at `vendor\VulkanSDK\Bin` are not in proper place.

#### PR impact

#### Proposed fix
A hacky solution without changing any implementation is to manually copy and paste the files in the output directory.
A somewhat automated solution is to use the premake's `postbuildcommand` to copy the files.

#### Additional context
Not the most robust or optimal solution, could be better.

### DIFF
--- a/Dependencies.lua
+++ b/Dependencies.lua
@@ -20,6 +20,7 @@ LibraryDir = {}
 
 LibraryDir["VulkanSDK"] = "%{VULKAN_SDK}/Lib"
 LibraryDir["VulkanSDK_Debug"] = "%{wks.location}/Hazel/vendor/VulkanSDK/Lib"
+LibraryDir["VulkanSDK_DLL"] = "%{wks.location}/Hazel/vendor/VulkanSDK/Bin"
 
 Library = {}
 Library["Vulkan"] = "%{LibraryDir.VulkanSDK}/vulkan-1.lib"
--- a/Hazelnut/premake5.lua
+++ b/Hazelnut/premake5.lua
@@ -27,6 +27,11 @@ project "Hazelnut"
 	{
 		"Hazel"
 	}
+	
+	postbuildcommands
+	{
+		"{COPY} %{LibraryDir.VulkanSDK_DLL} %{wks.location}/bin/" .. outputdir .. "/%{prj.name}"
+	}
 
 	filter "system:windows"
 		systemversion "latest"


LovelySanta:
Similar to previous attempts at using the post build command, we need to make sure the directory exists to prevent crashes. Since this is not entirely the same, can you confirm this works for a clean build?

Hazelnut/premake5.lua
```diff
@@ -27,6 +27,11 @@ project "Hazelnut"
 	{
 		"Hazel"
 	}
+	
+	postbuildcommands
+	{
+		"{COPY} %{LibraryDir.VulkanSDK_DLL} %{wks.location}/bin/" .. outputdir .. "/%{prj.name}"
+	}
```
freeman40, line 34: Doesn't this only need to be done for DEBUG build?
For the destination,  does replacing  `%{wks.location}/bin/" .. outputdir .. "/%{prj.name}`    with   `%{cfg.targetdir}` work?  The latter is more succinct.

Dependencies.lua
```diff
@@ -20,6 +20,7 @@ LibraryDir = {}
 
 LibraryDir["VulkanSDK"] = "%{VULKAN_SDK}/Lib"
 LibraryDir["VulkanSDK_Debug"] = "%{wks.location}/Hazel/vendor/VulkanSDK/Lib"
+LibraryDir["VulkanSDK_DLL"] = "%{wks.location}/Hazel/vendor/VulkanSDK/Bin"
```
freeman40, line 23: Might be better named VulkanSDK_DebugDLL   (assuming the post build step only needs to be done for DEBUG build)

Hazelnut/premake5.lua
freeman40: Also, should Sandbox be done as well as Hazelnut?

freeman40:
> 
> 
> Similar to previous attempts at using the post build command, we need to make sure the directory exists to prevent crashes. Since this is not entirely the same, can you confirm this works for a clean build?

The premake {COPY} token expands to something that does check directory existence (at least, it does on Windows).  I have checked (on my fork) that a clean checkout works (where "works" means that after building Hazelnut, the vulkan debug libs are there in the Hazelnut bin folder)

Hazelnut/premake5.lua
CybernetHacker14: Right now, I haven't accounted what happens with the RELEASE build. So someone might wanna take a look at that.
I will test out `%{cfg.targetdir}`, and see what's happening.

Yes, this should also be there for Sandbox. But Sandbox hasn't been touched around since ages. So I am skeptical to go and make changes there for the time being.

CybernetHacker14: `%{cfg.targetdir}` works on my end, for 'Debug' build mode

freeman40: Nice.  I suggest to move the whole `postbuildcommands` bit so that it is part of the 	`filter "configurations:Debug"` bit.
Then test that it still works (for both Debug and Release builds)

Also, I think the same changes should be made for the Sandbox project (even though sandbox has other issues at present.. nevertheless it would be good to tidy up all this postbuildstuff in the one PR)

CybernetHacker14: So, moved the `postbuildcommands` to `filter "configurations:Debug"` and it works. Both Debug and Release builds are running properly. I checked the output directories, and Debug has those DLLs copied, whereas the Release doesn't have any DLL copied. So I believe Release builds will work fine without those libraries.

CybernetHacker14: @freeman40  You can check out the latest commit, and if it's up-to the expectations, please mark this message thread as resolved.

LovelySanta: I think it wouldn't hurt to put `"` around the targetdir to account for whitespaces in the location, similar to #449?

freeman40: I'm not sure you need " in this case.  449 is creating the post build step command line directly, whereas in this PR the post build command line is generated by the premake {COPY} token, which should be smart enough to deal with white spaces in the paths.  (I will test this to make sure)

CybernetHacker14: So is `"` needed? Should I add it to the targetdir?

Pikachuxxxx: I just experimented with it, it would work for me only if both the source and destination arguments of the copy are enclosed in ""

CybernetHacker14: So it should be `"{COPY} "%{LibraryDir.VulkanSDK_DebugDLL}" "%{cfg.targetdir}""` ?

freeman40: Yes, Pikachuxxxx is correct, the " are needed.  I also just learnt that premake {COPY} is deprecated in favour of {COPYDIR}  (or {COPYFILE} if you are copying files.
So, it should be:
"{COPYDIR} \"%{LibraryDir.VulkanSDK_DebugDLL}\" \"%{cfg.targetdir}\""

freeman40 (approved):

Hazelnut/premake5.lua
GrisWoldDiablo: I just tried COPYDIR and it does not work, premake does not covert it, but COPY works fine
"{COPY} \"%{LibraryDir.VulkanSDK_DebugDLL}\" \"%{cfg.targetdir}\""
this is what i did.

0xworks (approved):

Hazelnut/premake5.lua
0xworks: yes, COPYDIR is only available from in premake5.0.0-alpha16

GrisWoldDiablo: So the new python script to get premake downloads the latest version, but only if you dont have it already, there is no version check. 
but since COPY does the trick we should go with that one.

CybernetHacker14: I have merged the commits from upstream master to this branch. This puts this PR on the same page with the commit where PR #441 is merged. In PR #441, the python script downloads the premake5.exe binary, and the version is set to alpha16. So hopefully, the COPYDIR and COPY issue gets resolved.

For those who have cloned the repository earlier, from time to time as new features/setup functionalities get added, they need to execute those batch scripts to download the necessary updated stuff manually, on their own local machine. We cannot control every aspect of setup - some of it needs to be handled by the end user itself. We can add utilities to automate some aspects of upgrading, but the execution of it can, and should be, done by the user itself.

Another alternate feature would be to check which version of premake is being used in the premake scripts, and then add conditional statements, but that would clutter up stuff unnecessarily.

LovelySanta: I also tested this, copydir works perfect, the new exe gets downloaded, this is no issue anymore, ready to be merged 👍

PR #441, opened 2021-05-12
Add python scripts to install Premake

3x3y3z3t:
This PR adds a python script to check and automatically download Premake5.exe if the file is not presented.

#### PR impact

#### Proposed fix
Another script, `Premake.py`, is called during setup. It will check if `Premake5.exe` is presented in the specified `vendor/premake/bin` directory. If the file is not presented, it will attempt to download from the specified URL and extract the file here.

#### Additional context
Currently, the script will attempt to download **v.5.0.0-alpha16** (which is the latest) **Windows** version of Premake.
Since I don't own any other platform, I can not test on them. Theoretically the script should works on other platform as well.
I have put a platform check so currently the script only try to download Premake if you are on Windows, so if you are running Hazel on other platform, you need to remove the checking first before running the script.


IshoAntar:
I don't think this is necessary for Linux (and Mac, maybe?). Since we have package managers, installing premake is as easy as:
```sh
# pacman -S premake
```
or
```sh
# apt-get install premake
```

Maybe throw an error if premake is not on the user's path? let them know that they should have it installed?

LovelySanta:
I am wondering why we need to install premake on windows? We just use premake without installation? The only think that could be checked is if the exe is present (as in the repository is cloned recursively)... Am I wrong?

IshoAntar:
Premake is not a submodule, so the exe should be preset even with a shallow clone. (unless if the user has deleted it, in that case it's their problem!).

LovelySanta:
Ah yes, my bad! I don't think this PR is a bad thing, as people dislike having binaries inside github... For that matter, I believe it would be nice to have such a script to run in the setup (since we have that now).

I won't merge it yet, as I do want to refactor some of the setup script as of now, since it is quite messy.

3x3y3z3t:
Sorry for late reply.


> Ah yes, my bad! I don't think this PR is a bad thing, as people dislike having binaries inside github... For that matter, I believe it would be nice to have such a script to run in the setup (since we have that now).

I also think that binaries should not be commit along with source code, that's why I made this PR, trying to throw the binaries problem to the user.

> I won't merge it yet, as I do want to refactor some of the setup script as of now, since it is quite messy.

In this PR I also follow the current setup script format. After they are refactored them I will also update the PR to match the new format.

> I don't think this is necessary for Linux (and Mac, maybe?). Since we have package managers, installing premake is as easy as:
> ...
> Maybe throw an error if premake is not on the user's path? let them know that they should have it installed?

With this approach, I'm trying to make installing Hazel one-click-only. But you're right, I forgot that Linux has package manager.
In the meantime, I will keep this code as it is. When Hazel officially supports Linux, I think this piece of code will change, or be removed.

AnantTiwari-Naman:
> With this approach, I'm trying to make installing Hazel one-click-only. But you're right, I forgot that Linux has package manager.

There are also third party package managers on windows (scoop, choco) and soon an official package manager (winget I think?)

AbdullahAmrSobh:
> I don't think this is necessary for Linux (and Mac, maybe?). Since we have package managers, installing premake is as easy as:
> 
> ```shell
> # pacman -S premake
> ```
> 
> or
> 
> ```shell
> # apt-get install premake
> ```
> 
> Maybe throw an error if premake is not on the user's path? let them know that they should have it installed?

iirc, premake5 can't be installed via package managers

IshoAntar:
> iirc, premake5 can't be installed via package managers

AbdullahAmrSobh:
> > iirc, premake5 can't be installed via package managers
> 
> 

If you go to [https://premake.github.io/download](url) only two options are available, either build from source. or download compressed prebuilt binaries
Screenshot_1

IshoAntar:
Because the distribution packages are probably maintained by the community and not premake's maintainers.

And, i checked. `premake5` is *not* available on ubuntu (`premake4` is though `sudo apt-get install premake4`), that's a bummer!

`premake5` *is* available on Arch though.
Arch wins...Again! ( jk :) )

LovelySanta:
> > I won't merge it yet, as I do want to refactor some of the setup script as of now, since it is quite messy.
> 
> In this PR I also follow the current setup script format. After they are refactored them I will also update the PR to match the new format.

#438 has been merge. Please create a `SetupPremake.py` file similar to `SetupVulkan.py` to download the premake binaries to the correct location.

3x3y3z3t:
> #438 has been merge. Please create a `SetupPremake.py` file similar to `SetupVulkan.py` to download the premake binaries to the correct location. For now I would only create a setup for windows as discussed above.

Sorry for late update.
I have just update SetupPremake.py to match other scripts.
I've tested it for a while and it seems good on my end, but please let me know if there is any problems.

LovelySanta:
This PR seems to have alot of file changes, I think something went wrong, as I see that this PR still tries to add [`scripts/Premake.py`](https://github.com/TheCherno/Hazel/pull/441/files#diff-b15d5bb8081eb8b79a545dc0af03f7d73a694cf12cdbeb89374844b21bda71a8) AND [`scripts/SetupPremake.py`](https://github.com/TheCherno/Hazel/pull/441/files#diff-01431d873c0f1c5d1696780a06fc7c0694502b414d082a158068d9e3e02a61a7).. seems like the old files are not deleted.

Another thought I had, why does the script try to change directories twice?
https://github.com/TheCherno/Hazel/blob/9aaa6fe6da2bc2d88fdba6ae5b54087df5229eca/scripts/Setup.py#L11-L13

https://github.com/TheCherno/Hazel/blob/9aaa6fe6da2bc2d88fdba6ae5b54087df5229eca/scripts/Setup.py#L15-L19

Shouldn't it look something like
```python
from SetupPremake import PremakeConfiguration as PremakeRequirements
from SetupVulkan import VulkanConfiguration as VulkanRequirements

os.chdir('./../') # Change from devtools/scripts directory to root

PremakeRequirements.Validate()
VulkanRequirements.Validate()
```

Looking at the flow of the script, I would think it has to be first installing Vulkan, then premake, and finally run the premake script?
```python
from SetupVulkan import VulkanConfiguration as VulkanRequirements
from SetupPremake import PremakeConfiguration as PremakeRequirements

os.chdir('./../') # Change from devtools/scripts directory to root

VulkanRequirements.Validate()
PremakeRequirements.Validate()

# Run the premake script
```

3x3y3z3t:
> This PR seems to have alot of file changes, I think something went wrong,
When I fetch from upstream (TheCherno/Hazel) then merge into this branch, it also applied new commits from TheCherno/Hazel as well. I know there was something wrong in merging process, but don't know what was wrong, and I also don't like those commits being stacked here.
Is there any way to fix it?

> I see that this PR still tries to add [`scripts/Premake.py`](https://github.com/TheCherno/Hazel/pull/441/files#diff-b15d5bb8081eb8b79a545dc0af03f7d73a694cf12cdbeb89374844b21bda71a8) AND [`scripts/SetupPremake.py`](https://github.com/TheCherno/Hazel/pull/441/files#diff-01431d873c0f1c5d1696780a06fc7c0694502b414d082a158068d9e3e02a61a7).. seems like the old files are not deleted.
Thanks for pointing out. For some reason I forgot to remove the old script.

> Another thought I had, why does the script try to change directories twice?
In a rush moment I've copy-pasted the two lines (for consistency in naming style) and forgot to remove one.

> Looking at the flow of the script, I would think it has to be first installing Vulkan, then premake, and finally run the premake script?
I think installing Premake first or Vulkan first (the installing order) has no difference... Or, Premake is for the build system, Vulkan is an engine dependency, so Premake should be installed first, right (Python (tool) -> Premake (tool) -> Vulkan (library))

LovelySanta:
> > This PR seems to have alot of file changes, I think something went wrong,
> 
> When I fetch from upstream (TheCherno/Hazel) then merge into this branch, it also applied new commits from TheCherno/Hazel as well. I know there was something wrong in merging process, but don't know what was wrong, and I also don't like those commits being stacked here.
> Is there any way to fix it?

The simples thing would be to copy paste the script files to a separate branch corresponding with a new PR... but that does not matter... just make sure the old files are deleted. I know which changes are already in master and which aren't...

> > Looking at the flow of the script, I would think it has to be first installing Vulkan, then premake, and finally run the premake script?
> 
> I think installing Premake first or Vulkan first (the installing order) has no difference... Or, Premake is for the build system, Vulkan is an engine dependency, so Premake should be installed first, right (Python (tool) -> Premake (tool) -> Vulkan (library))

That seems resonable.

3x3y3z3t:
I've deleted the old `Premake.py` file, and remove redundant directory selection line in `Setup.py`.
I still can do nothing about the number of file changed though, the only files that I have touched are `Setup.py` and `SetupPremake.py`.

scripts/SetupPremake.py
```diff
@@ -0,0 +1,45 @@
...
+    @classmethod
+    def CheckIfPremakeInstalled(cls):
+        premakeExe = Path(f"{cls.premakeDirectory}/premake5.exe");
+        if (not premakeExe.exists()):
+            permissionGranted = False
+            while not permissionGranted:
+                reply = str(input("Premake not found. Would you like to install Preamake {0:s}? [Y/N]: ".format(cls.premakeVersion))).lower().strip()[:1]
+                if reply == 'n':
+                    return False
+                permissionGranted = (reply == 'y')
```
LovelySanta, line 32: If we want to go 100% for consistency, the permission should be asked in `InstallPremake` while `CheckIfPremakeInstalled` should determine and inform if premake is installed or not.

scripts/SetupPremake.py
```diff
@@ -0,0 +1,45 @@
...
+
+        return True
+
+    @classmethod
+    def InstallPremake(cls):
+        print("    Downloading Premake...")
+        with urlopen(cls.premakeZipUrls) as zipresp:
+            with ZipFile(BytesIO(zipresp.read())) as zfile:
+                zfile.extractall(cls.premakeDirectory)
+        print(f"    Premake {cls.premakeVersion} has been downloaded to '{cls.premakeDirectory}'")
```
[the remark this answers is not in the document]
LovelySanta: Note, when using this suggestion some unused imports can be removed.

LovelySanta (changes requested):
I cloned the repo for thoroughly testing this PR, here I have some observations...

3x3y3z3t:
Thank you a lot for your suggestion.
I also put a check in the (main) script Setup.py so it won't attempt to call `Win-GenProject` script if Premake is not detected.

scripts/SetupPremake.py
```diff
@@ -0,0 +1,51 @@
+import sys
+import os
+from pathlib import Path
+
+import Utils
+
+class PremakeConfiguration:
+    premakeVersion = "5.0.0-alpha16"
+    premakeZipUrls = f"https://github.com/premake/premake-core/releases/download/v{premakeVersion}/premake-{premakeVersion}-windows.zip"
+    premakeLicenseUrl = "https://github.com/premake/premake-core/blob/master/LICENSE.txt"
```
LovelySanta, line 10: If you look at the current state of the [`license.txt`](https://github.com/TheCherno/Hazel/blob/b29332e58962fb75e2e61f16e301f124cc88fac4/vendor/premake/bin/LICENSE.txt) file, you'll see that it downloaded the actual github page rather than the txt file. This is easily resolved by changing the `premakeLicenseUrl` to the raw txt file:
```suggestion
    premakeLicenseUrl = "https://raw.githubusercontent.com/premake/premake-core/master/LICENSE.txt"
```

vendor/premake/bin/LICENSE.txt
```diff
@@ -1,27 +1,1554 @@
...
+  </div>
+</template>
+
+
+
+  
+
+  </body>
+</html>
+
```
LovelySanta, line 1554: I think what you did here (and similar for premake.exe), you did run the script and then committed the changes. Rather than doing that, you should delete the bin directory completely, that way these files will be **deleted** rather than **updated** to the latest version. The premake script is there to download these files using python rather than 'storing' a version on github. Please **remove** these files from this branch.

LovelySanta (changes requested):
Something is still not right...

> Btw how can I remove the "Changes request" message? Is it safe to ignore it?

Only I (or any other maintainer) can remove that by approving this PR.

scripts/SetupPremake.py
3x3y3z3t: I'm sorry, I didn't check the content of the license file, I just tested and see if it actually download something :(

vendor/premake/bin/LICENSE.txt
3x3y3z3t: Actually I did delete the bin directory, but then I went back, run the script (test the script) and forgot to remove the directory again before committing.
This time I made sure the directory was deleted.

scripts/Utils.py
```diff
@@ -1,13 +1,56 @@
...
+        raise ValueError(f"Failed to download {filepath}")
+    if not(type(url) is str):
+        raise TypeError("Argument 'url' must be of type list or string")
+
     with open(filepath, 'wb') as f:
-        ua = UserAgent()
-        headers = {'User-Agent': ua.chrome}
+        headers = {'User-Agent': "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.97 Safari/537.36"}
         response = requests.get(url, headers=headers, stream=True)
         total = response.headers.get('content-length')
```
LovelySanta, line 56: Due to the small file size of license.txt I ran into a small error, which I already fixed in the unzip function, but now also rises in the download function:

1. Due to the small file size, there is an issue as downloaded > total (due to rounding), which can even become 0 in some cases
2. Because of a decent download speed and the [inaccurate timer inside python](https://stackoverflow.com/questions/1938048/high-precision-clock-in-python/21301124#:~:text=time()%20function%20provides%20sub,problems%20due%20to%20process%20interrupts.), I encounter a download speed of 0, which also results in a `ZeroDivisionError` when calculating the avgKBPerSecond.

In order to resolve these issues, could you replace lines 66-69
https://github.com/TheCherno/Hazel/blob/531b9af00c1e5aef037f8475e409790bbcd8855f/scripts/Utils.py#L66-L69
to
```python
                try:
                    done = int(50*downloaded/total) if downloaded < total else 50
                    percentage = (downloaded / total) * 100 if downloaded < total else 100
                except ZeroDivisionError:
                    done = 50
                    percentage = 100
                elapsedTime = time.time() - startTime
                try:
                    avgKBPerSecond = (downloaded / 1024) / elapsedTime
                except ZeroDivisionError:
                    avgKBPerSecond = 0.0
```

scripts/SetupPremake.py
```diff
@@ -0,0 +1,51 @@
...
+
+        premakePath = f"{cls.premakeDirectory}/premake-{cls.premakeVersion}-windows.zip"
+        print("Downloading {0:s} to {1:s}".format(cls.premakeZipUrls, premakePath))
+        Utils.DownloadFile(cls.premakeZipUrls, premakePath)
+        print("Extracting", premakePath)
+        Utils.UnzipFile(premakePath, deleteZipFile=True)
+        print(f"Premake {cls.premakeVersion} has been downloaded to '{cls.premakeDirectory}'")
+
+        premakeLicensePath = f"{cls.premakeDirectory}/LICENSE.txt"
+        print("Downloading {0:s} to {1:s}".format(cls.premakeLicenseUrl, premakePath))
```
LovelySanta, line 47: Found another copy paste error (I swear this is the last one)
```suggestion
        print("Downloading {0:s} to {1:s}".format(cls.premakeLicenseUrl, premakeLicensePath))
```

LovelySanta (changes requested):
I may have been a bit too fast last time, but I think these are the last issues...

LovelySanta (approved):

PR #442, opened 2021-05-31
Mentioned an installation instruction in README.md

CybernetHacker14:
#### Describe the issue
The README has instructions about cloning the repository and it's submodules. But it doesn't mention how to create a Visual Studio solution file. Earlier this wasn't the problem, because a single `.bat` file existed which can be executed.

With latest additions of SPIR-V, as of right now, there are two `.bat` files which can be executed, and both will create a Visual Studio solution, out of which one will cause errors.

#### PR impact

#### Proposed fix
Add an instruction to README telling people coming to the repository for the first time to execute `Setup.bat` file.

#### Additional context
The additions with this commit may or may not prove to be redundant, during a future cleanup or refactor of the code.


AbdullahAmrSobh:
I think, when anyone reads "Win-GenProj" they would understand that this is what they should click on.

CybernetHacker14:
Umm, except for the fact that you shouldn't click on `Win-GenProj` anymore, cause `Setup.bat` is the new creation point as of now.
I have mentioned in the additional context, that if a cleanup happens, and the `Win-GenProj` file is removed, or altered, or something changes, then the README instruction will change accordingly as well.

This issue turned up, when a person cloned my engine, which follows the same progress and structure as Hazel, and executed the `Win-GenProj` file instead of `Setup` file.

LovelySanta:
Also mention which dll's have to be moved where (putting image for future reference)
image

CybernetHacker14:
Moving this to a new pull request, since this one's cluttered with a lot of unnecessary commits.

PR #444, opened 2021-06-12
Cleaned up the Renderer2D

JonasAlmaas:
#### Cleanup of the Renderer2D Class
The renderer2D class was a bit of a mess, so I took some time and cleaned it up. Now everything uses one super method for drawing. It functions exactly the same as the old code but looks a bit nicer.

Didn't make too much sense to use the template on this, hopefully that's not too much of an issue :)

### DIFF
--- a/Hazel/src/Hazel/Renderer/Renderer2D.h
+++ b/Hazel/src/Hazel/Renderer/Renderer2D.h
@@ -24,20 +24,25 @@ namespace Hazel {
 		static void Flush();
 
 		// Primitives
-		static void DrawQuad(const glm::vec2& position, const glm::vec2& size, const glm::vec4& color);
+		static void DrawQuad(const glm::mat4& transform, const Ref<Texture2D>& texture, const glm::vec4& tintColor, float tilingFactor = 1.0f);
+		static void DrawQuad(const glm::mat4& transform, const Ref<Texture2D>& texture, float tilingFactor = 1.0f);
+		static void DrawQuad(const glm::mat4& transform, const glm::vec4& color);
+
+		static void DrawQuad(const glm::vec3& position, const glm::vec2& size, const Ref<Texture2D>& texture, const glm::vec4& tintColor, float tilingFactor = 1.0f);
+		static void DrawQuad(const glm::vec2& position, const glm::vec2& size, const Ref<Texture2D>& texture, const glm::vec4& tintColor, float tilingFactor = 1.0f);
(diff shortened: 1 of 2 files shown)


Hazel/src/Hazel/Renderer/Renderer2D.h
```diff
@@ -24,20 +24,25 @@ namespace Hazel {
...
+		static void DrawQuad(const glm::vec3& position, const glm::vec2& size, const Ref<Texture2D>& texture, const glm::vec4& tintColor, float tilingFactor = 1.0f);
+		static void DrawQuad(const glm::vec2& position, const glm::vec2& size, const Ref<Texture2D>& texture, const glm::vec4& tintColor, float tilingFactor = 1.0f);
+		static void DrawQuad(const glm::vec3& position, const glm::vec2& size, const Ref<Texture2D>& texture, float tilingFactor = 1.0f);
+		static void DrawQuad(const glm::vec2& position, const glm::vec2& size, const Ref<Texture2D>& texture, float tilingFactor = 1.0f);
+		static void DrawQuad(const glm::vec3& position, const glm::vec2& size, const glm::vec4& color);
+		static void DrawQuad(const glm::vec2& position, const glm::vec2& size, const glm::vec4& color);
 
-		static void DrawRotatedQuad(const glm::vec2& position, const glm::vec2& size, float rotation, const glm::vec4& color);
+		static void DrawRotatedQuad(const glm::vec3& position, const glm::vec2& size, float rotation, const Ref<Texture2D>& texture, const glm::vec4& tintColor, float tilingFactor = 1.0f);
```
freeman40, line 38: I know you put in comments elsewhere that rotation is in radians.  However, it is such a common mistake to pass degrees instead of radians that I like to call the argument something like `rotationRadians` so that things like auto-complete or "intellisense" in the user's IDE will remind them its radians without having to go and look at code comments.

freeman40 (approved):

Hazel/src/Hazel/Renderer/Renderer2D.h
IshoAntar: Maybe it would be worth defining custom objects (structs?)
for degrees and radians (with no implicit conversion operators between the two),
and using those? That way the distinction would be enforced by the compiler?

TheCherno:
Will come back to this once Vulkan is implemented.

JonasAlmaas:
> h

I could update it so it doesn't conflict with anything. If needed

PR #448, opened 2021-06-20
Small refactor of NativeScriptComponent (fixes #336)

### LINKED ISSUE #336: Memory leak (?)
### Possible memory leak (?)
Hi, I was analyzing the project for possible bugs and specifically memory leaks. I discovered a possible memory leak in the *Components.h* file. Is this intended? 

image

image

freeman40:
Per comments on #336 memory is leaked every time a NativeScriptComponent is instantiated.
This PR addresses that issue, and also makes the following changes:
- renames ScriptableEntity to NativeScript  (so that a NativeScriptComponent wraps a NativeScript, and also the thing called "ScriptableEntity" is not really an entity, but rather a component).
- Changes NativeScriptComponent::Instance into a Hazel::Scope  (this fixes the memory leak)
- Constructor for NativeScript now sets its entity, so that you do not have to remember to do this after calling InstantiateScript()
- Allows passing arbitrary parameters to NativeScriptComponent::InstantiateScript
- Remove not-needed OnCreate()  (put that stuff in your script object's constructor instead)
- Remove not-needed OnDestroy()  (put that stuff in your script object's destructor instead)
- Instantiation of a NativeScript is now one line instead of three.

#### PR impact
As described above

Issues this solves       | #336

#### Proposed fix
The memory leak is fixed because NativeScriptComponent::Instance is now a Hazel::Scope (aka a smart pointer), which automatically calls the destructor when NativeScriptComponent is destroyed.  The destructor is virtual, so the correct derived NativeScript destructor will end up being invoked..

#### Additional context
The PR includes example usage of the refactored NativeScriptComponent in EditorLayer::OnAttach().
I have also fully tested this (on Windows) using my own small game on my fork of Hazel. (see HazelDash on my github if interested)

### DIFF
--- a/Hazel/src/Hazel.h
+++ b/Hazel/src/Hazel.h
@@ -20,7 +20,7 @@
 
 #include "Hazel/Scene/Scene.h"
 #include "Hazel/Scene/Entity.h"
-#include "Hazel/Scene/ScriptableEntity.h"
+#include "Hazel/Scene/NativeScript.h"
 #include "Hazel/Scene/Components.h"
 
 // ---Renderer------------------------
--- a/Hazel/src/Hazel/Scene/Components.h
+++ b/Hazel/src/Hazel/Scene/Components.h
@@ -7,7 +7,7 @@
 #include <glm/gtx/quaternion.hpp>
 
 #include "SceneCamera.h"
-#include "ScriptableEntity.h"
+#include "NativeScript.h"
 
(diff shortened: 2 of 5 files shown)


drsnuggles8:
Don't think this PR is working anymore now with the physics changes @0xworks

TheCherno:
Will revisit this once C# scripting has been added.

PR #451, opened 2021-06-25
Added Binary for setting up Vulkan on windows

Jaysmito101:
#### Added Vulkan Setup

Along with the Python setup scripts added a C# Vulkan setup which both downloads and extracts the vulkan libs and also sets up the Vulkan SDK.

## Why?

So that if someone does not have Python installed may be able to setup Vulkan Easily.


LovelySanta:
I'll close this for now as we'll stay with python alone for now. There are python packages out there to convert the python script to an executable which can then work as an 'installer' down the road, which then doesn't even require python to be installed.

PR #452, opened 2021-06-25
Added Estimated Time Remaining to Download and Unzip in Utils.py

Jaysmito101:
This adds a feature.

Now when running setup the Estimated time remaining is shown (Both for downloading and for Unzipping).
Also added a `FormatTime(seconds)`  function which may be used for other parts of setup.

#### Additional context
Tested on Windows 10

### DIFF
--- a/scripts/Utils.py
+++ b/scripts/Utils.py
@@ -6,6 +6,8 @@ import requests
 import time
 import urllib
 
+import datetime
+
 from zipfile import ZipFile
 
 def GetSystemEnvironmentVariable(name):
@@ -41,7 +43,7 @@ def DownloadFile(url, filepath):
                 print(f"HTTP Error  encountered: {e.code}. Proceeding with backup...\n\n")
                 os.remove(filepath)
                 pass
-            except:
+            except Exception:
                 print(f"Something went wrong. Proceeding with backup...\n\n")
                 os.remove(filepath)
                 pass
@@ -75,12 +77,13 @@ def DownloadFile(url, filepath):
                     avgKBPerSecond = (downloaded / 1024) / elapsedTime
                 except ZeroDivisionError:
                     avgKBPerSecond = 0.0
-
+                estimatedTime = ( (total - downloaded) / (avgKBPerSecond*1024) )
+                estimatedTimeString = FormatTime(estimatedTime)
                 avgSpeedString = '{:.2f} KB/s'.format(avgKBPerSecond)
                 if (avgKBPerSecond > 1024):
                     avgMBPerSecond = avgKBPerSecond / 1024
                     avgSpeedString = '{:.2f} MB/s'.format(avgMBPerSecond)
-                sys.stdout.write('\r[{}{}] {:.2f}% ({})     '.format('█' * done, '.' * (50-done), percentage, avgSpeedString))
+                sys.stdout.write('\r[{}{}] {:.2f}% ({}) (Estimated Time: {})    '.format('█' * done, '.' * (50-done), percentage, avgSpeedString, estimatedTimeString))
                 sys.stdout.flush()
     sys.stdout.write('\n')
 
@@ -110,18 +113,40 @@ def UnzipFile(filepath, deleteZipFile=True):
             except ZeroDivisionError:
                 done = 50
                 percentage = 100
-            elapsedTime = time.time() - startTime
+            elapsedTime = time.time() - startTime            
+            estimatedTime = 0
             try:
                 avgKBPerSecond = (extractedContentSize / 1024) / elapsedTime
+                estimatedTime = ( (zipFileContentSize - extractedContentSize) / (avgKBPerSecond*1024) )
             except ZeroDivisionError:
                 avgKBPerSecond = 0.0
+                estimatedTime = 0
+            estimatedTimeString = FormatTime(estimatedTime)
             avgSpeedString = '{:.2f} KB/s'.format(avgKBPerSecond)
             if (avgKBPerSecond > 1024):
                 avgMBPerSecond = avgKBPerSecond / 1024
                 avgSpeedString = '{:.2f} MB/s'.format(avgMBPerSecond)
-            sys.stdout.write('\r[{}{}] {:.2f}% ({})     '.format('█' * done, '.' * (50-done), percentage, avgSpeedString))
+            sys.stdout.write('\r[{}{}] {:.2f}% ({}) (Estimated Time: {})     '.format('█' * done, '.' * (50-done), percentage, avgSpeedString, estimatedTimeString))
             sys.stdout.flush()
     sys.stdout.write('\n')
 
     if deleteZipFile:
         os.remove(zipFilePath) # delete zip file
+
+def FormatTime(seconds):
+    h = int(seconds // 3600)
+    m = int(seconds % 3600 // 60)
+    s = int(seconds % 3600 % 60)
+    formatedTime = ""
+    if h == 0 and m == 0 and s == 0:
+        return "0s"
+    if seconds > 86400:
+        return "> 1 day"
+        # OR return "More than a day!"
+    if h > 0:
+        formatedTime += "{:02d}h ".format(h)
+    if m > 0 or h > 0:
+        formatedTime += "{:02d}m ".format(m)
+    if s > 0 or m > 0 or h > 0:
+        formatedTime += "{:02d}s ".format(s)
+    return formatedTime
--- a/scripts/Utils.py
+++ b/scripts/Utils.py
@@ -43,7 +43,7 @@ def DownloadFile(url, filepath):
                 print(f"HTTP Error  encountered: {e.code}. Proceeding with backup...\n\n")
                 os.remove(filepath)
                 pass
-            except Exception:
+            except:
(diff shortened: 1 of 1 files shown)


scripts/Utils.py
```diff
@@ -110,18 +113,40 @@ def UnzipFile(filepath, deleteZipFile=True):
...
+            elapsedTime = time.time() - startTime            
+            estimatedTime = 0
             try:
                 avgKBPerSecond = (extractedContentSize / 1024) / elapsedTime
+                estimatedTime = ( (zipFileContentSize - extractedContentSize) / (avgKBPerSecond*1024) )
             except ZeroDivisionError:
                 avgKBPerSecond = 0.0
+                estimatedTime = 0
+            estimatedTimeString = FormatTime(estimatedTime)
             avgSpeedString = '{:.2f} KB/s'.format(avgKBPerSecond)
```
LovelySanta, line 125: Same comment as on line 82

For flow of the program, quite a small detail, but I would suggest putting the speed first and then the time, as we first calculated the speed as well...
```suggestion
            avgSpeedString = '{:.2f} KB/s'.format(avgKBPerSecond)
            estimatedTimeString = FormatTime(estimatedTime)
```

scripts/Utils.py
```diff
@@ -75,12 +77,13 @@ def DownloadFile(url, filepath):
                     avgKBPerSecond = (downloaded / 1024) / elapsedTime
                 except ZeroDivisionError:
                     avgKBPerSecond = 0.0
-
+                estimatedTime = ( (total - downloaded) / (avgKBPerSecond*1024) )
+                estimatedTimeString = FormatTime(estimatedTime)
                 avgSpeedString = '{:.2f} KB/s'.format(avgKBPerSecond)
```
LovelySanta, line 82: For flow of the program, quite a small detail, but I would suggest putting the speed first and then the time, as we first calculated the speed as well...
```suggestion
                avgSpeedString = '{:.2f} KB/s'.format(avgKBPerSecond)
                estimatedTimeString = FormatTime(estimatedTime)
```

LovelySanta (changes requested):
A few of these comments point out some potential errors (from experience). Others are just my thoughts. I did only read over the code at this point, I did not run the code yet.

A final general suggestion/thought: The code actually representing the download or unzip speed is basically a bar that progresses with % next to it, followed by speed and time. This is now in 2 places, and other than variable names there is minimal difference between the two. Wouldn't it be useful to generalize it to a separate helper function and providing it with the correct arguments, while handling all the edge cases of both uses?

Jaysmito101:
@LovelySanta I tried to change as per you said can you please check?

scripts/Utils.py
```diff
@@ -110,18 +115,38 @@ def UnzipFile(filepath, deleteZipFile=True):
...
                 percentage = 100
-            elapsedTime = time.time() - startTime
+            elapsedTime = time.time() - startTime            
+            estimatedTime = 0
             try:
                 avgKBPerSecond = (extractedContentSize / 1024) / elapsedTime
+                estimatedTime = (zipFileContentSize - extractedContentSize) / (avgKBPerSecond * 1024)
             except ZeroDivisionError:
                 avgKBPerSecond = 0.0
+                estimatedTime = 0.0
```
LovelySanta, line 125:
```suggestion
                estimatedTime = -1.0 # representation of infinite
```

LovelySanta (commented):
You forgot a small one.

And I don't think there is a need for the `import datetime`? It's not used afaik? 

Also, did you give some thought yet to my comment in https://github.com/TheCherno/Hazel/pull/452#pullrequestreview-693098845 ?

> A final general suggestion/thought: The code actually representing the download or unzip speed is basically a bar that progresses with % next to it, followed by speed and time. This is now in 2 places, and other than variable names there is minimal difference between the two. Wouldn't it be useful to generalize it to a separate helper function and providing it with the correct arguments, while handling all the edge cases of both uses?

Jaysmito101:
@LovelySanta At first i thought making a separate function will not be that important as it is just 2 lines but then Whats the problem? so i added a  `CalculateEstimatedTime` function. I tried to keep it reusable in other cases also.

LovelySanta:
> @LovelySanta At first i thought making a separate function will not be that important as it is just 2 lines but then Whats the problem? so i added a `CalculateEstimatedTime` function. I tried to keep it reusable in other cases also.

Nonono, that's not what I meant, of course do not make a function for that simple calculation. I was more thinking a single function `ProgressBarUpdate` that does this:
https://github.com/TheCherno/Hazel/blob/9818942a89d8de7ecf40bee364c4d2e8c3da0f83/scripts/Utils.py#L68-L83

Such that the 'loop' simplifies to
```python
            for data in response.iter_content(chunk_size=max(int(total/1000), 1024*1024)):
                downloaded += len(data)
                f.write(data)
                ProgressBarUpdate(downloaded, total, startTime)
```

Please revert the creation of the `CalculateEstimatedTime` function as we both agree that is not what we wanted. I should explain myself better in the future...

Jaysmito101:
@LovelySanta i reverted it. and sorry that i misunderstood what you meant. And as for what you really meant (ProgressBar) i guess it will be better (i guess) to have a small ProgressBar class to make things really simpler(abstracting it). But i think that will be a different PR as this PR is just about adding Estimated time feature. Abstracting the Progress bar is bit different. (I may be wrong!)

Jaysmito101:
@LovelySanta Could you please review the changes you requested?

LovelySanta (changes requested):
Sorry that it took so long to review. I did ran the script a few days ago and I got this as a final output:
image
So when it is finished, it should say it is done, which it is not. I expect this is some `ZeroDevisionError` putting the time to infinity... However I cannot seem to pinpoint where it is coming from entirely..

I think the cause is this (but i'm not entirely sure):
https://github.com/TheCherno/Hazel/blob/7a5098bf40924c8850bdc278f988cfb796f5007e/scripts/Utils.py#L120-L125

When it unzips extremely fast, the `elapsedTime` == 0, causing the `ZeroDevisionError`.
In order to resolve this issue, I think the solution is to give conditional values to `elapsedTime` in the error handling, depending on the content downloaded:
```py
estimatedTime = 0 if zipFileContentSize == extractedContentSize else -1
```

A similar (more rare) case can occur when downloading as well:
https://github.com/TheCherno/Hazel/blob/7a5098bf40924c8850bdc278f988cfb796f5007e/scripts/Utils.py#L77-L82
Which can be handled equivalently:
```py
estimatedTime = 0 if total == downloaded else -1
```

Other than that, it seems to work as intended... If this is the cause, this could be a solution...

Jaysmito101:
@LovelySanta with a bit of testing I found that it in this case where the download or unzipping is happening very fast the time is going negative.  (I guess due to this line:)
    
    estimatedTime = -1.0 # representation of infinite

So, I think a potential solution could be:

    ...
    if seconds < 0:
        return "Finished!"
    if seconds < 0 or seconds > 86400:
        return "> 1 day" # OR return "More than a day!"
    if seconds == 0:
        return "Finished!" # OR return "0s"
    ...

In the FormatTime function.

And for my case its solves the issue too. 

```
Premake 5.0.0-alpha16 has been downloaded to './vendor/premake/bin'
Downloading https://raw.githubusercontent.com/premake/premake-core/master/LICENSE.txt to ./vendor/premake/bin/LICENSE.txt
[██████████████████████████████████████████████████] 100.00% (0.00 KB/s) (Estimated Time: Finished!)
Premake License file has been downloaded to './vendor/premake/bin'
Correct Premake located at C:\Users\Jaysmito Mukherjee\Desktop\a\Hazel\vendor\premake\bin
```

But i am not sure whether it will cause any other issue or not so please have a look.

LovelySanta:
> @LovelySanta with a bit of testing I found that it in this case where the download or unzipping is happening very fast the time is going negative. (I guess due to this line:)
> 
> ```
> estimatedTime = -1.0 # representation of infinite
> ```

Indeed, that's because of the time == 0, which results in the error.

> So, I think a potential solution could be:
> 
> ```
> ...
> if seconds < 0:
>     return "Finished!"
> if seconds < 0 or seconds > 86400:
>     return "> 1 day" # OR return "More than a day!"
> if seconds == 0:
>     return "Finished!" # OR return "0s"
> ...
> ```
> In the FormatTime function.

Not realy a solution, as the second if statement will never have seconds < 0 anymore...

> But i am not sure whether it will cause any other issue or not so please have a look.

I think the better solution is to check how much has been downloaded yet to make a distinction between the two cases:
1. Downloading hasn't started, causing `avgKBPerSecond` to be 0
2. Download has finished extremely fast, causing the `elapsedTime` to be 0

The error handling has to differentiate between those cases. The `elapsedtime` has to be equal to:
1. `avgKBPerSecond` being 0 means infinite time, thus `elapsedtime` must be -1
2. Downloading has finished, the `FormatTime` function requests it to be 0

Thus, inside the error handling, we have to change
```py
estimatedTime = -1.0 # representation of infinite
```
to (in case of the zip)
```py
estimatedTime = 0 if total == downloaded else -1
```
where `estimatedTime` will be 0 if the download has finished, otherwise it will be -1.
I mentioned a similar fix for the zip function.

I'm confident this will fix the issue at hand.

PR #453, opened 2021-07-05
Debug Libs to be unzipped in vulkanSDK location

GrisWoldDiablo:
#### Describe the issue
When you install the VulkanSDK you can install it where ever you want.
Meaning it might not be here ./Hazel/vendor/VulkanSDK and when the DebugLibs are unzipped they are placed in there regardless.
And when executing the project it will says that you do not have the shaderc_sharedd.dll, because it is looking in your actual vulkan sdk location.

#### PR impact
Issues this solves       | Closes #437

#### Proposed fix 
With the use of os.environ.get("VULKAN_SDK") we can garantee we will retrieve its location and when we download the DebugLibs we will place them at the proper location for them to be unzipped.

### DIFF
--- a/scripts/SetupVulkan.py
+++ b/scripts/SetupVulkan.py
@@ -11,6 +11,7 @@ from urllib.request import urlopen
 class VulkanConfiguration:
     requiredVulkanVersion = "1.2.170.0"
     vulkanDirectory = "./Hazel/vendor/VulkanSDK"
+	vulkanSDK = os.environ.get("VULKAN_SDK")
 
     @classmethod
     def Validate(cls):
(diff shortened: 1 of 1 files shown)


GrisWoldDiablo:
well i just noticed that on the Dependencies.lua you have to change
LibraryDir["VulkanSDK_Debug"] = "%{wks.location}/Hazel/vendor/VulkanSDK/Lib"
to
LibraryDir["VulkanSDK_Debug"] = "%{VULKAN_SDK}/Lib"

Since those libs below are not placed it /Hazel/vendor/VulkanSDK/Lib anymore.
its not clean but it works, I think it be best we just dont use VulkanSDK_Debug all together.

LovelySanta:
> When you install the VulkanSDK you can install it where ever you want.
> Meaning it might not be here ./Hazel/vendor/VulkanSDK and when the DebugLibs are unzipped they are placed in there regardless.

This is how the original script from cherno is intended to work. You install the SDK onto your system (not only for Hazel, but for general use) and for the development you download the DebugLibs to the project.

> And when executing the project it will says that you do not have the shaderc_sharedd.dll, because it is looking in your actual vulkan sdk location.

That is not entirely correct. When executing the project, it will look in the bin directory (as that is the working directory) for the dll's. To solve this, there should be a post build command to copy the dll's to the bin directory, which is implemented in #437.

I'll close this as this is not entirely a correct fix to the issue at hand.

PR #454, opened 2021-07-07
Fixes so a fresh checkout of master compiles, Hazelnut and Sandbox both work

0xworks:
1) After the "new shader system" commit,  Hazel does not compile cleanly after a fresh checkout from master.
2) Even supposing you can get it to compile, the Sandbox app does not render anything.

This PR fixes the second issue above.
The first requires that https://github.com/TheCherno/yaml-cpp/pull/1 is merged into Cherno's yaml-cpp fork, and then the submodules of this project updated accordingly.

Also note that until #437 is merged, then you still need to manually copying the Vulkan SDK debug dlls to the client application bin folder after building the solution.

Issues this solves       | Refer comments above

This fix basically brings Sandbox up-to-date with changes made in Hazel.  Note that several "assets" are duplicated between Hazelnut and Sandbox.  This is because Hazel currently loads the assets using a path relative to the client application's folder.  I have not refactored this as I do not want to make any assumptions about how Cherno wants to eventually handle assets.

#### Additional context
Tested (on Windows only) by doing a clean checkout, running the setup.bat script, building solution (which now does compile), then running both Hazelnut and Sandbox (which now both run (notwithstanding comment about copying dlls post build) and produce expected output)


CybernetHacker14:
Would the PR for `yaml-cpp` get merged though? When was the last time any one those forked submodules got updated?

0xworks:
The yaml PR is against Cherno's fork of it only,  and the change is only to his premake script.   It's nothing to do with the "real" yaml code, and it doesn't need to be merged upstream.

PR #455, opened 2021-07-09
Set staticruntime to off in Glad project

LalondeBrenton:
Set staticruntime off.  No issues yet with it being on but the rest of the projects are off.


3x3y3z3t:
I suggest change the title to something like "Set staticruntime to off in Glad project" for more meaningful.

Also, I found that `staticruntime on` also cause problem in yaml-cpp project. [Here](https://discord.com/channels/349857236656324608/496257118098292736/879743533337182279) is what happened (Discord chat).

Issue #465, opened 2021-07-31
readpixel less than -1


realjf:
Describe the bug
![image](https://user-images.githubusercontent.com/20607348/127740052-c28cec74-d806-4601-b99d-2018c43fc818.png)

![image](https://user-images.githubusercontent.com/20607348/127740209-6b945c87-9b3f-46f7-b21e-4ee2386b316b.png)

readpixel is less than -1

Operating system: (please complete the following information)

- OS: win-64
- Version: Windows 10

LovelySanta:
Can you reproduce this issue?

realjf:
I saw the solution [#419](https://github.com/TheCherno/Hazel/pull/419) , and found the reason, thank you

PR #419, opened 2021-02-11
Fix wrong viewport bounds when tab is visible

lfilipozzi:
#### Describe the issue (if no issue has been made)
The viewport bound does not account for the tab bar correctly. As a result, when the tab bar is visible, the mouse picking is vertically offset. Also, when the cursor is right below the tab bar, `Framebuffer::ReadPixel()` returns an undefined integer.

ezgif com-gif-maker

#### PR impact _(Make sure to add [closing keywords](https://help.github.com/en/articles/closing-issues-using-keywords))_
List of related issues/PRs this will solve:

 Impact                  | Issue/PR
------------------------ | ------
Issues this solves       | Resolves #401
Other PRs this solves    | Closes #402

#### Proposed fix 
Redefine m_ViewportBounds in EditorLayer.cpp by using `ImGui::GetWindowContentRegionMin()` and `ImGui::GetWindowContentRegionMax()` instead of `ImGui::GetCursorPos()`.

#### Additional context
The fix works with a visible and a hidden viewport tab bar.

Using `ImGui::GetForegroundDrawList()->AddRect(minBound, maxBound, IM_COL32(255, 255, 0, 255));` in EditorLayer.cpp to highlight the viewport edges also shows the viewport vertical bounds is not correct when the tab bar is present.

Before fix (note the bottom edge):
bug

After fix:
fix

___
Edit by @LovelySanta: Added PR impact and moved parts of description to additional context

PR #466, opened 2021-07-31
fix readpixel less than -1(fixs #465)

realjf:
fix readpixel less than -1


realjf:
I have seen the solution #419 , thank you

PR #478, opened 2021-09-09
Update Setup.py

ThomasLihoreau:
Adding the pip installation requirement

#### Description
Hi Cherno, I have fixed the build (that use Python) on my repository by added the installation of "pip" because I was so confused that Hazel doesn't work on my PC. 
Now everything is working perfectly XD

Remain to add that to the real Hazel's repository 🤣

#### PR impact

### DIFF
--- a/scripts/Setup.py
+++ b/scripts/Setup.py
@@ -1,9 +1,14 @@
 
 import os
+import sys
 import subprocess
 import platform
 
 from SetupPython import PythonConfiguration as PythonRequirements
+
+# Make sure the pip package manager is installed (python -m ensurepip)
+
+subprocess.check_call([sys.executable, "-m", "ensurepip"])
 
 # Make sure everything we need for the setup is installed
 PythonRequirements.Validate()


LovelySanta:
I am against this solution, but I do acknowledge the problem.

First of all, this code should reside inside [SetupPython](https://github.com/TheCherno/Hazel/blob/master/scripts/SetupPython.py) and should be prompted to the user. Not everyone installs python on their system using pip ([Anaconda](https://www.anaconda.com/products/individual-d) is equally accepted on Windows).

My proposition is as follows: check if pip is installed, and ask the user permission to install pip if missing. This follows the same approach as how the required packages are installed:
https://github.com/TheCherno/Hazel/blob/2b5ab711f95bfa49287737edb4bc73c85d58aec8/scripts/SetupPython.py#L32-L44

I suggest adding another class to the [SetupPython](https://github.com/TheCherno/Hazel/blob/master/scripts/SetupPython.py) file `PythonInstall` with two class methods `Validate(cls)` and `_InstallPip(cls)`. The validate function would call the install pip if pip is not present and after requesting permission from the user. The validate function can then be called in the main script as you ensure pip now.

PR #486, opened 2021-09-20
getting Hazel to compile under Linux

beaumanvienna:
#### Compile under Linux
Adding missing source files
Miscellaneous edits (capitalization in path names, etc.)


beaumanvienna:
@LovelySanta This is still a work in progress. I'll let you know when it's ready.

beaumanvienna:
@spector32 @AnantTiwari-Naman Would you like to test if it compiles for you under Linux? Hazelnut is still crashing when dragging files into the main canvas. That's because of this code [here](https://github.com/beaumanvienna/Hazel/blob/1972a31172542007b151be24c671917c1c909f32/Hazelnut/src/Panels/ContentBrowserPanel.cpp#L54). It has something to do with wide characters.
@LovelySanta : The Windows side is building ok when VULKAN_SDK is properly installed. But it's still not quite there to be pulled in. If the Linux side compiles in general and only crashes here and there, is it ok for you that we take this in a new PR? I didn't mean to make this one here too complex...

LovelySanta:
> The Windows side is building ok when VULKAN_SDK is properly installed. But it's still not quite there to be pulled in.

I'm not 100% following what you mean by the last part. It should be possible to pull it in, even if it is just file by file from somewhere?

> If the Linux side compiles in general and only crashes here and there, is it ok for you that we take this in a new PR? I didn't mean to make this one here too complex...

I could discuss this with Cherno if we want a community branch for the linux part, or just want to keep it inside a PR... I am pretty confident that cherno doesn't want it merged into master, as he wants to restructure the renderer and make it more uniform...

beaumanvienna:
Now the PR is ready to be pulled in!

> I'm not 100% following what you mean by the last part. It should be possible to pull it in, even if it is just file by file from somewhere?

I meant Hazel only builds fine under Windows, when VulkanSDK is installed per your instructions (= `setup.bat`). But that was yesterday :-) Now you can also build it w/o the external Vulkan libs. 
Also, I added support for MSYS2/gcc. It compiles when VULKAN_SDK is not set. I tested my last commit by cloning from scratch for
* Windows MSVC: with and w/o VULAKN_SDK, debug and release --> ok
* Windows MSYS2/gcc: w/o   VULAKN_SDK, debug and release --> ok
* Linux gcc:w/o   VULAKN_SDK, debug and release --> ok

If you could discuss with Cherno, I'd appreciate it. Tell him to keep the "Windows window" as is. The Mac & Linux users will survive it. I love this term, it's hilarious! I would not change it. It is already in all the YT videos.

Thanks @LovelySanta!

beaumanvienna:
@LovelySanta, I'd like to ask, can we delete the file `vendor/premake/premake5.lua`? It is causing an issue with the GitHub Actions Windows Workflow. I tested `scripts\setup.bat` after deleting `vendor/premake/premake5.lua`. The script `Win-GenProjects.bat` would still create the Visual Studio project files as usual.  Thanks!

LovelySanta:
> @LovelySanta, I'd like to ask, can we delete the file `vendor/premake/premake5.lua`? It is causing an issue with the GitHub Actions Windows Workflow. I tested `scripts\setup.bat` after deleting `vendor/premake/premake5.lua`. The script `Win-GenProjects.bat` would still create the Visual Studio project files as usual. Thanks!

What seems to be the problem with [`vendor/premake/premake5.lua`](https://github.com/TheCherno/Hazel/blob/f627b9c90923382f735350cd3060892bbd4b1e75/vendor/premake/premake5.lua)? It seems such a trivial file... It comes in handy when you have to configure/add to the submodules...

beaumanvienna:
The build script is using premake5 from elsewhere, see [here](https://github.com/beaumanvienna/Hazel/blob/d4bdd4aa0c328271b4897ed21b51e26922fa225f/.github/workflows/windows-workflow.yml#L26). Obviously, it has to run fully automated, so w/o `setup.bat`. Then it stumbles over the missing premake5 executable in the [post-build step](https://github.com/beaumanvienna/Hazel/blob/c9e4187aedf71410e46dbafff4ffa62d9f9f32be/vendor/premake/premake5.lua#L15). Did you understand why we need a post-build step to rebuild the VS files?

beaumanvienna:
All good @LovelySanta , we can keep the file. I fixed it in my last commit, see above. Both GitHub workflows are contained in this PR and check out ok.

PR #487, opened 2021-09-24
Drag a texture image directly into the viewport

MajisTcd:
#### Drag a texture image directly into the viewport

Until now, to add a texture to a sprite renderer component, the image has to be dragged onto the "Texture" button in the scene hierarchy panel. Now, in addition, one can drag a texture image directly onto a component in the viewport.

### DIFF
--- a/Hazelnut/imgui.ini
+++ b/Hazelnut/imgui.ini
@@ -58,7 +58,7 @@ Collapsed=0
 DockId=0x0000000B,0
 
 [Docking][Data]
-DockSpace           ID=0x3BC79352 Window=0x4647B76E Pos=137,190 Size=1600,876 Split=X Selected=0x995B0CF8
+DockSpace           ID=0x3BC79352 Window=0x4647B76E Pos=73,126 Size=1600,876 Split=X Selected=0x995B0CF8
   DockNode          ID=0x00000008 Parent=0x3BC79352 SizeRef=1228,876 Split=X
     DockNode        ID=0x00000001 Parent=0x00000008 SizeRef=370,696 Split=Y Selected=0xC89E3217
       DockNode      ID=0x00000005 Parent=0x00000001 SizeRef=370,435 Selected=0x9A68760C
--- a/Hazelnut/imgui.ini
+++ b/Hazelnut/imgui.ini
@@ -4,7 +4,7 @@ Size=1600,900
 Collapsed=0
 
 [Window][Debug##Default]
-ViewportPos=1180,1184
+ViewportPos=1180,1011
 ViewportId=0x9F5F46A1
 Size=400,400
 Collapsed=0
@@ -58,7 +58,7 @@ Collapsed=0
 DockId=0x0000000B,0
 
 [Docking][Data]
-DockSpace           ID=0x3BC79352 Window=0x4647B76E Pos=294,341 Size=1600,876 Split=X Selected=0x995B0CF8
+DockSpace           ID=0x3BC79352 Window=0x4647B76E Pos=137,190 Size=1600,876 Split=X Selected=0x995B0CF8
   DockNode          ID=0x00000008 Parent=0x3BC79352 SizeRef=1228,876 Split=X
     DockNode        ID=0x00000001 Parent=0x00000008 SizeRef=370,696 Split=Y Selected=0xC89E3217
       DockNode      ID=0x00000005 Parent=0x00000001 SizeRef=370,435 Selected=0x9A68760C
--- a/Hazelnut/src/EditorLayer.cpp
+++ b/Hazelnut/src/EditorLayer.cpp
@@ -287,17 +287,40 @@ namespace Hazel {
 				const wchar_t* path = (const wchar_t*)payload->Data;
 				std::filesystem::path parentPath = std::filesystem::path(path).parent_path();
 
-				if (parentPath == "scenes")			// Load scene
+				// MajiSubhra code to drag texture using directory
+				
+				//if (parentPath == "scenes")			// Load scene
+				//	OpenScene(std::filesystem::path(g_AssetPath) / path);
+				//else if (parentPath == "textures" && name != "None")	// Load texture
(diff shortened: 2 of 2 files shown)


VagueLobster:
May I suggest that you change it to something like this:
`std::filesystem::path filePath = std::filesystem::path(path);

				if (filePath.extension().string() == ".hazel" || filePath.extension().string() == ".ncz")
				{
					OpenScene(std::filesystem::path(g_AssetPath) / path);
				}
				else if (filePath.extension().string() == ".png")
				{
					std::filesystem::path texturePath = std::filesystem::path(g_AssetPath) / path;
					Ref<Texture2D> texture = Texture2D::Create(texturePath.string());
					if (texture->IsLoaded())
					{
						if (m_HoveredEntity && m_HoveredEntity.HasComponent<SpriteRendererComponent>())
							m_HoveredEntity.GetComponent<SpriteRendererComponent>().Texture = texture;
					}
					else
					{
						HZ_WARN("Could not load texture {0}", texturePath.filename().string());
					}
				}`

and keep the rest you had? Because the code you provided will crash if the user tries to drag a texture into the scene and there is no highlighted entity! this fixed that problem and also i personally do not like the style of using the folder structure from 'scenes' and 'textures'... what i mean is, what if somebody has a scene or texture outside of those folders? therefore i made this and you can freely copy it and update your pull request :) Please let me know if this is fine by you :)

MajiSubhra:
> May I suggest that you change it to something like this:
> 
> ```
> 			std::filesystem::path filePath = std::filesystem::path(path);
> 
> 			if (filePath.extension().string() == ".hazel")
> 			{
> 				OpenScene(std::filesystem::path(g_AssetPath) / path);
> 			}
> 			else if (filePath.extension().string() == ".png")
> 			{
> 				std::filesystem::path texturePath = std::filesystem::path(g_AssetPath) / path;
> 				Ref<Texture2D> texture = Texture2D::Create(texturePath.string());
> 				if (texture->IsLoaded())
> 				{
> 					if (m_HoveredEntity && m_HoveredEntity.HasComponent<SpriteRendererComponent>())
> 						m_HoveredEntity.GetComponent<SpriteRendererComponent>().Texture = texture;
> 				}
> 				else
> 				{
> 					HZ_WARN("Could not load texture {0}", texturePath.filename().string());
> 				}
> 			}
> ```
> 
> and keep the rest you had? Because the code you provided will crash if the user tries to drag a texture into the scene and there is no highlighted entity! this fixes that problem and also i personally do not like the style of using the folder structure from 'scenes' and 'textures'... what i mean is, what if somebody has a scene or texture outside of those folders? therefore i made this and you can freely copy it and update your pull request :) Please let me know if this is fine by you :)


Thanks for pointing this out. Initially, I used the extension to identify the action. Then I thought that a scene may not have the ".hazel" extension or a texture may not necessarily be a ".png" file always. Therefore, I changed it to directories. 

Regarding the crash, I think, a condition can be added to prevent that.

Bottom line is that it's a design issue and I have no idea whether Cherno will maintain the directory structure or the scene format or the texture format. It will be great if he takes a call on this. In the meantime, I will work on the bug you have pointed out.

Thanks again for your feedback.

MajiSubhra:
@VagueLobster  I have added your code in the new commit. I have also fixed the bug you pointed out. Please have a look at the updated version.

VagueLobster:
I have just looked at the new commit, and it works now, but why keep the old code? Is there a reason for that?

MajiSubhra:
As I mentioned in my first response that I have no idea whether Cherno will maintain the directory structure or the scene format or the texture format. Therefore, I leave it to him to keep it or remove it.

VagueLobster:
oh okay :) now i get it :D

TheCherno:
Overall a nice feature! Please remove comments, remove `imgui.ini` from commit, make sure it still merges with no conflicts, and I'll merge it in. 👍

VagueLobster:
@MajisTcd I think @TheCherno wanted you to remove the comments, not re-use the old code that you commented out! 🙂

MajisTcd:
@VagueLobster I was about to respond to you, I was just trying to figure out what exactly I need to do. In my machine, the code works fine. If you think you can still find bugs, I am changing it to your code. However, I am not sure about "delete the comments". If you have got it, can you please tell me which comments I should delete?

VagueLobster:
> @VagueLobster I was about to respond to you, I was just trying to figure out what exactly I need to do. In my machine, the code works fine. If you think you can still find bugs, I am changing it to your code. However, I am not sure about "delete the comments". If you have got it, can you please tell me which comments I should delete?

The bug, i thought i could reproduce is not a bug it seems! However, maybe it's just me, but i don't like to be forced to have to use the `textures` folder to be able to import textures.
I think that the user should have a choice  as to where they want they're textures to be 🙂 
Also the users shouldn't be required either to have their scenes inside `scenes` folder, again it should be their free choice!
That was why i changed your code back in September last year! 🙂

VagueLobster:
@MajisTcd Actually, there is a minor bug in your code, i just found:
Because of this line of code `else if (parentPath == "textures" && name != "None")`
the user cannot add textures to sprites that is named None! Again the user should have the freedom to choose
whatever they want imo! Of course maybe not in every instance, but this, i don't see any harm in having the freedom 🙂

VagueLobster:
> @VagueLobster Thanks for your time and suggestions. This is my first time contributing to a public repository. So I am having some trouble using git. I have kept your code but removed the comment mentioning that it's your code. Please let me know how to acknowledge your contribution and I will be happy to do so.

No problem 🙂 You don't have to mention me or that I have changed your code, i'm just happy to help 😄 
But you should change your repository to have the imgui.ini from Cherno's most recent pull request and NOT changing it any further, otherwise github registers it as a change that must be commited! Also you need to remove the `ColoredCube.hazel` from this repository when commiting, because it's not a file that Cherno has made! And it's not necessary 😄 Otherwise it looks good 👍

PR #494, opened 2021-09-29
Add unit test support & automated test runs on push/PR resolves #99

### LINKED ISSUE #99: Unit Tests
It looks like Hazel will expand to become quite a large and complicated code base (as is common for game engines). Because of this Cherno is making many temporary compromises in order to get things working so long. For example the hard-coded magnification settings in the Textures video. This will eventually lead to us refactoring a lot of code as a proper material system, asset manager, etc. gets added.

I think it is probably a good idea to have a unit test system so that if this refactoring (or anything else) breaks something unexpectedly, then we will be able to find where the issue is and hopefully fix it quickly without too much debugging.

I am not experienced enough to know how to actually go about doing this, but it seems like something that should be done at some point.

niklasweidenfeller:
Hi there,

I've added support for Googletest including a test executable. As far as I know unit tests are not present in Hazel as of now, but I would kindly suggest to start using tests in the future.

I have also enabled github actions for compiling with msvc and executing the unit tests as automated action on push/on PR (runns only in my fork currently, once the PR gets merged it will be enabled in this repo as well).

I guess this setting https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/ needs to be enabled, so that the git workflow runs when people do a PR from their private forks.

Note: This PR uses googletest as a submodule, which is currently a googletest fork in my own github profile. I've created a Premake file for googletest there. If this PR eventually gets integrated into Hazel, one would have to fork googletest into Cherno's Github and add [this premake file](https://github.com/niklasweidenfeller/googletest/blob/premake/premake5.lua).

grafik

grafik


niklasweidenfeller:
I guess this setting https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/ needs to be enabled, so that the git workflow runs when people do a PR from their private forks.

.gitmodules
```diff
@@ -20,3 +20,6 @@
 [submodule "Hazel/vendor/ImGuizmo"]
 	path = Hazel/vendor/ImGuizmo
 	url = https://github.com/thecherno/imguizmo
+[submodule "Hazel/vendor/googletest"]
+	path = Hazel/vendor/googletest
+	url = https://github.com/niklasweidenfeller/googletest
```
niklasweidenfeller, line 25: This one needs to be changed before merging.

PR #515, opened 2021-11-29
Update SetupPremake.py

abhra2020-smart:
No need for @classmethod


CybernetHacker14:
Doesn't that make it an instance method, whereas the goal is to call the method directly by the class name namespace without creating an object?

abhra2020-smart:
Yes.

Issue #535, opened 2022-01-26
static_assert(false) only compiles with Visual C++


nepp95:
#### Describe the bug
In `Hazel/Scene/Scene.cpp` there is a `static_assert(false)`, which gives us the kind reminder to add the `OnComponentAdded` function for a new component we add. Compilation fails if we have not added every template variant for every component. This is the desired functionality, but only compiles using Visual C++.
While this is the only supported IDE at this moment, it should be fixed, since compiler dependent code is generally not what you'd want. Also, IF Hazel would be to support other platforms/compilers, the error generated by the compiler is the most useless you will ever encounter.

```
error: static_assert failed
static_assert(false);
```

#### To Reproduce
1. Compile on any other platform/compiler than Visual C++
2. Get above error

#### Expected behavior
You would want it to compile any day, any time, any compiler. If it does not compile, you would want a clear error on the problem.

#### Operating system:
MacBook Pro M1 2020 (ARM architecture) with macOS Monterey 12.1 Beta

#### Additional context
Will be providing a pull request with the fix.

PR #536, opened 2022-01-26
Changed static assert condition

nepp95:
In `Hazel/Scene/Scene.cpp` there is a `static_assert(false)`, which gives us the kind reminder to add the `OnComponentAdded` function for a new component we add. Compilation fails if we have not added every template variant for every component. This is the desired functionality, but only compiles using Visual C++.
While this is the only supported IDE at this moment, it should be fixed, since compiler dependent code is generally not what you'd want. Also, IF Hazel would be to support other platforms/compilers, the error generated by the compiler is the most useless you will ever encounter.

#### PR impact

Issues this solves       | #535 

#### Proposed fix
Because the original condition in the assertion NEVER resolves to true, some compilers don't even look at the template function and just say "Nope, won't compile this". By adding in the parameter of the template, the compiler HAS to look at what this template consists of, and thus will compile if the condition is met. Example of the error generated when a component template function is missing (Compiler: Apple Clang 13):

```
error: static_assert failed due to requirement 'sizeof(IDComponent) == 0'
static_assert(sizeof(T) == 0);
note: in instantiation of function template specialization 'Scene::OnComponentAdded<IDComponent>' requested here m_scene->OnComponentAdded<T>(*this, component);
note: in instantiation of function template specialization 'Entity::AddComponent<IDComponent, UUID &>' requested here entity.AddComponent<IDComponent>(uuid);
```

Note: On this compiler, the `static_assert(false)` doesn't compile at all, even if all template specializations are there. This fix not only solves that, but generates a clear error when a template specialization is missing.

#### Additional context
Tested on two systems:

System: MacBook Pro M1 2020
OS: macOS Monterey 12.1 Beta
Compiler: Apple clang version 13.0.0 (clang-1300.0.29.30)

System: Intel Core i3-8100, AMD RX 580 8GB, 16GB DDR4 RAM
OS: Windows 11 Home 21H2
Compiler: Visual Studio 2019 v142 with windows sdk 10.0

If needed, I can provide more information about the compiler specific errors.


nepp95:
Optionally, since part of this issue is about the error being generated by the assertion, a custom error message can be passed to the assertion as the 2nd parameter.

PR #537, opened 2022-02-06
Dev

TheMastermindInventor:
In Hazel/Scene/Scene.cpp there is a static_assert(false), which gives us the kind reminder to add the OnComponentAdded function for a new component we add. Compilation fails if we have not added every template variant for every component. This is the desired functionality, but only compiles using Visual C++.
While this is the only supported IDE at this moment, it should be fixed, since compiler dependent code is generally not what you'd want. Also, IF Hazel would be to support other platforms/compilers, the error generated by the compiler is the most useless you will ever encounter.

error: static_assert failed
static_assert(false);
To Reproduce

Compile on any other platform/compiler than Visual C++
Get above error
Expected behavior
You would want it to compile any day, any time, any compiler. If it does not compile, you would want a clear error on the problem.

Operating system:
MacBook Pro M1 2020 (ARM architecture) with macOS Monterey 12.1 Beta

Additional context
Will be providing a pull request with the fix.

FIX : 
Add the fix in Hazel/Scene/Scene.cpp
just paste on top of void Scene::OnComponentAdded(Entity entity, T& component)
in the original project 

	template<typename T>
	struct foobar : std::false_type
	{ };
	template<typename T>
	void Scene::OnComponentAdded(Entity entity, T& component)
	{
		static_assert(foobar<T>::value);
	}


LovelySanta:
This should not be a PR but rather an issue. Closing as invalid.

PR #517, opened 2021-12-05
Fixed visualization for circle collider

Roshiago:
#### Describe the issue
Circle collider not appear correctly when we move editor camera behind circle entity

#### PR impact
List of related issues/PRs this will solve:

 Impact                  | Issue/PR
------------------------ | ------
Issues this solves       | None
Other PRs this solves    | None

#### Proposed fix
In youtube video [Visualizing Physics Colliders](https://youtu.be/rr1GPVzyG00?t=1838) TheCherno gave an exercise for fixing visualization of Circle Collider. This pull request is that exercise.

PR #559, opened 2022-06-25
Physics colliders now show on both sides.

VagueLobster:
Yan made a live stream last year, where he wrote this code to display physics colliders on both sides of sprites and circles, depending on the cameras projection z axis.

This is also what i mentioned in the last live stream (from 20/06/2022) where i thought this code was in the dev branch.

This won't resolve any previous PRs or issues that i know of, but it resolves having only physics colliders show on one side of the sprites and circles.

I have simply copied the code back then, into Hazel 2D, and thought that Cherno would like to not have to write it again :)


erikrl2:
I'm sorry, but this is a duplicate of #517.

PR #560, opened 2022-06-26
Additional Scope and Ref Functions

silvertakana:
There isn't an issue. This is just an addition for more functionality

### DIFF
--- a/Hazel/src/Hazel/Core/Base.h
+++ b/Hazel/src/Hazel/Core/Base.h
@@ -29,14 +29,29 @@ namespace Hazel {
 
 	template<typename T>
 	using Scope = std::unique_ptr<T>;
+	template<typename T>
+	constexpr Scope<T> GetScope(T& ptr)
+	{
+		return std::unique_ptr<T>(&ptr);
+	}
 	template<typename T, typename ... Args>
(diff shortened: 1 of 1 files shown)


JonathanHiggs:
`GetRef` is bugged, super easy to end up in undefined behaviour, either with a local variable that has been deleted while still in scope, or a ref that points to memory that has already been deleted (see examples below)

The issue is that passing in a T& implies that the caller still owns the memory, but you are passing that into shared_ptr which implies it is taking over ownership of the memory

```cpp
struct Foo
{
    bool& deleted;
    ~Foo() { deleted = true; }
};


TEST(RefTest, RefDeletesInScopeValue)
{
    bool deleted = false;
    Foo foo { deleted };

    {
        auto ref = GetRef<Foo>(foo);
    }

    // foo is still in scope but has been deleted by the ref going out of scope
    EXPECT_FALSE(deleted);
}

TEST(RefTest, RefPointsToReclaimedMemory)
{
    bool deleted = false;
    auto fnReturnRef = [&]() -> Ref<Foo> {
        Foo foo{ deleted };
        return GetRef<Foo>(foo);
    };

    auto ref = fnReturnRef();

    // foo was deleted when it went out of scope so ref points to reclaimed memory
    EXPECT_FALSE(deleted);
}
```

PR #562, opened 2022-06-28
Scene hierarchy won't show "Create Empty Entity" menu!

VagueLobster:
#### Describe the issue (if no issue has been made)
Currently with the code used in Hazel 2D, and when launching Hazel 2D,
you can't get the "Create Empty Entity" to pop up.

This bug is ONLY happening when you haven't loaded a scene OR created a new scene!

 Impact                  | Issue/PR
------------------------ | ------
Issues this solves       | None or #number(s)
Other PRs this solves    | None or #number(s)

#### Proposed fix _(Make sure you've read [on how to contribute])
By adding:
`m_SceneHierarchyPanel.SetContext(m_ActiveScene);`
after `m_ActiveScene = m_EditorScene;` inside `OnAttach()` in `EditorLayer.cpp`
right-clicking to display the menu works again!

Steps to reproduce the bug:
1: open up Hazel 2D.
2: Move mouse cursor over to the Scene hierarchy.
3: Right-click and see that the "Create Empty Entity" won't pop up.

PR #506, opened 2021-11-07
Fix new scenes being overwritten by old scene

CleverSource:
#### Bug 1
When you load a scene and then create a new one (via Ctrl + N; this runs the ``NewScene`` function) then your new scene will be overwritten the next time you try to play the scene, this is due to ``m_EditorScene`` never being set in the ``NewScene`` function.

#### Bug 2
This PR also fixes another problem where the editor does not correctly initialize a starting scene, which causes a crash if you play the scene, and it does not allow you to create entities in the "Scene Hierarchy Panel" until you've created a new scene (via Ctrl + N)

#### Miscellaneous Fix
When creating a new scene while a scene is already running the editor will crash. I have decided it is best to make ``NewScene`` do *nothing* when the editor is in a "playing" state. My reasoning for this is that the user might not want to overwrite their scene, or have their scene forcefully stopped while in play mode when creating a new scene.

I also noticed in a recent video from Cherno that he did not realize there was a ``Ctrl + S`` feature because it didn't appear in the dropdown menu, so I've added this to the dropdown menus.

#### Reproduction steps for Bug 1:
- Open the editor in ``Debug`` mode.
- Load a random scene (i.e. ``Physics2D.hazel``)
- Create a new scene (via dropdown menu or Ctrl + N)
- Press play. Your scene will now be overwritten permanently.

#### Reproduction steps for Bug 2:
- Open the editor in ``Debug`` mode.
- Try to create an entity in the "Scene Hierarchy Panel" (you will not be able to due to the hierarchy panel having no context of the scene)
- Attempt to press play. This will crash the program due to ``m_EditorScene`` not being set, which will supply the ``Scene::Copy`` function a ``nullptr`` scene.

#### Proposed fix _(Make sure you've read [on how to contribute](https://github.com/TheCherno/Hazel/blob/master/.github/CONTRIBUTING.md) to Hazel)_
My proposed fix for "Bug 1" was to set ``m_ActiveScene`` to ``m_EditorScene`` and initializing ``m_EditorScene`` in the ``NewScene`` function. This is inline with how ``OpenScene`` works.

My proposed fix for "Bug 2" was to run the ``NewScene()`` function to initialize the editor's first scene rather than initializing it with ``m_ActiveScene = CreateRef<Scene>()``.

#### Additional context
This has been tested in ``Debug`` and ``Release`` mode and runs properly in both.

PR #563, opened 2022-07-01
Fixed issues with `EditorLayer::NewScene()`

erikrl2:
#### Describe the issue(s)
* When you try to create a new empty scene while playing/simulating, Hazelnut crashes.
* When you open a new (empty) scene (without having loaded a scene file before), then modify and play the scene, an empty `m_EditorScene` is copied to `m_ActiveScene`, causing loss off all progress.

#### PR impact

#### Proposed fix
* Stop the scene if it is currently playing/simulating.
* Create a new `m_EditorScene` instead of `m_ActiveScene` in `NewScene()`.

### DIFF
--- a/Hazelnut/src/EditorLayer.cpp
+++ b/Hazelnut/src/EditorLayer.cpp
@@ -569,6 +569,8 @@ namespace Hazel {
 		m_EditorScene = CreateRef<Scene>();
 		m_EditorScene->OnViewportResize((uint32_t)m_ViewportSize.x, (uint32_t)m_ViewportSize.y);
 		m_SceneHierarchyPanel.SetContext(m_EditorScene);
+
+		m_ActiveScene = m_EditorScene;
 		
 		m_EditorScenePath = std::filesystem::path();
 	}
--- a/Hazelnut/src/EditorLayer.cpp
+++ b/Hazelnut/src/EditorLayer.cpp
@@ -563,9 +563,12 @@ namespace Hazel {
 
 	void EditorLayer::NewScene()
 	{
-		m_ActiveScene = CreateRef<Scene>();
-		m_ActiveScene->OnViewportResize((uint32_t)m_ViewportSize.x, (uint32_t)m_ViewportSize.y);
-		m_SceneHierarchyPanel.SetContext(m_ActiveScene);
+		if (m_SceneState != SceneState::Edit)
+			OnSceneStop();
+
+		m_EditorScene = CreateRef<Scene>();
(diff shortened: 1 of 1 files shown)


VagueLobster:
You just couldn't wait for @TheCherno to see my PR, huh? 😞

erikrl2:
> You just couldn't wait for @TheCherno to see my PR, huh? 😞 I find it disrespectful, but each to their own.

It's not meant to be disrespectful by any means and I'm sorry for that.
But it's still your PR that gets merged probably.

VagueLobster:
We'll see, but if you're so sure that my PR gets merged, then why even make this PR? It makes no sense to me.

erikrl2:
> We'll see, but if you're so sure that my PR gets merged, then why even make this PR? It makes no sense to me.

It resolves completely different issues. It just allows your code from #562 to look cleaner.

VagueLobster:
which issues are they? 🤔 My PR resolves the fact that when launching Hazel 2D, you cannot right-click in the Scene Hierarchy, it won't register it...

erikrl2:
> which issues are they? 🤔 My PR resolves the fact that when launching Hazel 2D, you cannot right-click in the Scene Hierarchy, it won't register it...

The two issues are described in the issue description above.

VagueLobster:
Aah! The first bug, i cannot replicate with my PR changes, like at all! But, the second one is true, i experience that bug myself!

VagueLobster:
Also i deleted the old messages. because i am a numbskull for not looking at your code first, before reacting to it!

erikrl2:
> Aah! The first bug, i cannot replicate with my PR changes, like at all! But, the second one is true, i experience that bug myself!

Oh alright, I see.
And also, my description is inaccurate. By "open a new scene" I mean creating a new scene. This issue isn't solved by your PR.
I will go ahead and change it.

CleverSource:
``OnSceneStop`` should not be called when ``NewScene`` is ran. This will result in lost work of an existing scene because the end user never has the chance to save their existing scene. 

It can be assumed that if an end user created a new scene (either through the dropdown menu or Ctrl+N) that they most likely accidentally did so. I can't imagine a situation where you're playing/simulating a scene and all of a sudden want to load a blank scene and lose existing progress before saving. The proper solution is doing nothing when creating a new scene if the user isn't in "edit" mode or giving them a chance to save their work before unloading the scene.

Cherno most likely prefers this too considering he intends to merge my PR (#506) that implements this fix. 

Personally it does not matter to me which PR is merged, as they both implement the same fix, although if for some reason Cherno merges your PR instead of mine I think this should be changed.

PR #571, opened 2022-07-19
Added function to convert parameter pack to void**

### LINKED ISSUE #570: Add function to convert parameter pack into void** for mono
#### Is your feature request related to a problem? Please describe
When passing parameters to Mono, it is a bit annoying to have to pack parameters manually into a void* array. It would be more convenient to be able to use a templated parameter pack and have it be able to pack the paramters into a correctly sized void* array.

#### Describe the solution you'd like
I have played around with a templated function in the ScriptEngine.cpp Utils namespace with the signature
```
template<typename T, typename Func, typename... Args>
static T* WrapArgs(Func function, Args&&... args)
```
`function` is a lambda function that will wrap the intended function to call which requires the void* array,
and `args` is a variadic parameter pack which is to be converted to a void* array. `T*` is the return type of the lambda and must be specified as a template parameter, it cannot be deduced.

Example of use:

```
const char* methodName = "PrintInts";

MonoObject* returnValue = Utils::WrapArgs<MonoObject>([=](void** argv, int argc) {
    MonoMethod* method = s_Data->EntityClass.GetMethod(methodName, argc);
    return s_Data->EntityClass.InvokeMethod(instance, method, argv);
}, returnValue, 5, 508);

// instance is assumed to already exist.
```

#### Describe alternatives you've considered
Originally tried to create a function with the signature `static T* WrapArgs(Func function, Args&&... args)` but it didn't prove possible to deduce the type of T* as it was only used as the return type of `function`.

amayesingnathan:
It would be more convenient to be able to use a parameter pack for mono method parameters than to have to always manually put parameters in a void* array. See #570 for more details.

Issues this solves       | Resolves #570

I have added a function to the Utils namespace of ScriptEngine.cpp with the signature 

```
template<typename T, typename Func, typename... Args>
static T* WrapArgs(Func function, Args&&... args)
```

The way to use this function is to pass a lambda as the first argument which wraps the function you intend to use that requires `void**` parameters. The lambda itself will have parameters of `(void** argv, int argc)` to be used with the intended function. The final arguments are a parameter pack which will gets packed into a correctly sized void* array and passed as a paremeter to the lambda. The function returns the result of the lambda, and the type of the return value cannot be deduced automatically so must be passed as template parameter to the function.

Example of use:

```
const char* methodName = "PrintInts";

MonoObject* returnValue = Utils::WrapArgs<MonoObject>([=](void** argv, int argc) {
    MonoMethod* method = s_Data->EntityClass.GetMethod(methodName, argc);
    return s_Data->EntityClass.InvokeMethod(instance, method, argv);
}, returnValue, 5, 508);

// instance is assumed to already exist here.
```

I've also added a templated struct to check whether a parameter type in the parameter pack is a pointer or not, so the correct conversion to a void* can be made.

#### Additional context
Could potentially drop the check to see if `sizeof...(Args)` is zero and just mandate that WrapArgs only be used with a non-empty parameter pack (as it doesn't really make sense to use the function with an empty parameter pack anyway). It was only originally required as `void* argv[argc] = { nullptr };` does not compile with `argc == 0`.

I have tested this compiles by actually using the templated function, but have removed from the main ScriptEngine code to keep as it was previously

### DIFF
--- a/Hazel/src/Hazel/Scripting/ScriptEngine.cpp
+++ b/Hazel/src/Hazel/Scripting/ScriptEngine.cpp
@@ -118,21 +118,6 @@ namespace Hazel {
 				return function(argv, static_cast<int>(argc));
 			}
 		}
-
-		/*
-			Example of use:
-
-				MonoObject* returnValue;
-				Utils::WrapArgs([=](void** argv, int argc) {
-					MonoMethod* method = s_Data->EntityClass.GetMethod("PrintInts", argc);
-					return s_Data->EntityClass.InvokeMethod(instance, method, argv);
-				}, returnValue, 5, 508);
-
-			Here instance is presumed to be captured by value from the same scope the WrapArgs function is in.
-			The WrapArgs cannot deduce the return type of the lambda function to use as its own return type, so the return value is passed back as the 2nd parameter.
-			The remaining paramters are the parameters wished to be packed into a void* array. They are packed inside WrapArgs and passed as parameters to the lambda function.
-		*/ 
-
 	}
 
 	struct ScriptEngineData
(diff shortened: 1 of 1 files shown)


amayesingnathan:
Saw way better solution to this on the discord

PR #577, opened 2022-07-28
Fixed imgui windows not updating if the application window is minimized

Tom5268:
ImGui windows that were dragged out of the application window aren't updating if the application window is minimized.

Instead of only calling OnImGuiRender if the application window isn't minimized, we need to call it always.


Satindar31 (approved):
LGTM!

Satindar31:
Why close?

PR #579, opened 2022-08-23
Added Windows gamepad support

Sausty:
# Gamepad support for Hazel

This pull request contains the basic implementation of a gamepad input system. It only supports regular gamepads, so it isn't possible to connect a guitar hero controller for example.

## Why?

Having an interface for gamepad input will be useful if Hazel ever gets ported to console. It's also very important for an engine to support gamepad input, as it is one of the most popular input methods in the industry.

## How does it work?

Additional functions have been added to the input class to poll button state, trigger state, joystick state and vibration state.
The input system handles deadzone values as well.
You may also notice that there is three new functions in the input class: Init, SetEventCallback and Update.
Init is used to initialise input polling libraries (XInput on Windows for example)
SetEventCallback is used so that the input system can send GamepadConnected and GamepadDisconnected events.
Update is used to actually update the state of the gamepads.

## Example code

```cpp
if (Input::IsGamepadButtonPressed(0, Gamepad::A))
    Input::SetGamepadVibration(0, glm::vec2(1.0f)); // Full speed on both motors

glm::vec2 leftJoystick = Input::GetGamepadLeftJoystick(0);
player_position.x += leftJoystick.x * dt;
player_position.y += leftJoystick.y * dt;
```

## What you might ask

- Why the fuck didn't you use GLFW for gamepad polling you absolute idiot?
    Because GLFW doesn't support gamepad vibration.
- Why is gamepad connect and disconnect an event? Can't you just add a Input::IsGamepadConnected function?
    No, because gamepad connection and disconnection should be handled as notifications : not just a plain old boolean state
- Without GLFW, it's not cross-platform, dude...
    LOL as if Hazel was cross-platform KEKW (on a serious note, when Hazel will be able to build on both Mac and Linux, I'll add
    the necessary backends for GameController and Linux joystick file descriptor)

### DIFF
--- a/Hazel/src/Hazel/Core/Application.cpp
+++ b/Hazel/src/Hazel/Core/Application.cpp
@@ -27,6 +27,9 @@ namespace Hazel {
 		m_Window = Window::Create(WindowProps(m_Specification.Name));
 		m_Window->SetEventCallback(HZ_BIND_EVENT_FN(Application::OnEvent));
 
+		Input::Init();
+		Input::SetEventCallback(HZ_BIND_EVENT_FN(Application::OnEvent));
+
 		Renderer::Init();
 
 		m_ImGuiLayer = new ImGuiLayer();
@@ -107,6 +110,8 @@ namespace Hazel {
 				}
 				m_ImGuiLayer->End();
 			}
+
+			Input::Update();
 
 			m_Window->OnUpdate();
 		}
--- a/Hazel/src/Hazel/Core/GamepadCodes.h
+++ b/Hazel/src/Hazel/Core/GamepadCodes.h
@@ -1,0 +1,34 @@
+#pragma once
+
+namespace Hazel
+{
+	const uint32_t MaxGamepadCount = 4;
+
+	using GamepadCode = uint16_t;
+
+	namespace Gamepad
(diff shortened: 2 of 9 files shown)


gamecoder-nz:
Nice.  I wasn't aware that glfw doesn't support gamepad vibration

gamecoder-nz:
Have you implemented this for mac and linux already?

Sausty:
> Have you implemented this for mac and linux already?

Already adressed in the QNA:

> LOL as if Hazel was cross-platform KEKW (on a serious note, when Hazel will be able to build on both Mac and Linux, I'll add
the necessary backends for Apple's GameController framework and Linux joystick file descriptor)

drsnuggles8:
Any specific reason for moving the EventCallbackFn?

Sausty:
> I think the OP chose to move it so it was easier to use: `static void SetEventCallback(const EventCallbackFn& callback);`
> 
> inside of Input.h **_and_** `virtual void SetEventCallback(const EventCallbackFn& callback) = 0;` inside of window.h

Correct. It's just so that the input system can also access the EventCallbackFn type. No reason not to move it for clarity.

PR #581, opened 2022-09-02
Sets focus to the entity's tag text field after a double-click on it …

rflgf:
…from Scene Hierarchy Panel

Not really much of an issue, just a convention that I believe most GUIs follow and that could be added to Hazelnut as well.

#### PR impact 

#### Proposed fix 
Quality of life improvement feature to Hazelnut. After the user double-clicks on an entity's tag from the Scene Hierarchy Panel, its text input is automatically set on focus. Two details to give special attention: a global `bool g_DoubleClick` was created for this, which could easily be transformed into a member variable of `SceneHierarchyPanel`. I'm not sure which is preferable but I volunteer to make the adaptation if necessary; second detail is, I couldn't get it to only select the entire text if it was double-clicked from the scene hierarchy panel, my solution was quite buggy, so I just made it so interacting with the tag text input always selects everything, which may be undesirable as well. This behavior is also demonstrated in the capture below.

https://user-images.githubusercontent.com/51202489/188043421-707786b4-c0b6-443d-a453-491f3bcd0617.mp4

### DIFF
--- a/Hazelnut/src/Panels/SceneHierarchyPanel.cpp
+++ b/Hazelnut/src/Panels/SceneHierarchyPanel.cpp
@@ -18,6 +18,7 @@
 namespace Hazel {
 
 	extern const std::filesystem::path g_AssetPath;
+	static bool g_DoubleClick = false;
 
 	SceneHierarchyPanel::SceneHierarchyPanel(const Ref<Scene>& context)
 	{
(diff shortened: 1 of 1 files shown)


rflgf:
I'll be closing this PR since after having reflected on it for a while I figured this implementation adds more complexity to the code base than the feature is worth this early in the series.

Issue #582, opened 2022-09-04
[Suggestion] Set the root namespace of Hazel-ScriptCore to "Hazal"


tadashuai:
#### Is your feature request related to a problem? Please describe
The default namespace is the project name("_Hazel-ScriptCore_"), so set it to "Hazel" for the new file.

However, because of the "Source" folder, we have a new issue. Once creating a new cs script in it, the namespace will be ```Hazel.Source```.
The original intention of this suggestion is to **not need to modify the namespace after creating a new script**, but it make no senses just like this. Perhaps creating a new script directly under the solution is an indescribable way. As for the sub namespace, for example ```Hazel.UI```, then we create the folder named UI at this point.

#### Describe the solution you'd like
In the Hazel\Hazel-ScriptCore\premake5.lua, add ```namespace "Hazel"``` below the ```project``` line.
```
project "Hazel-ScriptCore"
    namespace "Hazel"
```

PR #584, opened 2022-09-06
Add pre-build commands.

tadashuai:
1. Everytime after updating C# scripts, we need to build its solution manually, like Ctrl + B.
    Now just build Hazal.sln or **HIT F5** directly!
2. Also see #582 .

Issues this solves       | #582

1. Will build C# projects first when building Hazel C++ projects. Added the pre-build commands in the Hazel premake file.
    By the way, it checks whether the sandbox solution exists and notifies.
2. Only set C# project root namespace. <= Added ```namespace``` line in the C# premake files.


tadashuai:
But if we only changed the C# code, it means the Hazel project doesn't need to be rebuilt and no pre-build commands are executed. At this point we still need to manually build the C# project.

PR #599, opened 2022-12-07
Change class Input to struct

rigo2798:
all members are public (improve readability)


DaveAxiom:
This is unnecessary. `class` is more understandably an object as `struct` was allowed to have methods circa C++11.

PR #600, opened 2022-12-14
Linux

jeg7:
#### Implemented Linux platform compatibility
Added CMake support (for GNU compiler on Debian Linux (Ubuntu 22.04)). Further improvements and refinements can definitely be made, but I will wait to see if and how they should be implemented.


jeg7:
The multiple and somewhat confusing commits are because I initially made the mistake of using the "master" branch instead of the "project" branch, even though I added the updates from the "project" branch.  So I deleted the old files and then added the implementation.  In case you look through the commits, the only relevant ones are the last two (Added Linux platform compatibility. & Fixed a few paths.).

Issue #606, opened 2023-01-11
OnCreate() need invoke after all the ScriptEntities added into "s_Data->EntityInstances"


WhereIsMyID:
When i want get an entity by "**FindEntityByName()**" in "**OnCreate()**" function, there will be a promblem that it would  miss the instance from the function: "**void ScriptEngine::OnCreateEntity(Entity entity)**", because the "**InvokeOnCreate()**" will excute during every loop, so i think the invoke function need another way to do.

PR #607, opened 2023-01-11
OnCreate invkoe after all instances created

WhereIsMyID:
Here is my modification plan after I pushed this issuse -#606 .

Issues this solves       | 1 

I splited the method "OnCreateEntity(Entity entity)" into "CreateEntity(Entity entity)" and "OnCreateEntities()".
One is to register all instances in the instances map, another one is to invoke their OnCreate() function.

#### Additional context

### DIFF
--- a/Hazel/src/Hazel/Scene/Scene.cpp
+++ b/Hazel/src/Hazel/Scene/Scene.cpp
@@ -146,8 +146,9 @@ namespace Hazel {
 			for (auto e : view)
 			{
 				Entity entity = { e, this };
-				ScriptEngine::OnCreateEntity(entity);
+				ScriptEngine::CreateEntity(entity);
 			}
+			ScriptEngine::OnCreateEntities();
 		}
 	}
 
--- a/Hazel/src/Hazel/Scripting/ScriptEngine.cpp
+++ b/Hazel/src/Hazel/Scripting/ScriptEngine.cpp
@@ -283,10 +283,10 @@ namespace Hazel {
 		return s_Data->EntityClasses.find(fullClassName) != s_Data->EntityClasses.end();
 	}
 
-	void ScriptEngine::OnCreateEntity(Entity entity)
+	void ScriptEngine::CreateEntity(Entity entity)
 	{
 		const auto& sc = entity.GetComponent<ScriptComponent>();
-		if (ScriptEngine::EntityClassExists(sc.ClassName))
+		if (EntityClassExists(sc.ClassName))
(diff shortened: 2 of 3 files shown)


VagueLobster:
You must have done something wrong in your code, because I have tried to replicate your issue and I just cannot replicate it :/
I have even tried your "fix" and it does absolutely nothing on my 1-1 copy of little Hazel :(

WhereIsMyID:
> 

This is a logical issue with program execution. For example: There are two GameObjects A and B in the scene. We assume that A is created before B. There is no problem to use "GetComponent" from A in the OnCreate() function of B, but if A want to get a component from B or get the instance of B, there will be fail, because B was not yet created when A called the OnCreate() function.
I hope u can understand the meaning of my above paragraph :)

WhereIsMyID:
> You must have done something wrong in your code, because I have tried to replicate your issue and I just cannot replicate it :/ I have even tried your "fix" and it does absolutely nothing on my 1-1 copy of little Hazel :(

This is a logical issue with program execution. For example: There are two GameObjects A and B in the scene. We assume that A is created before B. There is no problem to use "GetComponent" from A in the OnCreate() function of B, but if A want to get a component from B or get the instance of B, there will be fail, because B was not yet created when A called the OnCreate() function.
I hope u can understand the meaning of my above paragraph :)

VagueLobster:
Yeah I think I get the gist of it, I just cannot replicate it for some reason... Like I said, it works fine for me...

VagueLobster:
And I think I know why I cannot reproduce the error... Yan has updated a lot of code when he went on to create the Asset Manager System, and while he did that, I think he has updated some of the code in Scene.cpp(?) which happens to fix the Hierarchy turning upside-down every time you save! So with the new code the Hierarchy only turns upside-down whenever you run the scenes! That could be it! But don't think that you can settle with just downloading the asset-manager branch and all is good, because you will need to watch the latest VOD on twitch, since Yan hasn't uploaded it yet to youtube! And if you don't know it, the VOD is only up for a total of 60 days on twitch, and it has already been 14 days since he streamed it! Hope this helps 🙂

VagueLobster:
My pleasure 😄👍

VagueLobster:
> Maybe so. I published this issue last year. I'll look at the current source code again. Thanks for communication. :D
I have just found out that I made a slight bummer ☹️It's the VOD from 21 days ago today that is the last little Hazel VOD, sorry about that!

PR #617, opened 2023-03-20
Fix for cloning issues with msdf-atlas-gen

Szilard0605:
Cloning with git will not include msdf-atlas-gen because of the backslash characters in .gitmodules.


Frozby:
just wanted to try out pulling for the first time because of this, had to change it for myself and redo the setup, would be nice if it was done its such a simple thing :D

eagleF4life (approved):

bread-game-dev-real:
idk

PR #633, opened 2023-07-09
Fixed unresponsive ui during window resize

### LINKED ISSUE #632: Unresponsive UI during window resize!!
### Probelm
The hazelnut UI becomes unresponsive when we resize the window. The reason for that is that our Application run loop pauses as it is still resolving Win32 messages. This creates the possibility of things like this:

image

* This isn't that big of an issue, but when our scene is running and we resize, our scene pauses and when we stop resizing we skip everything that happened.
* Our scripting engine also pauses, potentially causing unpredicted behavior (this is just a theory, since I haven't tested this much as I had already fixed this unresponsive UI problem on the engine that I am building alongside @TheCherno).
* Since we don't have a minimum size for our window, this is also possible.

image

### Solution
We can update our window and layers in the `OnWindowResize` function in `Application`.

zod24:
#### Issue
fixes #632.
#### Proposed fix
This issue can be resolved by updating our Window and Layers during window resize event.

### DIFF
--- a/Hazel/src/Hazel/Core/Application.cpp
+++ b/Hazel/src/Hazel/Core/Application.cpp
@@ -90,37 +90,40 @@ namespace Hazel {
 	{
 		HZ_PROFILE_FUNCTION();
 
-		while (m_Running)
-		{
+		while (m_Running) {
 			HZ_PROFILE_SCOPE("RunLoop");
+			OnUpdate();
+		}
+	}
 
-			float time = Time::GetTime();
-			Timestep timestep = time - m_LastFrameTime;
-			m_LastFrameTime = time;
+	void Application::OnUpdate()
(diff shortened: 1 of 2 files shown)


MirrasHue:
Hey Zod, this still has the problem of no updates happening when you just hold the border of the window with the mouse left button. This also causes huge delta times depending on how long you hold the border.

I was able to get around this by updating and rendering in a separate thread, and synchronizing the window resizes using two std::atomic's (it has to be done in the main thread). The only thing that doesn't work with this approach is imgui's viewports (floating windows), which is fine for me given all the other benefits.

https://github.com/MirrasHue/MirrasENGINE/blob/main/MirrasENGINE/src/Core/Application.cpp#L120

PR #634, opened 2023-07-10
Update .gitmodules to replace backslashes with forward slashes

PCForrest:
When executing scripts/setup.bat, the msdf-atlas-gen submodule cannot be initialised because the .gitmodules entry contains backslashes instead of forward slashes.

Replacing the backslashes with forward slashes fixes the problem.


PCForrest:
Duplicate of #617

